PR #18, opened 2018-10-19
Issue10 new proto

gregscullard:
Implements latest proto files with BUSY response code handling


gregscullard:
Note, this PR also includes a fix to timestamps which need a few seconds removed in order to cater for minor time differences between clients and node. This is now set to -10s

PR #56, opened 2019-01-11
Remove redundant nextBytes call.

si618:
At first I was wondering how 16 bytes = 64 bits, but unless my understanding of how SecureRandom.nextLong works is wrong, it isn't necessary anyway.


gregscullard:
Thank you for the PR. Accepted and merged.

PR #95, opened 2019-04-24
fix Travis build

abonander:
supercedes #91 since I can't write to that branch

* renames `.travis-ci.yml` to `.travis.yml`
* removes the `GRADLE_HOME` environment variable which is superfluous on Travis


abonander:
I have no idea why Spotless is failing on Travis because it's passing locally on this branch. Maybe time to nix it.

PR #109, opened 2019-05-24
waiting for record involves a different response code

### LINKED ISSUE #99: ContractExecuteTransaction.executeForRecord() returns RECORD_NOT_FOUND
When running a smart contract function with `ContractExecuteTransaction.executeForRecord()`, the method returns `RECORD_NOT_FOUND`.

example
```
ContractExecuteTransaction tx = new ContractExecuteTransaction(hedera.getClient());
        tx.setContract(new ContractId(shardNum, realmNum, contractNum));
        tx.setFunctionParameters(CallParams.function("set_message").add("hi there"));
        tx.setGas(gas);
        TransactionRecord record = tx.executeForRecord();
        FunctionResult result = record.getCallResult();
```

a workaround which gets a receipt first, then a record works

```
   ContractExecuteTransaction tx = new ContractExecuteTransaction(client);
        tx.setContract(newContractId);
        tx.setFunctionParameters(CallParams.function("set_message").add("hi there"));
        var operatorId = new AccountId(2);
        var transactionId = new TransactionId(operatorId);
        tx.setTransactionId(transactionId);

        tx.setGas(gas);

        var receipt = tx.executeForReceipt();
```

in the `executeForRecord()` scenario, the SDK should first wait for a receipt, and only fetch a record if the receipt `ResponseCodeEnum` is `OK` in the first place.

At the time the record is requested, the transaction status is likely still `UNKNOWN` leading to a `RECORD_NOT_FOUND` since one doesn't exist yet.

Note: Pausing the execution in debug, and waiting for the transaction to complete results in a record being recovered from the network.

abonander:
closes #99

@gregscullard we're not sure whether `setGenerateRecord(true)` is required or not because the testnets spuriously return empty records regardless of whether or not it is set:

image

cc @mehcode

### DIFF
--- a/src/main/java/com/hedera/hashgraph/sdk/HederaException.java
+++ b/src/main/java/com/hedera/hashgraph/sdk/HederaException.java
@@ -16,6 +16,8 @@ public class HederaException extends Exception implements HederaThrowable {
     static boolean isCodeExceptional(ResponseCodeEnum responseCode, boolean unknownIsExceptional) {
         switch (responseCode) {
         case UNKNOWN:
+        case RECEIPT_NOT_FOUND:
+        case RECORD_NOT_FOUND:
             if (unknownIsExceptional)
                 break;
             // fall through
--- a/examples/src/main/java/com/hedera/hashgraph/sdk/examples/advanced/TransferCrypto.java
+++ b/examples/src/main/java/com/hedera/hashgraph/sdk/examples/advanced/TransferCrypto.java
@@ -19,13 +19,14 @@ public final class TransferCrypto {
         System.out.println("" + operatorId + " balance = " + senderBalanceBefore);
         System.out.println("" + recipientId + " balance = " + receiptBalanceBefore);
 
-        new CryptoTransferTransaction(client)
+        var record = new CryptoTransferTransaction(client)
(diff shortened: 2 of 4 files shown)


gregscullard:
Currently, `setGenerateRecord(true)` is not required to get a record. A record is generated for every transaction regardless of this flag being set or not.
-If not set, the record is available for 3 minutes (generating this record incurs no additional cost)
-If set, the record is available for 1 hour (this incurs additional transaction fees for the longer term storage)

This doesn't explain the empty record, if a record doesn't ( or no longer ) exists a `RECORD_NOT_FOUND` error will be returned by the network.

src/main/java/com/hedera/hashgraph/sdk/HederaException.java
```diff
@@ -16,6 +16,8 @@
     static boolean isCodeExceptional(ResponseCodeEnum responseCode, boolean unknownIsExceptional) {
         switch (responseCode) {
         case UNKNOWN:
+        case RECEIPT_NOT_FOUND:
+        case RECORD_NOT_FOUND:
```
abonander: These need to be removed as they actually are exceptional.

PR #111, opened 2019-05-24
Port tests from Groovy/Spock to Java/JUnit5

abonander:
builds on #114 

cc @mehcode


abonander:
Codacy is failing on some visibility stuff that isn't really consequential for tests anyway, but it is a good inspection for the main source code. Should I just fix the visibilities in the tests or can we turn off this inspection for that directory?

PR #119, opened 2019-06-04
enables generation of a transaction protobuf that hasnt been signed

### LINKED ISSUE #101: CryptoTransferTransaction toProto() and toBytes() enforce validation of sigs
We have use cases were a transaction needs to be built but not signed until later on. The current implementation of `toProto()` and `toBytes()` enforces a validation that at least one signature is present.

Can we implement a `no validation` flag or alternative, or at least a `validateWithoutSignatureCheck`

QuestofIranon:
closes #101 

This introduces the ability to generate unsigned protobufs and byte arrays through the use of `unsigned` variants of `Validate`, `toProto`, and `toBytes`

### DIFF
--- a/src/main/java/com/hedera/hashgraph/sdk/account/AccountAddClaimTransaction.java
+++ b/src/main/java/com/hedera/hashgraph/sdk/account/AccountAddClaimTransaction.java
@@ -18,7 +18,7 @@ public final class AccountAddClaimTransaction extends TransactionBuilder<Account
         super(client);
     }
 
-    AccountAddClaimTransaction() {
+    public AccountAddClaimTransaction() {
         super(null);
     }
 
--- a/src/main/java/com/hedera/hashgraph/sdk/account/AccountAddClaimTransaction.java
+++ b/src/main/java/com/hedera/hashgraph/sdk/account/AccountAddClaimTransaction.java
@@ -8,18 +8,16 @@ import com.hedera.hashgraph.sdk.proto.Claim;
 import com.hedera.hashgraph.sdk.proto.*;
 import io.grpc.MethodDescriptor;
 
+import javax.annotation.Nullable;
+
 // corresponds to `CryptoAddClaimTransaction`
 public final class AccountAddClaimTransaction extends TransactionBuilder<AccountAddClaimTransaction> {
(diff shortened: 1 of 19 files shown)


src/main/java/com/hedera/hashgraph/sdk/account/AccountAddClaimTransaction.java
```diff
@@ -19,7 +19,7 @@ public AccountAddClaimTransaction(Client client) {
         super(client);
     }
 
-    AccountAddClaimTransaction() {
+    public AccountAddClaimTransaction() {
```
abonander: Why are all these constructors set to `public` here? These are for testing only.

QuestofIranon: Making it possible to create an unsigned transaction without a client

QuestofIranon:
@gregscullard Can you review this to see if it matches the functionality you were looking for?

gregscullard:
I've added an example `TxGenerateSignSend.java` to illustrate how this would work, however I have a question.
When the transaction is created, a `nodeId` is specified. When the transaction is eventually sent to the network, it *must* be sent to the same node. E.g. you can't specify 0.0.3 and send to 0.0.4.

lbaird:
Right. If you’re going to ask a node to do some work for you, then you have to pay that node. You can’t just pay someone else. So you have to include that node ID in the transaction you sign. And then you have to send that signed transaction to that node. Because, if you sent it to someone else, they wouldn’t get paid for their efforts in submitting it for you.


________________________________

mehcode:
The SDK has code to ensure the node selection is matched to the node ID in the protobuf.

gregscullard:
In which case I'm happy with the PR.

QuestofIranon:
Updated it so it matches the new standards and can be merged into master.

 @gregscullard I accidentally wiped your example, I didn't realize it was actually committed to my branch when I did a local rebase from master and push. Can you re-add your example or do it as a separate pull request?

gregscullard:
I'll do a separate PR, so I can test the changes at the same time against master ;)

PR #136, opened 2019-06-13
execute*Async(): provide overloads that pass `this` for context

abonander:
cc @mike-burrage-hedera @mehcode


abonander:
I need to add this to `Transaction` as well, hang on.

abonander:
Fixed.

mehcode:
@mike-burrage-hedera does this address what you were looking for?

mike-burrage-hedera:
That'll work (having access to the transaction on error).

PR #137, opened 2019-06-13
Create TxGenerateSignSend.java

gregscullard:


mehcode:
We have https://github.com/hashgraph/hedera-sdk-java/blob/master/examples/src/main/java/com/hedera/hashgraph/sdk/examples/advanced/MultiAppTransfer.java which is the same goal I believe so I'm closing this. Re-open if there is something different here you were trying to illustrate.

PR #139, opened 2019-06-14
CallParams: more types

### LINKED ISSUE #138: Smart contract calls with parameters other than string
Trying to run a smart contract function with a `uint8` input value to no avail.
This is how the call is put together, but the contract reverts.

```
byte newValue = 10;
        byte[] newValueBytes = ByteBuffer.allocate(1).put(newValue).array();
        new ContractExecuteTransaction(client).setContractId(newContractId)
        .setGas(100_000_000)
        .setFunctionParameters(CallParams.function("set_value")
            .add(newValueBytes))
        .executeForReceipt();
```
Suggest implementing methods for setting a `boolean`, `byte`, `int`, `long`, `bigInteger` and `address`.

Also, the methods for recovering data from a call are limited to `int` and `string`, what about `boolean`, `address`, `BigInteger`, etc... for datatypes larger than int or long (e.g. int256).

Finally, how would an array of values be passed to a contract and returned from it ?

abonander:
# REBASED ON #111, MERGE THAT FIRST

cc #138  

Looking at: https://solidity.readthedocs.io/en/v0.5.9/abi-spec.html#types

`Java/Solidity`

- [x] `boolean/bool`
- [x] Signed integer types: `byte/int<8>`, `short/int<16>`, `int/int<32>`, `long/int<64>`, `BigInteger,N/int<N>`
- [x] unsigned variants of the above
- [x] `<new Java type>/address`
- [x] `<new Java type>/function`
- [x] Test parameter encoding of all types and new `FunctionSelector`

Waiting for another PR:
* Array types of all of the above
* Fixed size byte arrays

Omitted:
* Fixed point types are not currently implemented in the Solidity spec or EthereumJ


abonander:
@gregscullard here's a more filled out `CallParams` API; I have yet to add array types. Do you want tuple types as well? (That method could take `(Object... params)` or something.)

gregscullard:
> @gregscullard here's a more filled out `CallParams` API; I have yet to add array types. Do you want tuple types as well? (That method could take `(Object... params)` or something.)

Tuple would be nice but not a priority.

gregscullard:
Note that as far as I know, solidity allows any multiple of 8 for its data types up to 256, so int24 and int40 are valid types. I guess that if they don't fit into a standard equivalent java type, they'd fit into the next available one (int40 fits into long(int64)).
Wondering whether an `add(int size, Object value)` would make sense ? I don't know how often solidity programmers use "non standard" types to save memory/storage/execution costs.
Likewise for the return value which could be any multiple of 8.

mehcode:
The non standard types don't tangibly exist outside of the operational intrinsics in the contract. All the integer types are encoded and decoded equivalently (which, yes, is a huge waste of space).

abonander:
@gregscullard that's the idea behind these methods:

* https://github.com/hashgraph/hedera-sdk-java/pull/139/files#diff-d4afc6344cf4061b472222e651042b7aR111

* https://github.com/hashgraph/hedera-sdk-java/pull/139/files#diff-d4afc6344cf4061b472222e651042b7aR134

* https://github.com/hashgraph/hedera-sdk-java/pull/139/files#diff-d4afc6344cf4061b472222e651042b7aR220

* https://github.com/hashgraph/hedera-sdk-java/pull/139/files#diff-d4afc6344cf4061b472222e651042b7aR241

These let you explicitly set the integer width, truncating or padding the value when passing `long`.

For widths less than 64 when passing a `long` it truncates the same as Java's semantics (literally just chopping it down to the last N bits with no regard for sign); this is safe even for negative numbers because if the user passes a negative `int` and it implicitly widens to a `long`, it gets sign-extended by Java so when it gets chopped down the sign bit is still correct.

For widths greater than 64 when passing a `long` the Solidity padding rules effectively sign-extend the value to a `uint256` so it should still be correct.

I required supplying the width when using `BigInteger` as assuming the width to be the size of the byte representation of the bigint seemed error prone, even though that's what EthereumJ does. I figured it was better to require the user to explicitly set the width for the parameter and then just assert that the bigint fits into that.

@gregscullard if you'd like to merge this to be able to play around with it I'm fine implementing arrays and tuples in a separate PR. I can also rebase #111 so I can write tests for this (it'd be easier than writing tests here and then having to deal with merge conflicts when I'd have to rebase #111 anyway).

abonander:
We've elected to do arrays (fixed-size and dynamic) in a separate PR.

PR #143, opened 2019-06-17
Getting a receipt requires waiting beyond just "OK", we want SUCCESS …

gregscullard:
…or an error


src/main/java/com/hedera/hashgraph/sdk/Transaction.java
```diff
@@ -197,7 +197,11 @@ protected TransactionId mapResponse(TransactionResponse response) throws HederaE
                 }
             } else {
                 HederaException.throwIfExceptional(receiptStatus);
-                return mapReceipt.apply(receipt);
+                if (receiptStatus != ResponseCodeEnum.OK) {
```
mehcode: This should probably be in the earlier `if` condition as that is what is going to wait for the retry delay. I also wouldn't do `!= OK` but `== OK` added to the condition (so errors still fall out).

gregscullard: Made further updates/changes per your comments.

gregscullard: Also reduced MAX_RETRY_ATTEMPT, 100 was unnecessary large
Also made the first wait > 0 (it was RETRY_DELAY * 0).

gregscullard:
Superseded

PR #145, opened 2019-06-21
Implement arrays for common types

### LINKED ISSUE #138: Smart contract calls with parameters other than string

abonander:
TODO:

- [x] Addresses
- [x] implement tests

closes #138


abonander:
cc @mehcode @gregscullard

PR #154, opened 2019-06-26
FunctionResult: implement more result getters

### LINKED ISSUE #150: Smart contract function returns need more parameter types
Only `int` and `string` are currently supported, need support for `address`, `bool` and other datatypes including `bigInteger` for datatypes that don't fit into a java long.

abonander:
closes #150

@gregscullard I added the following, are there any important ones that we're missing?

* `getBool()`
* `getBytes()`
* `getLong()`
* `getBigInt()`
* `getAddress()`
* `getRawValue()` (returning a 32-byte string)

I'm on the fence about returning `ByteString` in these APIs; converting to a byte array unfortunately requires a copy. It is possible to return a `java.nio.ByteBuffer` but reading from that also requires copying.

I'm also on the fence about keeping the overloaded `get*()` methods which return the 0th value interpreted as that type; I'm not sure how much convenience they add over doing `get[Bool, etc.](0)`.

cc @mehcode

### DIFF
--- a/src/main/java/com/hedera/hashgraph/sdk/FunctionResult.java
+++ b/src/main/java/com/hedera/hashgraph/sdk/FunctionResult.java
@@ -5,9 +5,12 @@ import com.hedera.hashgraph.sdk.contract.ContractId;
 import com.hedera.hashgraph.sdk.proto.ContractFunctionResultOrBuilder;
 import com.hedera.hashgraph.sdk.proto.ContractLoginfoOrBuilder;
 
-import javax.annotation.Nullable;
+import java.math.BigInteger;
+import java.nio.ByteBuffer;
 import java.util.List;
 import java.util.stream.Collectors;
+
+import javax.annotation.Nullable;
 
 public final class FunctionResult {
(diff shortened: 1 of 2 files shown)


mehcode:
I agree that those methods don't carry their weight. I'd add some deprecated annotations.

abonander:
@mehcode updated.

rahul-ki:
It seems like the relevant hasn't been added yet. (I cloned the repo and performed `mvn clean install`.)

rahul-ki:
@abonander 
How about returning arrays (and perhaps structs)?

abonander:
@rahul-ki we've decided to punt on tuples/structs in the `CallParams`/`FunctionResult` API for now as we'd like to come up with a design that's more strongly typed than how EthereumJ does things. It will probably involve some sort of builder or nested class.

abonander:
I will implement arrays in a subsequent PR, however.

rahul-ki:
What exactly is `valIndex`? Want to know what parameter to pass into while writing this line?
```
 long message = contractCallResult.getLong(valIndex)
```

Issue #163, opened 2019-07-05
Add fromString method to ContractId


cylon56:
Right now, there isn't a way to create a new ContractId object from a key String. Having a fromString method would be incredibly useful for Hedera applications that require a user to input a contract ID among other uses. See the example below:
```
System.out.println("Enter the contract ID:");
String contractIdString = s.nextLine();
System.out.println("");
ContractId  newContractId = new ContractId.fromString(contractIdString);
```
I'd be willing to work on this feature upon acceptance.

lbaird:
That’s a great idea. It should work as a round trip, so this is always true:

x == ContractId.fromString(x.toString())

It should also be more general in what it accepts. A number with leading zeros should be accepted. A separator of comma or space should be accepted. It might be best to say it will accept any string that contains exactly 3 substrings of contiguous digits, and allows leading zeros. So “1.2.3” would be the same as “$)hs dfgc 001 ^} }]2/3–&(/sh”.

A regular expression would be an easy way to extract the 3 sub strings. And the Long.fromString could be applied to each.


________________________________

cylon56:
Sounds good. I'll start work this weekend.

mehcode:
https://github.com/hashgraph/hedera-sdk-java/blob/master/src/main/java/com/hedera/hashgraph/sdk/account/AccountId.java#L27

It would be nice to think of a way to avoid that function in 3 places (Account, Contract, and File).

lbaird:
Maybe we could have 3 separate methods, but each of them calls the same static utility method, which converts a string to an array of 3 longs.


________________________________

abonander:
This should be closed now that #164 is merged.

abonander:
This should be available in the next release of the SDK which we just published today.

PR #164, opened 2019-07-08
Added IdUtils and fromString methods

cylon56:
Pull request for [Issue #163](https://github.com/hashgraph/hedera-sdk-java/issues/163)

I created an IdUtil file that holds one method for parsing ID strings. I then added a fromString method for both ContractId and FileId that implements this parsing function. I also refactored the existing AccountId.fromString method to use the parsing function as well. There is a test file included that passes for all four methods.

I recognize that having a separate IdUtil file for a 12 line method may be overkill. Ideally, I'd like to pull more of the fromString functionality into the same static utility method but the way each ID is uniquely built makes that impractical. I'm interested if there is a better solution that can be suggested.

### DIFF
--- a/src/main/java/com/hedera/hashgraph/sdk/IdUtil.java
+++ b/src/main/java/com/hedera/hashgraph/sdk/IdUtil.java
@@ -11,6 +11,7 @@ public final class IdUtil {
     public interface WithIdNums<R> {
         R apply(long shardNum, long realmNum, long entityNum);
     }
+
     public static <R> R parseIdString(String id, WithIdNums<R> withIdNums) {
         var rawNums = Splitter.on('.')
             .split(id)
--- a/src/main/java/com/hedera/hashgraph/sdk/IdUtil.java
+++ b/src/main/java/com/hedera/hashgraph/sdk/IdUtil.java
@@ -14,6 +14,4 @@ public final class IdUtil {
             Long.parseLong(rawNums.next())
         };
     }
-
-
 }
--- a/src/main/java/com/hedera/hashgraph/sdk/IdUtil.java
+++ b/src/main/java/com/hedera/hashgraph/sdk/IdUtil.java
@@ -12,7 +12,7 @@ public final class IdUtil {
             .split(id)
             .iterator();
 
-        return new long[]{
+        return new long[] {
             Long.parseLong(rawNums.next()),
             Long.parseLong(rawNums.next()),
             Long.parseLong(rawNums.next())
--- a/src/main/java/com/hedera/hashgraph/sdk/IdUtil.java
+++ b/src/main/java/com/hedera/hashgraph/sdk/IdUtil.java
@@ -1,0 +1,19 @@
+package com.hedera.hashgraph.sdk;
+
+import com.google.common.base.Splitter;
+
+public final class IdUtil {
+    public static long[] parseIdString(String id) {
+        var rawNums = Splitter.on('.')
+            .split(id)
+            .iterator();
+
+       return new long[] {
+            Long.parseLong(rawNums.next()),
+            Long.parseLong(rawNums.next()),
+            Long.parseLong(rawNums.next())
+        };
+    }
+
+
+}
--- a/src/main/java/com/hedera/hashgraph/sdk/IdUtil.java
+++ b/src/main/java/com/hedera/hashgraph/sdk/IdUtil.java
@@ -4,14 +4,15 @@ import com.google.common.base.Splitter;
 
 public final class IdUtil {
 
-    private IdUtil() {  }
+    private IdUtil() {
+    }
 
     public static long[] parseIdString(String id) {
         var rawNums = Splitter.on('.')
             .split(id)
             .iterator();
 
-        return new long[] {
+        return new long[]{
             Long.parseLong(rawNums.next()),
             Long.parseLong(rawNums.next()),
             Long.parseLong(rawNums.next())
--- a/src/main/java/com/hedera/hashgraph/sdk/IdUtil.java
+++ b/src/main/java/com/hedera/hashgraph/sdk/IdUtil.java
@@ -12,7 +12,7 @@ public final class IdUtil {
             .split(id)
             .iterator();
 
-        return new long[] {
+        return new long[]{
             Long.parseLong(rawNums.next()),
             Long.parseLong(rawNums.next()),
             Long.parseLong(rawNums.next())
--- a/src/main/java/com/hedera/hashgraph/sdk/IdUtil.java
+++ b/src/main/java/com/hedera/hashgraph/sdk/IdUtil.java
@@ -12,10 +12,6 @@ public final class IdUtil {
             .split(id)
             .iterator();
 
-        return new long[]{
-            Long.parseLong(rawNums.next()),
-            Long.parseLong(rawNums.next()),
-            Long.parseLong(rawNums.next())
-        };
+        return new long[]{Long.parseLong(rawNums.next()), Long.parseLong(rawNums.next()), Long.parseLong(rawNums.next())};
     }
 }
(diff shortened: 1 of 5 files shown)


src/main/java/com/hedera/hashgraph/sdk/file/FileId.java
```diff
@@ -21,6 +22,23 @@ public FileId(FileIDOrBuilder fileId) {
...
+    /** Constructs a `FileId` from a string formatted as <shardNum>.<realmNum>.<contractNum> */
+    public static FileId fromString(String account) throws IllegalArgumentException {
+
+        long[] rawNums = IdUtil.parseIdString(account);
+        var newFile = FileID.newBuilder();
+
+        try {
+            newFile.setRealmNum(rawNums[0]);
+            newFile.setShardNum(rawNums[1]);
+            newFile.setFileNum(rawNums[2]);
```
abonander: Same here with `FileId(long, long, long)`.

src/main/java/com/hedera/hashgraph/sdk/account/AccountId.java
```diff
@@ -25,16 +26,14 @@ public AccountId(long shardNum, long realmNum, long accountNum) {
...
+        long[] rawNums = IdUtil.parseIdString(account);
         var newAccount = AccountID.newBuilder();
 
         try {
-            newAccount.setRealmNum(Integer.parseInt(rawNums.next()));
-            newAccount.setShardNum(Integer.parseInt(rawNums.next()));
-            newAccount.setAccountNum(Integer.parseInt(rawNums.next()));
+            newAccount.setRealmNum(rawNums[0]);
+            newAccount.setShardNum(rawNums[1]);
+            newAccount.setAccountNum(rawNums[2]);
```
abonander: We have a `AccountId(long, long, long)` constructor, you don't need to call this yourself.

src/main/java/com/hedera/hashgraph/sdk/IdUtil.java
```diff
@@ -0,0 +1,17 @@
...
+    private IdUtil() {
+    }
+
+    public static long[] parseIdString(String id) {
+        var rawNums = Splitter.on('.')
+            .split(id)
+            .iterator();
+
+        return new long[]{Long.parseLong(rawNums.next()), Long.parseLong(rawNums.next()), Long.parseLong(rawNums.next())};
+    }
```
abonander: Instead of returning an array and indexing into that in the constructors of the other types, why not have something like this:

```java
@FunctionalInterface
public interface WithIdNums<R> {
    R apply(long shardNum, long realmNum, long entityNum);
}
```

And then this becomes: 

```java
public static <R> R parseIdString(String id, WithIdNums<R> withIdNums) {
    var rawNums = Splitter.on('.')
        .split(id)
        .iterator();

    return withIdNums.apply(Long.parseLong(rawNums.next()), Long.parseLong(rawNums.next()), Long.parseLong(rawNums.next()));
}
```

And invocation becomes as simple as `IdUtils.parseIdString(idString, AccountId::new)`, etc.

src/main/java/com/hedera/hashgraph/sdk/contract/ContractId.java
```diff
@@ -22,6 +23,23 @@ public ContractId(ContractIDOrBuilder contractID) {
...
+    /** Constructs a `ContractId` from a string formatted as <shardNum>.<realmNum>.<contractNum> */
+    public static ContractId fromString(String account) throws IllegalArgumentException {
+
+        long[] rawNums = IdUtil.parseIdString(account);
+        var newContract = ContractID.newBuilder();
+
+        try {
+            newContract.setRealmNum(rawNums[0]);
+            newContract.setShardNum(rawNums[1]);
+            newContract.setContractNum(rawNums[2]);
```
abonander: Same here with `ContractId(long, long, long)`

src/main/java/com/hedera/hashgraph/sdk/account/AccountId.java
abonander: The builder calls were in the wrong order originally anyway; I believe it's supposed to go `shard, realm, account/contract/file` but I'm not sure. It hasn't mattered yet because both have been `0` so far. It's probably been confused in other places as well. @gregscullard, @lbaird, can you comment on the intended ordering here?

src/main/java/com/hedera/hashgraph/sdk/contract/ContractId.java
```diff
@@ -22,6 +23,23 @@ public ContractId(ContractIDOrBuilder contractID) {
...
+        } catch (Exception e) {
+            throw new IllegalArgumentException("Invalid Id format, should be in format {shardNum}.{realmNum}.{accountNum}");
```
abonander: Can you fix this copypaste error? Should be `{contractNum}` at the end.

src/main/java/com/hedera/hashgraph/sdk/file/FileId.java
```diff
@@ -21,6 +22,23 @@ public FileId(FileIDOrBuilder fileId) {
...
+        } catch (Exception e) {
+            throw new IllegalArgumentException("Invalid Id format, should be in format {shardNum}.{realmNum}.{accountNum}");
```
abonander: Can you fix this copypaste error? Should be `{fileNum}` at the end.

src/test/java/com/hedera/hashgraph/sdk/IdUtilTest.java
```diff
@@ -0,0 +1,49 @@
...
+class IdUtilTest {
+    @Test
+    @DisplayName("parses id string correctly")
+    void testParseString() {
+        long[] correctArray = {0,0,0};
+        long[] returnedArray = IdUtil.parseIdString("0.0.0");
+        assertEquals( correctArray[0], returnedArray[0]);
+        assertEquals( correctArray[1], returnedArray[1]);
+        assertEquals( correctArray[2], returnedArray[2]);
+    }
```
abonander: 'd test the failure paths as well but I wouldn't block this PR on that.

cylon56: Thanks for the suggestions! I've implemented all your advice except for the failure paths. I tried to use assertThrows but I'm not very familiar with Java tests. Are there any good examples you suggest looking at?

abonander: Sure, there's an example here: https://github.com/hashgraph/hedera-sdk-java/blob/master/src/test/java/com/hedera/hashgraph/sdk/CallParamsTest.java#L163-L185

You simply pass the class of the exception you're expecting and then a lambda which should throw that exception; the exception itself is returned from the `assertThrows()` call which you can then assert the message (the extra `message` parameter on `assertThrows()` is the error message for if _that_ assertion fails, not the expected value of the assertion message; I made that mistake before).

cylon56: Thanks! I just added those failure tests.

cylon56:
@abonander Is there anything else I can do to improve this pull request?

abonander:
@cylon56 looks good, I'll get someone with write access to merge it shortly. You can squash your commits if you like, or whoever can do it during merge.

cylon56:
@abonander I haven't done a squash before and I don't think what I just tried was correct. I'll let someone else do it just to be safe. Please let me know if there's anything I need to correct from that attempt.

abonander:
@cylon56 Github isn't showing that anything changed so it should be fine. To squash you would do an interactive rebase on top of the parent commit of f063b83 and mark all commits after the first as `squash` or `fixup` (the former concatenates the commit messages together, the latter discards them).

PR #165, opened 2019-07-10
Fix AccountAddClaims and add an example

QuestofIranon:
Encountered a regression where AccountAddClaims was no longer working after changing from version `0.4.6` to `0.5.1` on bintray. It would return an `invalid accountId` message from the testnet. 

This appears to be the fix. I tested it multiple times and am no longer getting `invalid accountId`. 

I Also added an example of how to use it.


QuestofIranon:
Ignore this, I was testing against an old version of the testnet.

PR #175, opened 2019-08-01
Get query cost

abonander:
a copy of #146 rebased and simplified, still have some problems with `COST_ANSWER` though:

* for most queries it appears to require a nonzero payment even though `COST_ANSWER` should be free


abonander:
This needs to be significantly rebased and will probably just be easier to start over.

PR #177, opened 2019-08-02
Adding method for Bytes32 parameter

JaimeSilver:
Bytes32 parameters need to be fixed and called explicitly


abonander:
This already exists as `addBytes(byte[], int)` where the second parameter is `32`.

PR #178, opened 2019-08-02
Include bytes32 getter

JaimeSilver:
Please include the Bytes32 getter

### DIFF
--- a/src/main/java/com/hedera/hashgraph/sdk/FunctionResult.java
+++ b/src/main/java/com/hedera/hashgraph/sdk/FunctionResult.java
@@ -73,6 +73,12 @@ public final class FunctionResult {
         return getByteString(offset + 32, offset + 32 + len);
     }
 
+    /** Get the nth returned value as a string  from byte 32*/
+    public String getBytes32(int valIndex) {
+        var offset = getInt(valIndex);
+        return getByteString(offset + 0, offset + 32).toStringUtf8();
+    }
+
     /**
      * Get the nth 32-byte value as an untyped byte string.
      */


src/main/java/com/hedera/hashgraph/sdk/FunctionResult.java
```diff
@@ -73,6 +73,12 @@ public ByteString getBytes(int valIndex) {
         return getByteString(offset + 32, offset + 32 + len);
     }
 
+    /** Get the nth returned value as a string  from byte 32*/
+    public String getBytes32(int valIndex) {
```
abonander: `String` isn't the right return type for `bytes32` though. This should return `ByteString` which the user can then call `toStringUtf8` on themselves if they like.

src/main/java/com/hedera/hashgraph/sdk/FunctionResult.java
```diff
@@ -73,6 +73,12 @@ public ByteString getBytes(int valIndex) {
...
+        var offset = getInt(valIndex);
+        return getByteString(offset + 0, offset + 32).toStringUtf8();
```
abonander: `offset + 0` is redundant.

src/main/java/com/hedera/hashgraph/sdk/FunctionResult.java
```diff
@@ -73,6 +73,12 @@ public ByteString getBytes(int valIndex) {
         return getByteString(offset + 32, offset + 32 + len);
     }
 
+    /** Get the nth returned value as a string  from byte 32*/
+    public String getBytes32(int valIndex) {
+        var offset = getInt(valIndex);
```
abonander: `bytes32` is a fixed-width type so doesn't need an offset.

abonander:
Instead of `getBytes32` I would prefer something like `getFixedWidthBytes(int valIndex, int fixedWith)` so the user can pass any width between 1 and 32.

abonander:
Actually, `getRawValue()` already exists as the method you want; it will return the nth fixed width value which if not otherwise interpreted is basically the same as `bytes32`. Would you like to edit the docs for that method to explain that more clearly? Otherwise I don't see a reason to accept this PR.

JaimeSilver:
Hello Austin,

I tried all the options you got, and Solidity reverted. That is why I suggest putting the blocks. To be clear, I am sending a Bytes 32 but the specific field is only 5, so all conflict at some point. 

Jaime

JaimeSilver:
Yes, that is the idea. I decided to cast externally and pass the getBytes[...,32]

Please consider for future releases.

abonander:
@JaimeSilver I don't understand what you're trying to do.

JaimeSilver:
Hello Austin;

 

You can close 178. I used the generic getter provided getRawValue as per your suggestion.

 

I thought you were going to provide 1 getter for each type but the solution provided works fine.

 

Thank you, Jaime.

 

I need your help fixing src/main/java/com/hedera/hashgraph/sdk/CallParams.java <https://github.com/hashgraph/hedera-sdk-java/pull/184#discussion_r310743400>  eliminating the square brackets of the parameter name. It must be called bytes32 when sending to EVM:

Line 194-195 addParamType("bytes" + fixedLen );

 

Without this fix, EVM reverts… So no brackets, please.

 

Thank you for your help (and nice code)… 

 

 

Jaime Plata
Co-founder, All Over Chain

 <mailto:jp@aochain.net> jp@aochain.net


WhatsApp/Mobile : +1 (704) 3073702 

 

This email may contain material that is confidential, privileged and/or attorney work product for the sole use of the intended recipient.  Any review, reliance or distribution by others or forwarding without express permission is strictly prohibited.  If you are not the intended recipient, please contact the sender and delete all copies.

abonander:
@JaimeSilver you have a style error on #184 that's preventing it from being merged, otherwise it should be fine.

JaimeSilver:
Hi,

Sorry I missed this notification. I did the change.

Thanks, J

PR #184, opened 2019-08-03
Following up on issue #177.

JaimeSilver:
Incorrect tag for the type to the Solidity Interface. It needs to go as bytes32, not bytes[32]


src/main/java/com/hedera/hashgraph/sdk/CallParams.java
```diff
@@ -191,7 +191,8 @@ private static ByteString encodeDynArr(List<ByteString> elements, boolean prepen
                 "bytesN cannot have a length greater than 32; given length: " + fixedLen);
         }
 
-        addParamType("bytes[" + fixedLen + "]");
+        //addParamType("bytes[" + fixedLen + "]");
+        addParamType("bytes" + fixedLen );
```
abonander: There should be no space between `fixedLen` and the closing `)`.

PR #187, opened 2019-08-07
Added setMaxTransactionFee()

SimiHunjan:
Added setMaxTransactionFee()


abonander:
As per my comments on your other PRs, please just inline the value and use `_` as a thousands-separator.

PR #188, opened 2019-08-07
Update CreateFile.java

SimiHunjan:
Added txFee variable and setTransactionFee()


abonander:
As per my comments on your other PRs, please just inline the value and use `_` as a thousands-separator.

abonander:
merged in #207

PR #189, opened 2019-08-07
Update CreateAccount.java

SimiHunjan:
added txFee variable and setTransactionFee()


examples/src/main/java/com/hedera/hashgraph/sdk/examples/advanced/CreateAccount.java
```diff
@@ -17,11 +17,13 @@ public static void main(String[] args) throws HederaException {
         System.out.println("public key = " + newPublicKey);
 
         var client = ExampleHelper.createHederaClient();
-
+        
+        var txFee = 10000000;
```
abonander: This can just be inlined, it's not a magic number if it's only used once in an example.

abonander: Also if you would please, use `_` separators to make the number easier to read: `10_000_000`

abonander:
merged in #207

PR #190, opened 2019-08-07
Added transaction fees

SimiHunjan:


abonander:
You can just directly provide the values; I don't see any reason to have it as a separate variable. 

Also please use `_` as a thousands-separator to make the numbers easier to read, e.g. `1_000_000_000` instead of `1000000000`.

It might also be helpful to narrow down these values a little more but I wouldn't block this PR on it.

abonander:
merged in #207

PR #193, opened 2019-08-13
fixed typo in comment (habr -> hbar)

isiko-de:


abonander:
This might get overwritten the next time the protobufs are updated, but thanks.

PR #221, opened 2019-09-10
Improves CreateStatefulContract.java

gregscullard:
Updates the example with examples on how to work out actual transaction costs via a record, and query for query costs while updating the gas amounts to more reasonable amounts.

Also implements 3 examples of gas being too low, query fee too low and transaction fee too low.

Finally, the example as is fails with INSUFFICIENT_TX_FEE, this fixes this issue


mehcode:
@gregscullard I'm fairly sure the intention here is that instead of requesting cost everywhere you instead set a maximum payment amount and let the SDK request the exact cost for you.

See: https://github.com/hashgraph/hedera-sdk-java/blob/master/src/main/java/com/hedera/hashgraph/sdk/Client.java#L85

---

You are right though. The examples do need to be updated.

mehcode:
This has since been fixed in master. All examples should work as-is.

PR #250, opened 2019-11-09
implement retry with backoff for `BUSY` on `HederaCall.execute[Async]()`

### LINKED ISSUE #231: Client should retry when getting `BUSY` error with `.executeForReceipt()`
Assume a client is created with several nodes.
It turns out that upon using `.executeForReceipt()`, the client immediately fails with "BUSY" and doesn't retry.
According to Greg, `.execute()` and `.executeForRecord()` retry automatically, but I couldn't exactly test this or verify it.

Cheers guys!

abonander:
closes #231

* `Transaction.execute()` and `Transaction.executeAsync()` (and same methods on `TransactionBuilder` will retry if they get `BUSY` until the `validDuration` of the transaction elapses
* `TransactionReceiptQuery` (and thus `Transactinon.queryReceipt()` and `Transaction.queryReceiptAsync()`) will retry on `BUSY`, `OK` and `UNKNOWN` until 5 minutes elapses (maximum value for `validDuration` plus the 3 minutes that the receipt exists for)
* These timeouts can be overridden by the user by calling overloaded methods which take a `Duration` as their last argument; this only times-out the retry process itself, not any blocking calls internally, so may not be exact.

### DIFF
--- a/src/main/java/com/hedera/hashgraph/sdk/Backoff.java
+++ b/src/main/java/com/hedera/hashgraph/sdk/Backoff.java
@@ -1,0 +1,70 @@
+package com.hedera.hashgraph.sdk;
+
+import java.time.Duration;
+import java.time.Instant;
+import java.util.Optional;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.function.Consumer;
+import java.util.function.Predicate;
+
+final class Backoff {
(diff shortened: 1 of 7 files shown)


abonander:
cc @gregscullard @donaldthibeau @atul-hedera

PR #251, opened 2019-11-13
Refactor some objects into POJOs

abonander:
This refactors some objects to deprecate getters in favor of `public final` fields to improve usability and pave the way for supporting serialization with [Jackson](https://github.com/FasterXML/jackson). This also brings these classes more in line with their equivalents in the Javascript SDK.

* `AccountId`, `ContractId`, and `FileId` had their `getShardNum()` and `getRealmNum()` getters deprecated and `shard`/`realm` fields provided instead. Same with `getAccountNum()`/`getContractNum()`/`getFileNum()`, respectively.
* `Transaction.getTransactionId()` has been deprecated in favor of a new `id` field
* `AccountInfo`, `ContractInfo` and `FileInfo` have had all their getter methods deprecated in favor of public fields in the same fashion as the above.

### DIFF
--- a/src/main/java/com/hedera/hashgraph/sdk/SolidityUtil.java
+++ b/src/main/java/com/hedera/hashgraph/sdk/SolidityUtil.java
@@ -44,9 +44,9 @@ public final class SolidityUtil {
 
     public static String addressFor(ContractId contractId) {
         return addressForEntity(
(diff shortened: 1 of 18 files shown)


abonander:
cc @gregscullard @donaldthibeau @atul-hedera

PR #259, opened 2019-11-26
unify idempotent method naming

### LINKED ISSUE #255: Unify verbs for methods on Client, Transaction, and Query objects to `getXYZ`
We have `requestCost`, `queryReceipt`, `queryRecord`, `getAccountBalance`, etc.. We should only have `getCost`, `getReceipt`, `getAccountBalance`, etc.. These methods are all idempotent, that is the determination to me here.

abonander:
rename `Transaction.query*()` to `Transaction.get*()` 

rename `QueryBuilder.requestCost*()` to `QueryBuilder.getCost*()`

add appropriate deprecation messages for existing APIs

closes #255

### DIFF
--- a/src/main/java/com/hedera/hashgraph/sdk/QueryBuilder.java
+++ b/src/main/java/com/hedera/hashgraph/sdk/QueryBuilder.java
@@ -109,10 +109,24 @@ public abstract class QueryBuilder<Resp, T extends QueryBuilder<Resp, T>> extend
         return (T) this;
     }
 
+    public final long getCost() throws HederaException, HederaNetworkException {
+        return new CostQuery().execute();
+    }
+
+    public final void getCostAsync(Consumer<Long> withCost, Consumer<HederaThrowable> onError) {
+        new CostQuery().executeAsync(withCost, onError);
+    }
+
+    /**
+     * @deprecated renamed to {@link #getCost()}
+     */
     public final long requestCost() throws HederaException, HederaNetworkException {
         return new CostQuery().execute();
     }
 
+    /**
+     * @deprecated renamed to {@link #getCostAsync(Consumer, Consumer)}
+     */
     public final void requestCostAsync(Consumer<Long> withCost, Consumer<HederaThrowable> onError) {
         new CostQuery().executeAsync(withCost, onError);
     }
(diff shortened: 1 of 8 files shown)


abonander:
cc @gregscullard @atul-hedera

Motivation on #255:
> We have `requestCost`, `queryReceipt`, `queryRecord`, `getAccountBalance`, etc.. We should only have `getCost`, `getReceipt`, `getAccountBalance`, etc.. These methods are all idempotent, that is the determination to me here.

src/main/java/com/hedera/hashgraph/sdk/QueryBuilder.java
```diff
@@ -109,11 +109,25 @@ public T setPaymentDefault(long paymentAmount) {
...
+     * @deprecated renamed to {@link #getCost()}
+     */
+    public final long queryCost() throws HederaException, HederaNetworkException {
+        return new CostQuery().execute();
+    }
+
+    /**
+     * @deprecated renamed to {@link #getCostAsync(Consumer, Consumer)}
+     */
+    public final void queryCostAsync(Consumer<Long> withCost, Consumer<HederaThrowable> onError) {
```
mike-burrage-hedera: Why are there 2 new methods added that are deprecated?

src/main/java/com/hedera/hashgraph/sdk/Transaction.java
```diff
@@ -318,7 +338,7 @@ public final void executeForReceiptAsync(BiConsumer<Transaction, TransactionRece
      * while fetching a record.
      *
      * Call {@link #executeAsync(Consumer, Consumer)} then
-     * {@link #queryRecordAsync(Consumer, Consumer)} and handle the errors separately
+     * {@link #getRecordAsync(Consumer, Consumer)} and handle the errors separately
      * from each.
```
mike-burrage-hedera: Line 328 above, comment needs to be updated.

src/main/java/com/hedera/hashgraph/sdk/QueryBuilder.java
abonander: Initial refactor that I forgot to revert. Fixed.

src/main/java/com/hedera/hashgraph/sdk/Transaction.java
abonander: Actually these `BiConsumer` overloads we decided to not carry over since the user shouldn't be doing everything in a single method chain anyway, so these comments can just be removed.

PR #266, opened 2019-11-27
Remove transaction and query classes pertaining to claims

### LINKED ISSUE #258: Remove AddClaim example and Claim transaction/query types

abonander:
These classes are removed without a deprecation cycle as claims were never implemented on the platform and so anything using them would never have functioned to begin with:

* `AccountAddClaimTransaction`
* `AccountClaimQuery`
* `AccountDeleteClaimTransaction`

The `AccountInfo.getClaims()` getter and the `Claim` class itself are left intact as users could be inspecting that, but should have only ever observed an empty list. This removes the public `claims` field that was added in #251 as it was not yet released.

closes #258


abonander:
cc @gregscullard @atul-hedera for approval.

mike-burrage-hedera:
LGTM

PR #282, opened 2019-12-10
Move optional timeout to before the callback parameters

mehcode:
When reviewing, it was surprising that timeout came after the async callback parameters.

---

After we decide on the change, can you help me identify what functions need a deprecated equivalent.


abonander:
@mehcode the overloads with the timeouts haven't been released yet, right? shouldn't need deprecations

mehcode:
If that's the case, then good. it's just been awhile since the release. Haha.

PR #295, opened 2019-12-19
ConsensusClient: add `.keepAliveTime(2 minutes)` to prevent subscriptions from being silently dropped

### LINKED ISSUE #294: Topic subscriptions unreliable
**Detailed Description**
When subscribing to a topic via the java sdk (`0.9.0-beta-3`), all is well if the topic receives a frequent flow of transactions.
However if a topic is left idle (no transactions) for 30 minutes (not 100% sure on the actual time, but this seems to be repeatable), the subscriber receives no notifications until the subscription is restarted.

**Actual Behavior**
-Subscribe to a topic
-Send a transaction, witness the notification from mirror node
-Wait 30 minutes without sending a transaction
-Send a transaction, witness no notification from mirror node

**Expected Behavior**
-Lack of transactions should not prevent notifications from occurring if subscribed.

**Environment:**
- Java: [e.g. OpenJDK 11.0.4]
- Node: [e.g. v12.9.1-x86]
- OS: [e.g. Ubuntu 18.04]
- Version: [e.g. 1.0.1]

**Additional Context**
I have a client which drops the subscription every 2 minutes, then re-starts it. No messages have been lost after 6 hours and posting a new transaction once an hour.
Will try additional tests without dropping the subscription and providing a constant flow of transactions (every 30s) and a slow rate of transactions (every hour) to see if transaction throughput is necessary to keep the subscription "alive".

Note: Cross posting in the mirror-node project.
https://github.com/hashgraph/hedera-mirror-node/issues/454

abonander:
closes #294

cc @gregscullard

### DIFF
--- a/src/main/java/com/hedera/hashgraph/sdk/consensus/ConsensusClient.java
+++ b/src/main/java/com/hedera/hashgraph/sdk/consensus/ConsensusClient.java
@@ -1,11 +1,11 @@
 package com.hedera.hashgraph.sdk.consensus;
 
 import com.google.common.collect.Iterators;
-import com.hedera.hashgraph.sdk.Experimental;
-import com.hedera.hashgraph.sdk.TimestampHelper;
 import com.hedera.hashgraph.proto.mirror.ConsensusServiceGrpc;
 import com.hedera.hashgraph.proto.mirror.ConsensusTopicQuery;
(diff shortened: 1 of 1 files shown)


abonander:
cherry-picked onto `experiment/hcs`

PR #314, opened 2020-01-03
TransactionRecord API changes

abonander:
* deprecated `getTransfers()` in favor of field `transfers`
* (breaking-beta) renamed `memo` to `transactionMemo`


abonander:
@mehcode moving `Transfer` out of the class is a breaking change, should I add a deprecation notice?

mehcode:
Yes. Just in case they happen to be using that class definition.

PR #317, opened 2020-01-05
Fixed example

SimiHunjan:


examples/src/main/java/com/hedera/hashgraph/sdk/examples/advanced/DeleteFile.java
```diff
@@ -39,12 +39,10 @@ public static void main(String[] args) throws HederaException {
...
         byte[] fileContents = "Hedera hashgraph is great!".getBytes();
 
-        TransactionId txId = new FileCreateTransaction().setExpirationTime(
-            Instant.now()
-                .plus(Duration.ofSeconds(2592000)))
-            // Use the same key as the operator to "own" this file
+        TransactionId txId = new FileCreateTransaction()
             .addKey(OPERATOR_KEY.publicKey)
             .setContents(fileContents)
+            .setMaxTransactionFee(200_000_000)
```
abonander: We'd like to push people to use the new `Hbar` class, so this would be `Hbar.of(2)`.

SimiHunjan: Sweet - updated!

examples/src/main/java/com/hedera/hashgraph/sdk/examples/advanced/DeleteFile.java
```diff
@@ -64,6 +63,7 @@ public static void main(String[] args) throws HederaException {
 
         FileInfo fileInfo = new FileInfoQuery()
             .setFileId(newFileId)
+            .setPaymentAmount(Hbar.of(2))
```
mehcode: Why are you setting a explicit payment amount here? Is there a bug in the network? I ask because you **should** be able to set a _maximum_ with `.setMaxPaymentAmount(Hbar.of(2))` and then the SDK will dynamically ask the cost of the query and pay the exact amount as long as its under your configured maximum (the default being 1h).

SimiHunjan: I had an older sdk from a month ago so I just grabbed the latest one right now to check
 and I have options for setMaxQueryPayment, setQueryPayment, and depreciated setPaymentAmount. If I use setMaxQueryPayment it fails and returns an insufficient transaction fee response. If I use setQueryPayment or setPaymentAmount it is successful. Screenshots:

Screen Shot 2020-01-07 at 12 15 37 AM
Screen Shot 2020-01-07 at 12 16 07 AM

mehcode: That feels like a bug in the network and the response from the COST_ANSWER query. I'll open an issue to investigate this further. Do leave in the explicit payment amount for now.

abonander:
Excellent, this looks like it needs to be rebased though. @SimiHunjan do you want to do that? I can also take care of it if you're busy. We avoid the "Update Branch" button because it muddies the commit history.

examples/src/main/java/com/hedera/hashgraph/sdk/examples/advanced/DeleteFile.java
abonander: The network returns a `COST_ANSWER` of zero because the file is deleted.

SimiHunjan:
If you don't mind that would be great! Thank you!

PR #331, opened 2020-01-14
1.0 refactors

abonander:


src/main/java/com/hedera/hashgraph/sdk/Hbar.java
```diff
@@ -165,10 +181,11 @@ public long asTinybar() {
     public String toString() {
         // print with the original unit if we did a conversion, possibly more useful for debugging
         if (originalUnit != HbarUnit.Tinybar) {
-            return as(originalUnit) + " " + originalUnit + " (" + tinybar + " tinybar)";
+            return as(originalUnit) + " " + originalUnit.getSymbol() +
+                " (" + tinybar + " " + HbarUnit.Tinybar.getSymbol() + ")";
```
danielakhterov: Perhaps use `String.format()` here to increase readability?

abonander: Concatenation is technically faster, though it's not like it matters since this should only be used for debugging anyway.

PR #332, opened 2020-01-14
Hcs fixes and examples

### LINKED ISSUE #323: Update HCS code for final version of HCS protobuf
ConsensusClient.getMessages/getMessagesUntil and ConsensusMessageSubmitTransaction.setTopicID() use *topic* as param name, where other functions call it *topicId*. Please use the same param name for all references to this topicId.

@SimiHunjan can update documentation when these changes are in.

### LINKED ISSUE #327: HCS related updates/fixes
CreateTopic:
- set autorenew period to 90 days by default (it's the only required parameter and the period has the same restriction as for other entities)

UpdateTopic:
- explicitly add function to clear the topic's memo. Clearing the server-side topic's memo is done by sending memo as empty string (non-null StringValue with 0 length string).
- explicitly add function to clear the topic's admin/submit keys (which is done by sending an empty keylist or empty threshold key or empty non-null any kind of key actually)
- explicitly add function to clear the topic's autorenewaccount (which is done by sending account 0.0.0)

There should (eventually) be similar methods for clearing keys/memos/etc for other entities which may be done slightly differently in the protobuf itself (clearing contract's memo will be different as new fields will need to be added to do this). Some way to explicitly say that you want to remove attribute X from the server-side entity would be helpful.

DeleteTopic:
- builder is the builder for UpdateTopic, not DeleteTopic.

Please add javadoc to all these classes (and submitmessage)

~~There also needs to be a way to add additional signing keys to Update/Delete/Submit due to the admin/submitKey on the topic. It's not clear on how to do this (similar functionality should exist for files and contracts). Is it expected to be done via a distinct Client for each topic/file based on their admin/submit keys?~~ Looks like build().sign().execute() is the preferred method.

### LINKED ISSUE #333: IllegalStateException "Unchecked Key Case" in getInfo() calls on entities having threshold/keylist
Specifically I was testing ConsensusTopicInfoQuery() on the experiment/hcs branch for a topic with a threshold key and this exception was thrown upon execute()ing the query.

Likely affects info queries on file/contract/crypto where the entity has non-trivial keys (threshold or keylist).

I have a fix as part of updates I'm doing for #332 , just need to add some unit test cases and I'll update that PR.

mike-burrage-hedera:
- Fix ConsensusTopicDeleteTransaction.
- Set default entity lifetime for topic in ConsensusTopicCreateTransaction (and update account/contract create to use same constant).
- Update parameter name "topic" to "topicId" in several places for consistency.
- Add "clearXyz()" methods to ConsensusTopicUpdateTransaction for clearing various attributes on the topic entity.
- Add example - HCS pub-sub with submitKey.
- Add example - HCS topic adminKey as a 2-of-3 thresholdKey, then updated to a 3-of-4 adminKey via UpdateTopic.
- Fixes GetTopicInfo (switch statement not constructing TopicInfo on response)
- Fixes PublicKey not translating from KeyList and ThresholdKey protobufs

Fixes #327 (minus requirement to update a bunch of javadoc - I'll create another issue)
Fixes #323
Fixes #333

### DIFF
--- a/examples/src/main/java/com/hedera/hashgraph/sdk/examples/advanced/AdvancedExample.java
+++ b/examples/src/main/java/com/hedera/hashgraph/sdk/examples/advanced/AdvancedExample.java
@@ -1,0 +1,42 @@
+package com.hedera.hashgraph.sdk.examples.advanced;
+
+import com.hedera.hashgraph.sdk.Client;
+import com.hedera.hashgraph.sdk.account.AccountId;
+import com.hedera.hashgraph.sdk.crypto.ed25519.Ed25519PrivateKey;
+import io.github.cdimascio.dotenv.Dotenv;
+
+import java.util.Objects;
+
+/**
+ * Base class for advanced examples.
+ * Responsible for setting up the HAPI Client based on environment variable settings.
+ */
+class AdvancedExample {
+    private Client hapiClient;
+
+    public AdvancedExample() {
+        setupHapiClient();
+    }
+
+    public Client getHapiClient() {
+        return hapiClient;
+    }
+
+    private void setupHapiClient() {
+        // The Hedera Hashgrpah node's IP address, port, and account ID.
+        AccountId nodeID = AccountId.fromString(Objects.requireNonNull(Dotenv.load().get("NODE_ID")));
+        String nodeAddress = Objects.requireNonNull(Dotenv.load().get("NODE_ADDRESS"));
+
+        // Transaction payer's account ID and ED25519 private key.
+        AccountId payerId = AccountId.fromString(Objects.requireNonNull(Dotenv.load().get("OPERATOR_ID")));
+        Ed25519PrivateKey payerPrivateKey =
+            Ed25519PrivateKey.fromString(Objects.requireNonNull(Dotenv.load().get("OPERATOR_KEY")));
+
+        // Interface used to publish messages on the HCS topic.
+        hapiClient = new Client(nodeID, nodeAddress);
+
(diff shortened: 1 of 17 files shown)


examples/src/main/java/com/hedera/hashgraph/sdk/examples/advanced/AdvancedExample.java
```diff
@@ -0,0 +1,42 @@
...
+        Ed25519PrivateKey payerPrivateKey =
+            Ed25519PrivateKey.fromString(Objects.requireNonNull(Dotenv.load().get("OPERATOR_KEY")));
+
+        // Interface used to publish messages on the HCS topic.
+        hapiClient = new Client(nodeID, nodeAddress);
+
+        // Defaults the operator account ID and key such that all generated transactions will be paid for by this
+        // account and be signed by this key
+        hapiClient.setOperator(payerId, payerPrivateKey);
+    }
```
abonander: We actually inlined this code into the examples on purpose because we found that people were copying the examples verbatim including the `ExampleHelper` class and not really understanding how `Client` was constructed or how they get `Ed25519PrivateKey`. Having everything in the same class makes it much easier to explain the flow.

mehcode:
Thanks for the awesome work @mike-burrage-hedera. I merged this manually ( https://github.com/hashgraph/hedera-sdk-java/commits/hcs ) without any example changes as we do need to keep each example file self-contained. Can you please re-submit the examples in a new PR ?

PR #346, opened 2020-01-27
Add validation of HCS running hash along with example.

mike-burrage-hedera:
- Add class to validate HCS running hash
- Update an example to validate running hash
- Fix style issues in 2 existing HCS example classes

This should be part of release 1.2.0 (along with fragmentation/reassembly).


mehcode:
We try to avoid using non-semantic overloading to maintain parity between the SDKs. So the 3 constructors need to be separated out into 2 static methods and 1 constructor or 3 static methods.

Some API thoughts.

 * This feels a lot like HMAC to me so I'm going to suggest a lot of names from the HMAC family of crypto so we can have familiar terminology.

 * `ConsensusTopicMac` instead of `ConsensusTopicRunningHash` ?

 * Add `#verify(byte[] runningHash, long sequenceNumber)` - returns bool

 * `#update(byte[] runningHash, long sequenceNumber, Instant consensusTimestamp)` which mutates instead of `#getUpdatedRunningHash(byte[], Instant)` **and** calls `#verify` before it updates. Should throw an exception if verify fails.

---

Thoughts @mike-burrage-hedera ?

lbaird:
The running hash is not a MAC. Because a Message Authentication Code (MAC) is a way of mixing a secret key into something, in order to authenticate it to someone else who knows the secret. We aren’t mixing in a secret.

The running hash is actually what is called a “running hash” in the literature. It’s a hash that is updated for each object in a sequence by combining that the hash of that object with the previous running hash.

So ConsensusTopicRunningHash would be better than ConsensusTopicMAC.



________________________________

mehcode:
@lbaird I appreciate the insight, thank you.

---

My only reservation with the class being named `ConsensusTopicRunningHash` is it has a field named `runningHash`, which is the "current" hash.

```java
ConsensusTopicRunningHash#runningHash
```

Ideas:

 * `ConsensusTopicRunningHash#current`

 * `ConsensusTopicRunningHash#latest`

 * `ConsensusTopicRunningHash#state`

 * `ConsensusTopicRunningHash#digest`

lbaird:
If the field represents the current running hash, then I’d think either “current” or “currentRunningHash” would make sense.  Though “latest” or “latestRunningHash” would also be ok.



________________________________

mike-burrage-hedera:
It's not necessary current or latest.

It's whatever runninghash is associated with that sequenceNumber and topicId at some point in time.

I'd prefer `digest` for the name if we make a change for that field.

lbaird:
It’s not a digest of a message. And it’s not a digest of all the messages. So “running hash” is a more precise name for it.

But it does sound like it’s the current running hash, meaning that it is the latest version of the running hash, as of the point in time being considered by that object.


________________________________

mehcode:
@lbaird How does `ConsensusTopicState` as the class name and `runningHash` as the field name sound?

```
ConsensusTopicState#runningHash
```

instead of 

```
ConsensusTopicRunningHash#runningHash
```

lbaird:
Does ConsensusTopicState contain everything about its state, such as memo, expiration time, number of messages, running hash?

If so, then ConsensusTopicState is fine. Or even just ConsensusTopic.


________________________________

mehcode:
It only contains enough to calculate the running hash.
 * `runningHash`
 * `sequenceNumber`
 * `topicId`

On "update" the consensusTimestamp and the message bytes gets passed in.

---

For context, I'm really trying to avoid calling this class ConsensusTopicRunningHash as `ConsensusTopicRunningHash#runningHash` is stuttering and really awkward to look at.

mike-burrage-hedera:
Updated for mutability and changed class name to ConsensusTopicState.

mehcode:
We will be rebasing this and reworking it a bit with #401

danielakhterov:
Since this MR is really out of date now and we've released SDK v2 since this MR has been created, I figured it'd be better to make this an issue and close this PR.

Issue #350, opened 2020-02-02
Incorrect parameter encoding for solidity address in latest Hedera Java SDK


hualinyc:
I have discovered a potential bug in the Hedera Java SDK. This bug is related to solidity address encoding. I was working with a solidity contract its constructor has an address as argument. The contract was deployed on testnet with success. However, contract calls that reference that address failed. Based on my analysis, the root cause is the incorrect padding in the method com.hedera.hashgraph.sdk.contract.ContractFunctionParams.addAddress(String). To my knowledge, this bug has not been reported in https://github.com/hashgraph/hedera-sdk-java/issues, https://github.com/hashgraph/hedera-mirror-node/issues nor in https://discordapp.com/invite/FFb9YFX.

Bug details:
The account in question is "0.0.155271" and it's 20 byte solidity address in hex form is:
String addressInHex = "0000000000000000000000000000000000025e87";

I used the following code to create the contract:
TransactionId contractTxId = new ContractCreateTransaction()
.setBytecodeFileId(newFileId)
.setGas(100_000_000)
.setConstructorParams(
new ContractFunctionParams()
.addAddress(addressInHex))
.execute(client);

The constructor parameter (32 bytes, 64 hex characters) encoded with the following sdk method: com.hedera.hashgraph.sdk.contract.ContractFunctionParams.addAddress(String) in hex is:
0000000000000000000000000000000000025e87000000000000000000000000

which is incorrect due to padding to the right. The correct encoding (with left-hand side padding) should be:
0000000000000000000000000000000000000000000000000000000000025e87

In my test, when I did custom encoding with left-hand side padding, the same contract calls after contract deployment succeeded.

mehcode:
Thanks for reporting the issue. This was fixed in #351 and should be out soon.

PR #351, opened 2020-02-04
Fix `ContractFunctionParams.addAddress()`

danielakhterov:
#350 


mehcode:
Note: The Android example app failing is a known issue and it's not a bug in the SDK. It's a missing dep in maven. Something updated and we need to update some versions I imagine.

PR #352, opened 2020-02-07
fix `Transaction.executeAsync()` to auto-sign the transaction

abonander:
just like `.execute()` does

### DIFF
--- a/src/main/java/com/hedera/hashgraph/sdk/Transaction.java
+++ b/src/main/java/com/hedera/hashgraph/sdk/Transaction.java
@@ -2,8 +2,18 @@ package com.hedera.hashgraph.sdk;
 
 import com.google.protobuf.ByteString;
 import com.google.protobuf.InvalidProtocolBufferException;
-import com.hedera.hashgraph.proto.*;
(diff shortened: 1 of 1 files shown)


src/main/java/com/hedera/hashgraph/sdk/Transaction.java
```diff
@@ -2,8 +2,18 @@
...
+import com.hedera.hashgraph.proto.FileServiceGrpc;
 import com.hedera.hashgraph.proto.FreezeServiceGrpc;
+import com.hedera.hashgraph.proto.SignatureMap;
+import com.hedera.hashgraph.proto.SignatureMapOrBuilder;
+import com.hedera.hashgraph.proto.SignaturePair;
+import com.hedera.hashgraph.proto.SignaturePairOrBuilder;
+import com.hedera.hashgraph.proto.SmartContractServiceGrpc;
+import com.hedera.hashgraph.proto.TransactionBody;
+import com.hedera.hashgraph.proto.TransactionBodyOrBuilder;
+import com.hedera.hashgraph.proto.TransactionResponse;
```
abonander: IntelliJ auto-optimized these imports; I guess it doesn't like wildcards.

PR #356, opened 2020-02-11
Add new exception types for different sources of status errors

abonander:
These include additional context where available.

This solves the problem that `getRecord()` throws on the `getReceipt()` internal call even though records may contain useful information in the error case; in this case only `Busy` and `Unknown` are bubbled up and other status codes are ignored. If the transaction failed, the specific subclass of `HederaStatusException` will be `HederaRecordStatusException` which contains the record we fetched.

TODO:
- [x] override `getMessage()` in the subclasses to return more context

### DIFF
--- a/src/main/java/com/hedera/hashgraph/sdk/HederaPrecheckStatusException.java
+++ b/src/main/java/com/hedera/hashgraph/sdk/HederaPrecheckStatusException.java
@@ -1,0 +1,31 @@
+package com.hedera.hashgraph.sdk;
+
+import com.hedera.hashgraph.proto.ResponseCodeEnum;
+
+/**
+ * A {@link HederaStatusException}, thrown specifically by a transaction precheck (validation
+ * on the node it was submitted to).
+ */
+public class HederaPrecheckStatusException extends HederaStatusException {
+    /**
+     * The ID of the transaction that failed, in case that context is no longer available
+     * (e.g. the exception was bubbled up).
+     */
+    public final TransactionId transactionId;
+
+    HederaPrecheckStatusException(ResponseCodeEnum responseCode, TransactionId transactionId) {
+        super(responseCode);
+        this.transactionId = transactionId;
+    }
+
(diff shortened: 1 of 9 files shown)


src/main/java/com/hedera/hashgraph/sdk/HederaPrecheckStatusException.java
```diff
@@ -0,0 +1,26 @@
...
+     * The ID of the transaction that failed, in case that context is no longer available
+     * (e.g. the exception was bubbled up).
+     */
+    public final TransactionId transactionId;
+
+    HederaPrecheckStatusException(ResponseCodeEnum responseCode, TransactionId transactionId) {
+        super(responseCode);
+        this.transactionId = transactionId;
+    }
+
```
QuestofIranon: Needs getMessage override so it will display the transactionID and the responsecode.

src/main/java/com/hedera/hashgraph/sdk/HederaReceiptStatusException.java
```diff
@@ -0,0 +1,42 @@
...
+     * {@link TransactionReceipt#status}.
+     */
+    public final TransactionReceipt receipt;
+
+    HederaReceiptStatusException(ResponseCodeEnum responseCode, TransactionId transactionId, TransactionReceipt receipt) {
+        super(responseCode);
+        this.transactionId = transactionId;
+        this.receipt = receipt;
+    }
+
```
QuestofIranon: Needs getMessage override so it will display the transactionID and the responsecode.

src/main/java/com/hedera/hashgraph/sdk/HederaRecordStatusException.java
```diff
@@ -0,0 +1,40 @@
...
+     * logs).
+     */
+    public final TransactionRecord record;
+
+    HederaRecordStatusException(ResponseCodeEnum responseCode, TransactionRecord record) {
+        super(responseCode);
+        this.record = record;
+        this.transactionId = record.transactionId;
+    }
+
```
QuestofIranon: Needs getMessage override so it will display the transactionID and the responsecode.

PR #358, opened 2020-02-12
start integration testing by covering new exception types

abonander:
- [x] tested `HederaPrecheckStatusException` and found a bug (we were calling the wrong version of `.throwIfExceptional()`)
- [x] tested `HederaReceiptStatusException` (apparently `INVALID_SIGNATURE` errors aren't returned during precheck so that's a good one, can just copy the precheck test and fix the amounts)
- [ ] tested `HederaRecordStatusException` (maybe with a contract execute to test logging? we'll want to create and destroy a contract though which requires setup/teardown)

Actually running these tests on CI is blocked on getting a testnet account and private key for CI.

### DIFF
--- a/.github/workflows/java.yml
+++ b/.github/workflows/java.yml
@@ -70,6 +70,7 @@ jobs:
 
     - name: install SDK
       run: mvn install # implicitly runs tests
+      # TODO: turn on `integration-testing` profile when we have testnet account for CI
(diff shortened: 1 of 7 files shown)


mehcode:
As this has some fixes I'm going to merge. Please follow up with more tests.

PR #359, opened 2020-02-12
Update max size of transaction 4kb -> 6kb

SimiHunjan:
cc: @bmgentile


mike-burrage-hedera:
6144 bytes specifically.

SimiHunjan:
Thanks @mike-burrage-hedera updated.

PR #368, opened 2020-02-25
Mnemonic: add validate() method and remove BIP-39 lib

abonander:
Motivation: https://github.com/hashgraph/hedera-sdk-js/issues/210

Additionally, the BIP-39 library we were using is GPL-licensed which is incompatible with our Apache 2.0 license. I replaced it with a from-scratch implementation.

(Eventually) equivalent implementation in JS SDK: https://github.com/hashgraph/hedera-sdk-js/pull/211


src/main/java/com/hedera/hashgraph/sdk/crypto/Mnemonic.java
```diff
@@ -35,18 +37,68 @@
...
+
+    /**
+     * Construct a mnemonic from a 24-word list.
+     *
+     * @throws BadMnemonicError if {@code words.size() != 24} or if any words in the list do not
+     * exist in the BIP-39 standard English word list (as typo detection).
+     *
+     * @param words the 24-word list that constitutes a mnemonic phrase.
+     */
+    public Mnemonic(List<? extends CharSequence> words) {
```
abonander: I'm decently sure this is a forward-compatible change (`List<CharSequence>` -> `List<? extends CharSequence>`) but it makes the constructor much nicer to use, e.g. if you have a `List<String>` (which won't automatically coerce to `List<CharSequence>`).

PR #383, opened 2020-04-01
implement periodic node health checks in `Client.pickNode()`

abonander:
* implement `Client.ping()`

Tested with the `CreateAccount` example with the network disconnected, it throws the expected exception:

```
Exception in thread "main" com.hedera.hashgraph.sdk.HederaNetworkException: all nodes failed their health checks
```

Tested again with the network reconnected and the example completes normally.


mehcode:
This was moved to `develop`

PR #392, opened 2020-04-27
fix: allow last node to be picked

Sheng-Long:


danielakhterov:
Looks good to me. Just waiting for CI

PR #419, opened 2020-09-29
update example

SimiHunjan:


SimiHunjan:
This was based in using the wrong version for testing.

PR #421, opened 2020-10-06
example update

SimiHunjan:


mehcode:
Thanks. We try to keep `examples/` compiling in JDK8. But it can be hard to remember to check as we use JDK14 to develop.

PR #460, opened 2021-03-11
Rename Executable retry limit field, getter and setter

### LINKED ISSUE #459: Rename `Executable.setMaxRetry()` -> `Executable.setMaxAttempts()`

Regan-Koopmans:
This PR resolves #459 

These changes are entirely semantic and replace `Retry` with `Attempt`, since this more closely captures the intent of this code. We are keeping track of the limit/count of retries, rather than a reference to any specific retry.

### DIFF
--- a/sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
+++ b/sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
@@ -25,6 +25,22 @@ abstract class Executable<SdkRequestT, ProtoRequestT, ResponseT, O> implements W
     Executable() {
     }
 
+    /**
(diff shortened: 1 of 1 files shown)


danielakhterov:
Thanks for the MR @Regan-Koopmans , but unfortunately we cannot simply replace `[set|get]MaxRetry()`, we have to keep both. I should have specified that in the issue; since we have released a major version change we need to deprecate the old methods and add the new methods, while maintaining functionality for both.

Regan-Koopmans:
@danielakhterov That makes complete sense. I have reintroduced the methods with deprecation annotations.

danielakhterov:
Fantastic work, thanks again @Regan-Koopmans for the PR.

PR #490, opened 2021-05-14
feat: Schedule Create Transaction now prints even if not signed, fixe…

andrix10:
…d scheduled Topic Message Submit test

Signed-off-by: Andrei Kuzmiankov <andrei@launchbadge.com>

### DIFF
--- a/sdk/src/integrationTest/java/ScheduleCreateIntegrationTest.java
+++ b/sdk/src/integrationTest/java/ScheduleCreateIntegrationTest.java
@@ -4,6 +4,8 @@ import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
 
+import java.nio.charset.StandardCharsets;
+import java.time.Instant;
 import java.util.Collections;
 import java.util.Objects;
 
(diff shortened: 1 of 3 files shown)


sdk/src/integrationTest/java/ScheduleCreateIntegrationTest.java
```diff
@@ -185,104 +187,98 @@ void canSignSchedule2() {
...
+                .getReceipt(testEnv.client);
+
+            info = new ScheduleInfoQuery()
+                .setScheduleId(scheduleId)
+                .setNodeAccountIds(testEnv.nodeAccountIds)
+                .execute(testEnv.client);
+
+            assertNotNull(info.executedAt);
+
+            System.out.println(info.executedAt.toString());
```
danielakhterov, line 280: Should not have prints in tests.

sdk/src/main/java/com/hedera/hashgraph/sdk/ScheduleCreateTransaction.java
```diff
@@ -94,4 +94,11 @@ boolean onFreeze(TransactionBody.Builder bodyBuilder) {
     void onScheduled(SchedulableTransactionBody.Builder scheduled) {
         throw new IllegalStateException("Cannot schedule `ScheduleCreateTransaction`");
     }
+
+    @Override
+    @SuppressWarnings("LiteProtoToString")
+    public String toString() {
```
danielakhterov, line 100: We shouldn't be overriding this method, but instead updating the current `Transaction.toString()` to work even if transaction is not frozen.

danielakhterov (approved):

PR #492, opened 2021-05-18
feat: Added token decimal support to AccountBalance

andrix10:
Signed-off-by: Andrei Kuzmiankov <andrei@launchbadge.com>

### DIFF
--- a/sdk/src/integrationTest/java/AccountBalanceIntegrationTest.java
+++ b/sdk/src/integrationTest/java/AccountBalanceIntegrationTest.java
@@ -120,6 +120,7 @@ class AccountBalanceIntegrationTest {
                 .setTokenName("ffff")
                 .setTokenSymbol("F")
                 .setInitialSupply(10000)
+                .setDecimals(50)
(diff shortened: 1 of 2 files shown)


sdk/src/integrationTest/java/AccountBalanceIntegrationTest.java
```diff
@@ -134,6 +135,7 @@ void canFetchTokenBalancesForClientOperator() {
                 .execute(testEnv.client);
 
             assertEquals(balance.token.get(tokenId), 10000);
+            assertEquals(balance.decimal.get(balance.token.get(tokenId)), 50);
```
mehcode, line 138: @danielakhterov 

JS has `balance.tokens`. We should deprecate `.token` here and have `.tokens` and `.tokenDecimals`.

danielakhterov: Yea, that makes sense. Should double check that Go has `.Tokens` and not `.Token` as well.

danielakhterov (approved):

PR #501, opened 2021-06-01
Show warning during transaction retries

steven-sheehy:
Summary:
- Add a retry log for non-gRPC errors
- Change retry log level to warning
- Fix build failing due to depending on unstable master branch of jabel plugin
- Fix using verbose multi-line logs and printing a stacktrace

Details:
When the SDK is retrying a failed transaction, there is no logging by default to indicate there is an issue. With the default maxAttempts, that could mean up to almost a minute (~48s) could go by without any feedback to the user that something is wrong. They could increase the log levels to trace, but then the logs would become too verbose with request and response objects being printed.

It looks like the logs used to print at error level until very recently, most likely because they were too verbose due to using both a multi-line format and including a stacktrace. I think using multi-line logs should be avoided if at all possible as it's too verbose, harder to read logs next to it and difficult to parse by log collectors like Loki. Stacktrace should not be printed until the final attempt fails.

### DIFF
--- a/sdk/build.gradle
+++ b/sdk/build.gradle
@@ -17,7 +17,7 @@ configurations {
 
 dependencies {
 	// https://github.com/bsideup/jabel
-    annotationProcessor 'com.github.bsideup.jabel:jabel-javac-plugin:0.3.0'
+	annotationProcessor 'com.github.bsideup.jabel:jabel-javac-plugin:0.3.0'
 
 	// custom annotation processor to help generate methods around executables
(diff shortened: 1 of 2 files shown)


danielakhterov (approved):

danielakhterov:
Changes look good, but this does require a rebase since I fixed the Jabel issues on `master`; I didn't realize you fixed it here as well.

danielakhterov (approved):

PR #503, opened 2021-06-01
Various TopicMessageQuery fixes

### LINKED ISSUE #502: TopicMessageQuery fixes and improvements
In using the SDK to subscribe to the mirror node gRPC API (which I'm the architect), I've encountered a lot of issues related to its implementation:

- Retry does not update startTime. This will cause duplicate messages to be delivered to the client.
- Retry does not update limit. This will cause extra messages to be delivered to the client beyond the initial specified limit.
- No option to disable or lower retry
- No notification that the server completed the stream when limit or endTime is reached
- No indication to user that subscription has failed and is in the process of retrying
- No max backoff. Retry attempts can wait up to 4 minutes between attempts.
- Thread interrupts inappropriately ignored by library
- Not retried for transient [resource exhausted](https://github.com/hashgraph/hedera-mirror-node/blob/8dbb54f733af3ff0e7baff2ea37a5614124b5c27/hedera-mirror-grpc/src/main/java/com/hedera/mirror/grpc/controller/ConsensusController.java#L109) errors. This can occur if the db or app instance is temporarily overloaded.
- Not retried for low level stream reset errors. This can occur if a proxy/load balancer goes down while the API stays up.
- No option to customize which errors should be retried
- Error handler not called for non-retryable StatusRuntimeExceptions
- Error handler should not be called for application exceptions that occur during `onNext()`. The application can already handle/catch those since `onNext()` is application code.
- Should not throw `java.lang.Error` on retries exhausted. Errors should be reserved for unrecoverable system problems and are/should not normally caught by applications
- No default error handler that logs the error
- No tests
- Mirror node grpc channel not customizable to support in-process testing

PR incoming to address all but the last two

steven-sheehy:
Fixes #502 

This class has zero unit tests and I don't have the bandwidth to add that test coverage at this time. It would be nice if the SDK team could look into adding some in the future. You can do something [similar](https://github.com/hashgraph/hedera-mirror-node/blob/8dbb54f733af3ff0e7baff2ea37a5614124b5c27/hedera-mirror-monitor/src/test/java/com/hedera/mirror/monitor/subscribe/grpc/GrpcClientSDKTest.java#L78) to what we do in the mirror node by using a stub for the mirror node gRPC API.

I would also recommend you allow the mirror node's grpc channel to support the [in-process channel builder](https://grpc.github.io/grpc-java/javadoc/io/grpc/inprocess/InProcessChannelBuilder.html) to reduce brittle timeout issues that can occur when using a real grpc channel in constrained CI environments.

### DIFF
--- a/sdk/src/main/java/com/hedera/hashgraph/sdk/TopicMessageQuery.java
+++ b/sdk/src/main/java/com/hedera/hashgraph/sdk/TopicMessageQuery.java
@@ -1,29 +1,43 @@
 package com.hedera.hashgraph.sdk;
 
-import com.hedera.hashgraph.sdk.proto.TransactionID;
-import com.hedera.hashgraph.sdk.proto.mirror.ConsensusServiceGrpc;
-import com.hedera.hashgraph.sdk.proto.mirror.ConsensusTopicQuery;
-import com.hedera.hashgraph.sdk.proto.mirror.ConsensusTopicResponse;
 import io.grpc.CallOptions;
 import io.grpc.ClientCall;
 import io.grpc.Status;
 import io.grpc.StatusRuntimeException;
 import io.grpc.stub.ClientCalls;
 import io.grpc.stub.StreamObserver;
-import java8.util.function.BiConsumer;
-import java8.util.function.Consumer;
-import org.threeten.bp.Instant;
-
-import javax.annotation.Nullable;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Objects;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+import java.util.function.Predicate;
+import java.util.regex.Pattern;
+import javax.annotation.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.threeten.bp.Instant;
+
+import com.hedera.hashgraph.sdk.proto.TransactionID;
+import com.hedera.hashgraph.sdk.proto.mirror.ConsensusServiceGrpc;
+import com.hedera.hashgraph.sdk.proto.mirror.ConsensusTopicQuery;
+import com.hedera.hashgraph.sdk.proto.mirror.ConsensusTopicResponse;
 
 public final class TopicMessageQuery {
-    private final ConsensusTopicQuery.Builder builder;
 
-    @Nullable
-    private BiConsumer<Throwable, TopicMessage> errorHandler;
+    private static final Logger LOGGER = LoggerFactory.getLogger(TopicMessageQuery.class);
+    private static final Pattern RST_STREAM = Pattern
+            .compile(".*(rst.stream.*internal.error|internal.error.*rst.stream).*",
+                    Pattern.CASE_INSENSITIVE | Pattern.DOTALL);
+
+    private final ConsensusTopicQuery.Builder builder;
+    private Runnable completionHandler = this::onComplete;
+    private BiConsumer<Throwable, TopicMessage> errorHandler = this::onError;
+    private int maxAttempts = 10;
+    private Duration maxBackoff = Duration.ofSeconds(10L);
+    private Predicate<Throwable> retryHandler = this::shouldRetry;
 
     public TopicMessageQuery() {
         builder = ConsensusTopicQuery.newBuilder();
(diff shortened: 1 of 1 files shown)


sdk/src/main/java/com/hedera/hashgraph/sdk/TopicMessageQuery.java
```diff
@@ -1,29 +1,43 @@
...
+import com.hedera.hashgraph.sdk.proto.mirror.ConsensusTopicResponse;
 
 public final class TopicMessageQuery {
-    private final ConsensusTopicQuery.Builder builder;
 
-    @Nullable
-    private BiConsumer<Throwable, TopicMessage> errorHandler;
+    private static final Logger LOGGER = LoggerFactory.getLogger(TopicMessageQuery.class);
+    private static final Pattern RST_STREAM = Pattern
+            .compile(".*(rst.stream.*internal.error|internal.error.*rst.stream).*",
```
steven-sheehy, line 32: This matches both okhttp and netty's format for indicating a stream reset has occurred.

sdk/src/main/java/com/hedera/hashgraph/sdk/TopicMessageQuery.java
```diff
@@ -1,29 +1,43 @@
...
+                    Pattern.CASE_INSENSITIVE | Pattern.DOTALL);
+
+    private final ConsensusTopicQuery.Builder builder;
+    private Runnable completionHandler = this::onComplete;
+    private BiConsumer<Throwable, TopicMessage> errorHandler = this::onError;
```
steven-sheehy, line 37: This could be changed to a `Consumer<Throwable>` since a non-null `TopicMessage` will never be passed to it now with the change to not call the error handle for application errors during `onNext()`. But I didn't want to break API.

danielakhterov: One question that comes to mind, how will the user know their code erred in the callback? We had an issue with another project awhile ago where the application code was faulty and died within the callback; the user wasn't informed of this and didn't know if it was their code, or the SDK's code that was faulty, but all the knew is the callback stopped being called.

Another thing to mention is that I know there are some project that already use this specific feature, changing this behavior will break their code at runtime when the update to the latest SDK. So, although this isn't a API breaking change, I think it's fair to say this is still a breaking change.

steven-sheehy:
Just noticed a major bug that my code did not resolve. On retry, the start time in the grpc request is not actually updated. There's a variable that's updated but it's never used. Putting back to draft to work on a fix.

sdk/src/main/java/com/hedera/hashgraph/sdk/TopicMessageQuery.java
steven-sheehy: > how will the user know their code erred in the callback?

That's the thing, it's their code. They should be catching the error and handling it.

But the breaking compatibility is a fair point. So I tested it and I think it doesn't break compatibility. If you throw your own exception inside the custom `onNext()` handler, it will cause the stream to cancel and wrap that user's exception in a `io.grpc.StatusRuntimeException: CANCELLED: Failed to read message.` That exception will propagate to the `onError()`. Since our retryHandler doesn't retry status codes of `CANCELLED` by default it will in turn call the `errorHandler`. So effectively it is the same behavior of calling the `errorHandler` without explicitly calling, resulting in cleaner code. Please test it yourself to confirm, though.

steven-sheehy:
@danielakhterov This is ready for review. I did manual testing myself for the scenarios listed in the related issue, but please test it out to make sure I didn't mess anything up.

sdk/src/main/java/com/hedera/hashgraph/sdk/TopicMessageQuery.java
```diff
@@ -1,110 +1,199 @@
...
     }
 
     public TopicMessageQuery setEndTime(Instant endTime) {
+        Objects.requireNonNull(endTime, "endTime must not be null");
         builder.setConsensusEndTime(InstantConverter.toProtobuf(endTime));
         return this;
     }
 
     public TopicMessageQuery setLimit(long limit) {
+        if (limit < 0) {
```
danielakhterov, line 66: We try to avoid doing parameter validation on the client side because the server requirements have often changed; I think it'd be best to simply allow the user to set the limit to any value, and let the server respond with an error instead.

sdk/src/main/java/com/hedera/hashgraph/sdk/TopicMessageQuery.java
```diff
@@ -1,110 +1,199 @@
...
-        // startTime must be `final` or `effectively final` to be used within closures.
-        Instant[] startTime
-    ) {
-        if (attempt > 10) {
-            if (errorHandler != null) {
-                errorHandler.accept(new Error("Failed to connect to mirror node"), null);
-                return;
-            }
+    private void onComplete() {
+        TopicId topicId = TopicId.fromProtobuf(builder.getTopicID());
```
danielakhterov, line 108: Can we use `var` instead of consistency's sake; I'm just going to make one comment here, but there are other places where this is also the case.

danielakhterov (changes requested):

sdk/src/main/java/com/hedera/hashgraph/sdk/TopicMessageQuery.java
steven-sheehy: Sure, switched most variables I added to `var`.

sdk/src/main/java/com/hedera/hashgraph/sdk/TopicMessageQuery.java
steven-sheehy: Removed

danielakhterov (approved):

sdk/src/main/java/com/hedera/hashgraph/sdk/TopicMessageQuery.java
mehcode: This is definitely still a breaking change that the exception is being wrapped. 

> That's the thing, it's their code. They should be catching the error and handling it.

The behavior here was modeled after how promises work in JavaScript. If you throw an exception in the resolver it calls the same error handler as if the promise initially resolved with an exception.

In languages that don't have exceptions, where does the error go? The error has to be returned from `onNext` or handled in `onNext`. In the latter, the user would always need to write a second function, that is called in `onNext` to handle their errors.

Issue #471, opened 2021-04-05
SDK should sign transactions on demand


xin-hedera:
By default, when executing a transaction, SDK picks 1/3 nodes to construct signed transaction with. Each of these transaction bytes is signed though in most cases only one is needed. Private key signing is resource intensive, we should optimize to sign the transactions on demand.

mehcode:
- [ ] `Client.setOperatorPersistent(true)` that will mark the operator as **persistent**, meaning that its always there and we can sign on-demand instead of needing to sign up front

 - [ ] When using `Client.setOperator`, default `operatorPersistent` to `true`, otherwise `false`.

---

cc @danielakhterov

danielakhterov:
This change should affect more than just the operator @mehcode. Perhaps we have `Transaction.sign[With|WithOperator]Persistent()` and/or `Transaction.persistantSigning(true)` which is `true` by default?

PR #513, opened 2021-06-10
Sign on demand

danielakhterov:
Resolves: #471

### DIFF
--- a/sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
+++ b/sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
@@ -42,7 +42,6 @@ public final class Client implements AutoCloseable {
 
     @Nullable
     private Operator operator;
-    private boolean signonDemand = true;
 
     Duration requestTimeout = Duration.ofMinutes(2);
 
@@ -363,15 +362,6 @@ public final class Client implements AutoCloseable {
 
         this.operator = new Operator(accountId, publicKey, transactionSigner);
         return this;
-    }
-
-    public synchronized Client setSignonDemand(boolean signonDemand) {
-        this.signonDemand = signonDemand;
-        return this;
-    }
-
-    public boolean getSignonDemand() {
-        return signonDemand;
     }
 
     /**
--- a/sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
+++ b/sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
@@ -42,6 +42,7 @@ public final class Client implements AutoCloseable {
 
     @Nullable
     private Operator operator;
+    private boolean operatorPersistent = true;
 
     Duration requestTimeout = Duration.ofMinutes(2);
 
@@ -362,6 +363,15 @@ public final class Client implements AutoCloseable {
 
         this.operator = new Operator(accountId, publicKey, transactionSigner);
         return this;
+    }
+
+    public synchronized Client setOperatorPersistent(boolean operatorPersistent) {
+        this.operatorPersistent = operatorPersistent;
+        return this;
+    }
+
+    public boolean getOperatorPersistent() {
+        return operatorPersistent;
     }
 
     /**
--- a/sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
+++ b/sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
@@ -364,6 +364,11 @@ public final class Client implements AutoCloseable {
         return this;
     }
 
+    public synchronized Client setMaxNodesPerTransaction(int maxNodesPerTransaction) {
+        this.network.setMaxNodesPerTransaction(maxNodesPerTransaction);
+        return this;
+    }
+
     /**
      * Get the ID of the operator. Useful when the client was constructed from file.
      *
--- a/sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
+++ b/sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
@@ -42,7 +42,7 @@ public final class Client implements AutoCloseable {
 
     @Nullable
     private Operator operator;
-    private boolean operatorPersistent = true;
+    private boolean signonDemand = true;
 
     Duration requestTimeout = Duration.ofMinutes(2);
 
@@ -365,13 +365,13 @@ public final class Client implements AutoCloseable {
         return this;
     }
 
-    public synchronized Client setOperatorPersistent(boolean operatorPersistent) {
-        this.operatorPersistent = operatorPersistent;
+    public synchronized Client setSignonDemand(boolean signonDemand) {
+        this.signonDemand = signonDemand;
         return this;
     }
 
-    public boolean getOperatorPersistent() {
-        return operatorPersistent;
+    public boolean getSignonDemand() {
+        return signonDemand;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
```diff
@@ -348,6 +349,15 @@ public synchronized Client setOperatorWith(AccountId accountId, PublicKey public
...
 
+    public synchronized Client setSignonDemand(boolean signonDemand) {
+        this.signonDemand = signonDemand;
+        return this;
+    }
+
+    public boolean getSignonDemand() {
+        return signonDemand;
+    }
+
```
Sean-Tedrow-LB, line 360: Is there any reason why the user of the SDK would want to NOT sign-on-demand?  Is there a reason the user of the SDK has a say in this?  This seems like it should be an under-the-hood change.

sdk/src/main/java/com/hedera/hashgraph/sdk/Transaction.java
```diff
@@ -792,16 +819,62 @@ public T freezeWith(@Nullable Client client) {
...
+                    .setSignedTransactionBytes(
+                        signedTransactions.get(i)
+                            .setSigMap(signatures.get(i))
+                            .build()
+                            .toByteString()
+                    ).build());
+            }
+
+            return;
+        }
```
Sean-Tedrow-LB, line 841: It seems like this should be factored into its own method, something like `buildAllTransactions()`, which then iterates and calls `buildTransaction(i)`

sdk/src/main/java/com/hedera/hashgraph/sdk/Transaction.java
```diff
@@ -792,16 +819,62 @@ public T freezeWith(@Nullable Client client) {
...
+
+        signTransaction(index);
+
+        transactions.add(com.hedera.hashgraph.sdk.proto.Transaction.newBuilder()
+            .setSignedTransactionBytes(
+                signedTransactions.get(index)
+                    .setSigMap(signatures.get(index))
+                    .build()
+                    .toByteString()
+            ).build());
```
Sean-Tedrow-LB, line 857: Should we check whether the transaction in question has already been built before building it?

Sean-Tedrow-LB (changes requested):

mehcode:
I've thought about this some more. 

- `.sign(key)` should temporarily store the private key on the transaction and sign on-demand, making sure to wipe the key after execute
- `.signWith(..)` should temporarily store the closure (same as above)
- `.signWithImmediate(..)` should sign ALL transactions

```js
tx.freezeWith(client); // freezes transaction and builds transaction index 0 only

await tx.sign(key); // signs transaction index 0, stores for later use in case of retries
await tx.signWith(...); // signs transaction index 0, stores for later use in case of retries
await tx.signWithImmediate(...); // builds remainder of transactions, signs all with this 
 // closure in one loop, does not sign with previous two signing methods 
 // until the transaction of the index is executed
```

`signWithImmediate` is for hardware wallet support where we ask the device to sign a block of "the same" transactions at once to support retries, as signing on a retry is illegal in a hardware wallet scenario

additionally, we should have

```js
client.setOperatorImmediate(true); // use signWithImmediate when signing for operator
```

xin-hedera (changes requested):

sdk/src/main/java/com/hedera/hashgraph/sdk/Transaction.java
```diff
@@ -792,16 +794,73 @@ public T freezeWith(@Nullable Client client) {
...
+        }
+    }
+
+    /**
+     * Will build the specific transaction at {@code index} and will fill with `null` for any empty indices before it
+     *
+     * @param index
+     */
+    void buildTransaction(int index) {
+        if (transactions.size() < index) {
```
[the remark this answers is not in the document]
danielakhterov: With this change if we were to have an empty list `[]` for `transactions` the size would be `0` and the index would be `0`. This conditional would pass since `index + 1` is `1` and `transaction.size()` is `0`. In the for loop within this conditional 
```java
for (var i = transactions.size(); i <= index; ++i) {
```
`i` would be `0` and `index` would be `0` and since we use the `<=` operator this for loops condition would pass resulting in us adding a `null` as the first `transaction`. 
Also, out of curiosity I made this change locally and ran the tests. The test quickly erred with `INVALID_TRANSACTION_BODY`

sdk/src/main/java/com/hedera/hashgraph/sdk/Transaction.java
```diff
@@ -49,6 +48,9 @@
     // ]
     int nextTransactionIndex = 0;
 
+    private List<PublicKey> publicKeys = new ArrayList<>();
```
mehcode, line 51: Can we make this a Set instead? Would remove the looped equality check in "already signed".

danielakhterov: We need this to be ordered the same as `signers` since so we know the signer for `publicKeys.get(i)` is `signers.get(i)`. We _can_ add another field which stores the public keys in a set, and only use that set for the already signed check.

sdk/src/main/java/com/hedera/hashgraph/sdk/Transaction.java
```diff
@@ -803,16 +805,73 @@ public T freezeWith(@Nullable Client client) {
...
+            ).build());
+    }
+
+    /**
+     * Will sign the specific transaction at {@code index} and will fill with `null` for any empty indices before it
+     *
+     * @param index
+     */
+    void signTransaction(int index) {
+        if (signatures.get(index).getSigPairCount() != 0) {
```
Sean-Tedrow-LB, line 849: It's possible I'm being dumb and missing something obvious, but when do the null signatures get added to the `signatures` list?  It looks like the very first thing this function does is `signatures.get(index)`, but the doc comment seems to imply that it should be adding nulls to the list from `signatures.size()` up to `index` first.

sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
```diff
@@ -352,6 +352,11 @@ public synchronized Client setOperatorWith(AccountId accountId, PublicKey public
         return this;
     }
 
+    public synchronized Client setMaxNodesPerTransaction(int maxNodesPerTransaction) {
+        this.network.setMaxNodesPerTransaction(maxNodesPerTransaction);
+        return this;
+    }
```
Sean-Tedrow-LB, line 358: Should the `maxNodesPerTransaction` on the client be persistent?  EG, if you `setNetwork()` to a new network after setting `maxNodesPerTransaction`, should the `maxNodesPerTransaction` you set on the client carry over to that new network?  That's the behavior I'd probably intuitively expect.  It LOOKS like this sets a property of the client, but it actually doesn't at all, it sets a property of the network that it _happens to be using_ at that time.

Is this just a convenience function to use while building a client so users don't need to `client.network.setMaxNodesPerTransaction()` on a separate line?  I'm not sure if the potential for confusion is worth that.  I would remove it entirely, I think.

Sean-Tedrow-LB (approved):

Sean-Tedrow-LB (dismissed):

Sean-Tedrow-LB (approved):

danielakhterov:
@Sean-Tedrow-LB approved, and is part of the SDK team so even though GitHub is complaining about missing an approval, I'll still merge.

PR #518, opened 2021-06-12
docs: Added SDK_ANATOMY.md

Sean-Tedrow-LB:
Signed-off-by: Sean-Tedrow-LB <sean.tedrow@launchbadge.com>

### DIFF
--- a/SDK_ANATOMY.md
+++ b/SDK_ANATOMY.md
@@ -1,0 +1,314 @@
+## Tools and Libraries:
+
+Gradle plugins are used for including libraries and tools.
+
+### Protobufs:
+
+Files with extension `.proto` get compiled into classes used for serializing and deserializing data, which can be used to save data locally, or send data over a channel.
+
+### gRPC:
+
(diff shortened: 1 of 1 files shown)


danielakhterov:
This looks great, but can we clearly separate intent of the SDK vs how the Java SDK accomplishes that intent.

SDK_ANATOMY.md
```diff
@@ -0,0 +1,330 @@
...
+
+`ManagedNode` has the methods `inUse()`, which causes the `ManagedNode` to record that it is being used, `getChannel()`, `close()`, and `getUserAgent()`.
+
+The user agent is a string that is used to identify the client to the server.  In this case, it's `"hedera-sdk-java/v{NUMBER}"`.
+
+
+
+
+
+### `Executable`:
```
danielakhterov, line 124: This is what I was thinking with separating intent from java specific implementation
```markdown
Intent:

Extract common code from all requests that retries the request on different nodes with node deprioritization, exponential back-off, and response code handling.

How we acheive this in the Java SDK:

We create several `abstract` methods which must be overwritten so `Executable` can work correctly.

<list of abstract methods and what they're used for>
```

PR #520, opened 2021-06-15
chore: getting rid of some compiler warnings

Sean-Tedrow-LB:
Signed-off-by: Sean-Tedrow-LB <sean.tedrow@launchbadge.com>

### DIFF
--- a/sdk/src/main/java/com/hedera/hashgraph/sdk/ChunkedTransaction.java
+++ b/sdk/src/main/java/com/hedera/hashgraph/sdk/ChunkedTransaction.java
@@ -121,7 +121,8 @@ abstract class ChunkedTransaction<T extends ChunkedTransaction<T>> extends Trans
         return transactionHashes;
     }
 
-    @Override @FunctionalExecutable(type = "java.util.List<TransactionResponse>")
+    @Override 
+    @FunctionalExecutable(type = "java.util.List<TransactionResponse>")
     public CompletableFuture<List<com.hedera.hashgraph.sdk.TransactionResponse>> executeAllAsync(Client client) {
         if (!isFrozen()) {
             freezeWith(client);
--- a/sdk/src/main/java/com/hedera/hashgraph/sdk/ChunkedTransaction.java
+++ b/sdk/src/main/java/com/hedera/hashgraph/sdk/ChunkedTransaction.java
@@ -9,9 +9,15 @@ import com.hedera.hashgraph.sdk.proto.TransactionID;
 import java8.util.concurrent.CompletableFuture;
 import java8.util.concurrent.CompletionStage;
 import java8.util.function.Function;
+import com.hedera.hashgraph.sdk.proto.SignedTransaction;
 
 import javax.annotation.Nullable;
-import java.util.*;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.List;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Objects;
 
 abstract class ChunkedTransaction<T extends ChunkedTransaction<T>> extends Transaction<T> implements WithExecuteAll {
     private static final int CHUNK_SIZE = 1024;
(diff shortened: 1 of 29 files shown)


sdk/src/main/java/com/hedera/hashgraph/sdk/ChunkedTransaction.java
```diff
@@ -115,20 +121,21 @@ public int getMaxChunks() {
         return transactionHashes;
     }
 
-    @FunctionalExecutable(type = "java.util.List<TransactionResponse>")
+    @Override @FunctionalExecutable(type = "java.util.List<TransactionResponse>")
```
danielakhterov, line 124: Can we have the annotations on different lines; this would be more clear.
```java
@Override
@FunctionalExecutable(...)
```

sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
```diff
@@ -80,7 +80,7 @@ public SdkRequestT setNodeAccountIds(List<AccountId> nodeAccountIds) {
 
     abstract CompletableFuture<Void> onExecuteAsync(Client client);
 
-    @FunctionalExecutable
+    @Override @FunctionalExecutable
```
danielakhterov, line 83: Can we put annotation on separate lines.

sdk/src/main/java/com/hedera/hashgraph/sdk/ManagedNode.java
```diff
@@ -5,10 +5,12 @@
 
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
+import javax.annotation.Nullable;
 
 abstract class ManagedNode {
     String address;
-    ManagedChannel channel;
+    @Nullable
```
danielakhterov, line 12: Personal preference, but can you add a new line before and after the field with an annotation; I think that's more readable that way
```java
String address;

@Nullable
ManagedChannel channel = null;

final ExecutorService executor;
```

sdk/src/main/java/com/hedera/hashgraph/sdk/MirrorNetwork.java
```diff
@@ -2,8 +2,12 @@
 
 import org.threeten.bp.Duration;
 import org.threeten.bp.Instant;
+import com.google.errorprone.annotations.Var;
 
-import java.util.*;
+//import java.util.*;
```
danielakhterov, line 7: Can you remove this line?

sdk/src/main/java/com/hedera/hashgraph/sdk/Transaction.java
```diff
@@ -3,14 +3,25 @@
 import com.google.errorprone.annotations.Var;
 import com.google.protobuf.ByteString;
 import com.google.protobuf.InvalidProtocolBufferException;
-import com.hedera.hashgraph.sdk.proto.*;
+//import com.hedera.hashgraph.sdk.proto.*;
```
danielakhterov, line 6: We can remove this now.

sdk/src/main/java/com/hedera/hashgraph/sdk/TransactionId.java
```diff
@@ -1,18 +1,15 @@
 package com.hedera.hashgraph.sdk;
 
-import com.google.protobuf.ByteString;
+
```
danielakhterov, line 3: Very much a nit pick, but can we keep it to one new line between `package ...` and the first `import`?

danielakhterov (changes requested):
Just left some minor nit picks on code style.

sdk/src/main/java/com/hedera/hashgraph/sdk/Transaction.java
Sean-Tedrow-LB: I need to start looking at my diffs before I commit

sdk/src/main/java/com/hedera/hashgraph/sdk/ManagedNode.java
```diff
@@ -28,15 +31,8 @@ synchronized ManagedChannel getChannel() {
             return channel;
         }
 
-        var channelBuilder = ManagedChannelBuilder.forTarget(address);
```
danielakhterov, line 31: This is my bad call. I thought we can remove the conditional on if address ends with `:50212` or `:433`, but after the weekly meeting I realized this is meat to stay in the SDK; so can you undo this change?

danielakhterov (changes requested):

danielakhterov (approved):

PR #523, opened 2021-06-16
fix: check if node ids actually map to nodes

Sean-Tedrow-LB:


sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
```diff
@@ -82,15 +87,30 @@ public SdkRequestT setNodeAccountIds(List<AccountId> nodeAccountIds) {
 
     @FunctionalExecutable
     public CompletableFuture<O> executeAsync(Client client) {
-        return onExecuteAsync(client).thenCompose((v) -> executeAsync(client, 1, null));
+        return onExecuteAsync(client).thenCompose((v) -> {
+            if(nodeAccountIds.isEmpty()) {
```
danielakhterov, line 91: Can we extract this node stuff into a separate method; something like `setNodes()`.

sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
```diff
@@ -82,15 +87,30 @@ public SdkRequestT setNodeAccountIds(List<AccountId> nodeAccountIds) {
...
+                throw new IllegalStateException("Request node account IDs were not set before executing");
+            }
```
danielakhterov, line 93: Can you add a new line after this to separate the previous conditional from this one?

sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
```diff
@@ -82,15 +87,30 @@ public SdkRequestT setNodeAccountIds(List<AccountId> nodeAccountIds) {
...
+            if(nodes.isEmpty()) {
+                for(AccountId id : nodeAccountIds) {
```
danielakhterov, line 95: I know we briefly discussed this, but can we use `var accountId` here instead of declaring the type? This would be more consistent with the rest of the SDK.

sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
```diff
@@ -82,15 +87,30 @@ public SdkRequestT setNodeAccountIds(List<AccountId> nodeAccountIds) {
...
+                    @Nullable
+                    Node node = client.network.networkNodes.get(id);
```
danielakhterov, line 97: As above, can we use `var node` here instead of declaring the type for SDK consistency.

sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
```diff
@@ -82,15 +87,30 @@ public SdkRequestT setNodeAccountIds(List<AccountId> nodeAccountIds) {
...
+        });
     }
 
     private CompletableFuture<O> executeAsync(Client client, int attempt, @Nullable Throwable lastException) {
         if (attempt > maxAttempts) {
             return CompletableFuture.<O>failedFuture(new Exception("Failed to get gRPC response within maximum retry count", lastException));
         }
 
-        var node = client.network.networkNodes.get(getNodeAccountId());
+        var node = nodes.get(nextNodeIndex);
```
danielakhterov, line 113: Not sure if this suggestion is a good one, but since I recommended extracting some stuff from `executeAsync` into `setNodes()` maybe it'd make sense to implement `getNode()` which simply returns `nodes.get(nextNodeIndex)` and is called here `var node = getNode()`?

danielakhterov (changes requested):
Just some minor nit picks; the logic looks sound to me.

sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
Sean-Tedrow-LB: We can, but it will only ever be used in one place.  I'm generally in the habit of not factoring something out until it's used in multiple places.

danielakhterov (approved):

PR #524, opened 2021-06-16
Feat: add NFTs and HIP 18

Sean-Tedrow-LB:
Closes: #528 
Closes: #534 
Adds misc bug-fixes.


sdk/src/main/java/com/hedera/hashgraph/sdk/TokenCreateTransaction.java
```diff
@@ -13,14 +13,19 @@
...
+    AccountId treasuryAccountId = null;
+    @Nullable
+    AccountId autoRenewAccountId = null;
 
     public TokenCreateTransaction() {
         builder = TokenCreateTransactionBody.newBuilder();
 
         setAutoRenewPeriod(DEFAULT_AUTO_RENEW_PERIOD);
         setMaxTransactionFee(new Hbar(30));
+        setInitialSupply(0);
```
danielakhterov, line 28: We should not be setting the default initial supply. Let Hedera error here.

sdk/src/main/java/com/hedera/hashgraph/sdk/TokenMintTransaction.java
```diff
@@ -51,6 +54,28 @@ public long getAmount() {
         return builder.getAmount();
     }
 
+    public TokenMintTransaction addMetadata(byte[] metadata) {
+        requireNotFrozen();
+        builder.addMetadata(ByteString.copyFrom(metadata));
+        return this;
+    }
+
+    public TokenMintTransaction addMetadatas(List<byte[]> metadatas) {
```
danielakhterov, line 63: I think you mean to use `setMetadata()` here instead of `addMetadatas()`. _imo_ This should completely override the list as well.

sdk/src/main/java/com/hedera/hashgraph/sdk/TransferTransaction.java
```diff
@@ -91,19 +59,26 @@ public TransferTransaction() {
...
-
-        var current = tokenTransfer.containsKey(accountId) ?
-            Objects.requireNonNull(tokenTransfer.get(accountId)) :
-            0L;
+    public Map<TokenId, List<NftTransfer>> getTokenNftTransfers() {
+        return nftTransfers;
+    }
 
-        tokenTransfer.put(accountId, current + value);
+    public TransferTransaction addTokenNftTransfer(NftId nftId, AccountId sender, AccountId receiver) {
```
danielakhterov, line 70: Perhaps `addNftTransfer()`? I know we added `Token*` for all the classes, but I don't fee it's necessary here.

danielakhterov (changes requested):

sdk/src/main/java/com/hedera/hashgraph/sdk/TokenCreateTransaction.java
Sean-Tedrow-LB: I have a vague memory of having a good reason for doing this.  I'll talk to you about it tomorrow.

Sean-Tedrow-LB: From `TokenCreate.proto`:

`    uint64 initialSupply = 4; // Specifies the initial supply of tokens to be put in circulation. The initial supply is sent to the Treasury Account. The supply is in the lowest denomination possible. *In the case for NON_FUNGIBLE_UNIQUE Type the value must be 0*`

So this was my dirty solution.

sdk/src/main/java/com/hedera/hashgraph/sdk/TransferTransaction.java
```diff
@@ -187,4 +178,66 @@ boolean onFreeze(TransactionBody.Builder bodyBuilder) {
...
+    private Map<AccountId, Long> getTokenTransferMap(TokenId tokenId) {
+        // Cannot use `Map.merge()` as it uses `BiFunction`
+        var map = tokenTransfers.containsKey(tokenId) ?
+            Objects.requireNonNull(tokenTransfers.get(tokenId)) :
+            new HashMap<AccountId, Long>();
+        tokenTransfers.put(tokenId, map);
+        return map;
+    }
+
+    private List<NftTransfer> getNftTransferList(TokenId tokenId) {
```
danielakhterov, line 191: Can we update this to `Map<TokenId, NftTransfer>` where `NftTransfer` does not contain `tokenId`. Right now, the Go SDK returns `map[TokenID]NftTransfer` which _I thinik_ makes a little more sense. Especially considering the protobufs `TokenTransferList` contains `token` and `nftTransfers` for that specific `TokenId`.

sdk/src/main/java/com/hedera/hashgraph/sdk/TokenUpdateTransaction.java
```diff
@@ -158,6 +159,16 @@ public TokenUpdateTransaction setSupplyKey(Key key) {
         return this;
     }
 
+    public Key getCustomFeeKey() {
```
danielakhterov, line 162: Can we use `feeScheduleKey` instead of  `customFeeKey`. I feel like the former reads just a bit better. Up for debate though as long as we're consistent with the other SDKs.

sdk/src/main/java/com/hedera/hashgraph/sdk/TokenCreateTransaction.java
danielakhterov: I guess what we should do then is set initial balance to 0 when someone `setTokenType(TokenType.NON_FUNGIBLE_UNIQUE)`, but not by default. Although, I'm pretty sure the default value for this field is 0 anyways so this is technically unneeded.

sdk/src/main/java/com/hedera/hashgraph/sdk/CustomFeeUpdateTransaction.java
```diff
@@ -0,0 +1,125 @@
...
+import org.threeten.bp.Duration;
+import org.threeten.bp.Instant;
+
+import java.util.LinkedHashMap;
+import javax.annotation.Nullable;
+import java.util.Objects;
+import java.util.List;
+import java.util.ArrayList;
+
+public class CustomFeeUpdateTransaction extends Transaction<CustomFeeUpdateTransaction> {
```
danielakhterov, line 17: Lets name this `TokenFeeScheduleUpdateTransaction` for consistency. If we decide to change this later, we can deprecate this class and make a new one.

danielakhterov (changes requested):

sdk/src/main/java/com/hedera/hashgraph/sdk/CustomFeeUpdateTransaction.java
Sean-Tedrow-LB: Ryan was the one who initially suggested I follow the `CustomKey` naming convention.  We need to settle on our naming convention.  I think we should have that conversation tomorrow with Ryan.  Last I heard from Ryan, he was having second thoughts.  I favor the `CustomKey` naming convention, personally, and in that case I think we should use it consistently.  But we'll see how that conversation goes!

sdk/src/main/java/com/hedera/hashgraph/sdk/TransferTransaction.java
Sean-Tedrow-LB: Not sure what your intention is, there.  It's possible (and expected) for multiple NFTs with the same `TokenId` to be transferred in one `TransferTransaction`, and it would be an error to prevent that.  Unless you meant `Map<NftId, NftTransfer>` instead of `Map<TokenId, NftTransfer>`?

danielakhterov: Closing as this was my misunderstanding.

danielakhterov (approved):

PR #527, opened 2021-06-17
feat: Adding Id checksums

andrix10:
Closes: #525

### DIFF
--- a/sdk/src/main/java/com/hedera/hashgraph/sdk/AccountBalanceQuery.java
+++ b/sdk/src/main/java/com/hedera/hashgraph/sdk/AccountBalanceQuery.java
@@ -63,9 +63,14 @@ public final class AccountBalanceQuery extends Query<AccountBalance, AccountBala
     }
 
     @Override
-    void validateNetworkOnIds(@Nullable NetworkName networkName) {
-        EntityIdHelper.validateNetworkOnIds(this.accountId, networkName);
-        EntityIdHelper.validateNetworkOnIds(this.contractId, networkName);
+    void validateNetworkOnIds(Client client) {
+        if (accountId != null) {
+            accountId.validate(client);
+        }
+
+        if (contractId != null) {
+            contractId.validate(client);
+        }
     }
 
     @Override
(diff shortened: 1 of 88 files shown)


sdk/src/main/java/com/hedera/hashgraph/sdk/AccountBalanceQuery.java
```diff
@@ -20,12 +19,19 @@
...
+    ContractId contractId;
+
     public AccountBalanceQuery() {
         builder = CryptoGetAccountBalanceQuery.newBuilder();
     }
 
     public AccountId getAccountId() {
-        return AccountId.fromProtobuf(builder.getAccountID());
+        if (accountId == null) {
+            return new AccountId(0);
```
mehcode, line 31: We should not be returning "ID 0" if the account ID or contract ID has not been set.

sdk/src/main/java/com/hedera/hashgraph/sdk/AccountId.java
```diff
@@ -37,6 +44,33 @@ public AccountId(@Nonnegative long shard, @Nonnegative long realm, @Nonnegative
...
+                this.checksum = EntityIdHelper.checksum(network.toString(), shard + "." + realm + "." + num);
+            } else {
+                this.checksum = checksum;
+            }
+        } else {
+            this.checksum = null;
+        }
+    }
+
+    public static AccountId withNetwork(@Nonnegative long num, NetworkName network) {
```
mehcode, line 68: I think we need to discuss this more. `AccountId.withNetwork(15, NetworkName.MAINNET)` doesn't feel right to me.

---

The goal here is to enable printing an account ID with a checksum - which requires knowing the network. Users will have an `AccountId` type already (most likely) and will want to print the ID with a checksum against the current network. The current network in this case most likely means a network from a `Client` instance. 

 - `client.formatEntityId(accountId)`
 - `accountId.withChecksum(client).toString()`

cc @lbaird @steven-sheehy @Sean-Tedrow-LB @danielakhterov @andrix10 

Pinging some from Hedera who might be interested in how this API gets figured out. As printing a valid ID with a  checksum is almost as important as validating it with the checksum.

danielakhterov: What we can do is set the network on the entity ID implicitly when executing with a client that has a network set. So if someone constructs a client using `Client.for[Mainnet|Testnet|Previewnet]()`, but uses entity IDs without a network, we can set the network accordingly when they execute any request.

sdk/src/main/java/com/hedera/hashgraph/sdk/NetworkName.java
```diff
@@ -0,0 +1,39 @@
+package com.hedera.hashgraph.sdk;
+
+public enum NetworkName {
+    MAINNET(0),
+    TESTNET(1),
+    PREVIEWNET(2);
```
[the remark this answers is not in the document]
danielakhterov: Since users do not have access to the underlying data of each enumeration variant I'm going to just go with just an integer. I think we'll be able to update to be a byte array later without an issue.

danielakhterov:
Merging as this PR is needed for today's release and test are passing. DCO isn't passing since this PR originally started in a different repo where DCO wasn't required.

PR #543, opened 2021-07-06
fix: pull requests should run on forks using this repos secrets

danielakhterov:
**Description**:
Attempt to fix CI when run from a fork. Current it fails since the secrets cannot be accessed from a forks pull request.


danielakhterov:
Merging since this change _should_ simply allow forked repo's PRs to run integration tests successfully.

Issue #478, opened 2021-05-04
SDK should use a set of validated nodes for session run


Nana-EC:
**Problem**
The SDK uses a set of nodes (Static or custom provided by user) to which it randomly submits transactions.
Occasionally some of these nodes are unavailable from startup.

The SDK does have logic to prioritize nodes, however, if a node is unavailable from startup it should just be ignored instead of periodically trying it and getting UNAVAILABLE.

**Solution**
- [Ping](https://github.com/hashgraph/hedera-sdk-java/blob/master/sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java#L302) all the nodes on start and determine a set of validated nodes that were successfully reached.
- Set the validated nodes as the Clients networkMap and use these nodes for the session of the SDK.

**Alternatives**

**Additional Context**
Users of the SDK may be using the SDK to run tests to verify scenarios against the network.
In those cases UNAVAILABLE nodes make troubleshooting more involving, especially tests run in CI/CD etc

donaldthibeau:
I believe this feature should also leverage the address book for pulling the latest node address book.

mehcode:
> The SDK does have logic to prioritize nodes, however, if a node is unavailable from startup it should just be ignored instead of periodically trying it and getting UNAVAILABLE.

I want to make it clear that this is (almost) exactly what happens. You should never, ever be getting `UNAVAILABLE`. You'll see scary looking logs that print warnings that you got it. However, the SDK then will retry on another node. I believe the action items here are:

 - [ ] Stop printing the stack trace when we retry ( ref https://github.com/hashgraph/hedera-sdk-java/blob/master/sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java#L127 )

 - [ ] Add `Client.setMaxAttempts` to allow defaulting `setMaxAttempts` for an executable at the `Client` 

 - [ ] Add `Client.setNodeMaxAttempts` (defaults to inf.) to configure the maximum number of times we will try talking to a node after an UNAVAILABLE until we simply stop talking to that node 

 - [ ] Add `Client.setNodeWaitTime` to configure the minimum wait time we will wait after talking to a failing node

 - [ ] Add `Client.pingAsync` (and `WithPingAsync`) and reset wait times and attempt counts on the node in question, remove `Client.ping` 

 - [ ] Add `Client.pingAllAsync` (and `WithPingAllAsync`) - that iterates through the network and pings all nodes (at once, be async)

[...]

PR #564, opened 2021-07-20
feat: implement `Client.ping[All]Async()`

danielakhterov:
**Description**:

**Related issue(s)**:

Closes: #478

**Notes for reviewer**:
`Client.getRequestTimeout()` is required to be exposed for `FunctionalExecutable` to work since it doesn't require `client` to be passed in as a parameter.

**Checklist**

- [X] Implement `Client.pingAll()`
- [X] Implement `Client.ping(AccountId)`
- [X] Implement `Client.[get|set]MaxNodeAttempts()`
- [X] Implement `Client.[get|set]NodeWaitTime()`
- [X] Implement `Client.[get|set]MaxAttempts()`
- [X] Implement `Client.getRequestTimeout()` - This is needed for `FunctionalExecutable` to work with `ping` and `pingAll`

### DIFF
--- a/sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
+++ b/sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
@@ -20,7 +20,7 @@ import static com.hedera.hashgraph.sdk.FutureConverter.toCompletableFuture;
 abstract class Executable<SdkRequestT, ProtoRequestT, ResponseT, O> implements WithExecute<O> {
     private static final Logger logger = LoggerFactory.getLogger(Executable.class);
 
-    protected int maxAttempts = 10;
+    protected int maxAttempts = Client.DEFAULT_MAX_ATTEMPTS;
     protected int nextNodeIndex = 0;
     protected List<AccountId> nodeAccountIds = Collections.emptyList();
     protected List<Node> nodes = new ArrayList<>();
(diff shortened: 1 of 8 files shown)


sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
```diff
@@ -86,6 +86,10 @@ public SdkRequestT setNodeAccountIds(List<AccountId> nodeAccountIds) {
     @Override
     @FunctionalExecutable
     public CompletableFuture<O> executeAsync(Client client) {
+        if (maxAttempts == Client.DEFAULT_MAX_ATTEMPTS) {
+            maxAttempts = client.getMaxAttempts();
+        }
```
Sean-Tedrow-LB, line 91: This way of handling maxAttempts seems odd to me.  If I understand correctly, the idea is that the user can set `maxAttempts` on the `Executable`, and then that new, non-default value will stick, instead of being set to `client.getMaxAttempts()` on execute.

However, if the user sets the `maxAttempts` on the `Executable` to 10 (the value of `Client.DEFAULT_MAX_ATTEMPTS`), their custom `maxAttempts` of 10 will be overwritten by `client.getMaxAttempts()`, which might not be 10 if the user has changed it.

Sean-Tedrow-LB (approved):

PR #565, opened 2021-07-20
chore: add some string-conversion stuff

### LINKED ISSUE #561: `Hbar.[from|to]String()` should be reversible
### Problem

`Hbar.toString()` produces a string, but you cannot use that string within `Hbar.fromString()`.

### Solution

Update `Hbar.fromString()` to correctly decode the `HbarUnit` type from the string and use that for the `Hbar` construction.

### Alternatives

_No response_

Sean-Tedrow-LB:

**Description**:

- Add `toString()` to `AccountBalance`
- Make `Hbar.fromString()` understand unit symbol prefixes.
- Add `toStringRaw()` and `toStringDER()` to `PrivateKey` and `PublicKey`
- Add tests for these additions.

**Related issue(s)**:

Fixes #561

### DIFF
--- a/sdk/src/main/java/com/hedera/hashgraph/sdk/Hbar.java
+++ b/sdk/src/main/java/com/hedera/hashgraph/sdk/Hbar.java
@@ -41,7 +41,7 @@ public final class Hbar implements Comparable<Hbar> {
 
     Hbar(BigDecimal amount, HbarUnit unit) {
         var tinybars = amount.multiply(BigDecimal.valueOf(unit.tinybar));
-
+        
         if (tinybars.doubleValue() % 1 != 0) {
             throw new IllegalArgumentException("Amount and Unit combination results in a fractional value for tinybar.  Ensure tinybar value is a whole number.");
         }
@@ -64,6 +64,28 @@ public final class Hbar implements Comparable<Hbar> {
      */
     public static final Hbar MIN = Hbar.from(-50_000_000_000L);
 
+    private static int findNumber(CharSequence text) {
+        for(int i = 0; i < text.length(); i++) {
+            var c = text.charAt(i);
+            if(Character.isDigit(c) || c == '+' || c == '-') {
+                return i;
+            }
+        }
+        throw new IllegalArgumentException("Attempted to convert string \"" + text + "\" to Hbar, but no number was found");
+    }
+
+    private static HbarUnit getUnit(String symbolString) {
+        if(symbolString.length() == 0) {
+            return HbarUnit.HBAR;
+        }
+        for(var unit : HbarUnit.values()) {
+            if(unit.getSymbol().equals(symbolString)) {
+                return unit;
+            }
+        }
+        throw new IllegalArgumentException("Attempted to convert string to Hbar, but symbol \"" + symbolString + "\" was not recognized");
+    }
+
     /**
      * Converts the provided string into an amount of hbars.
      *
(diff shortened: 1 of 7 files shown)


sdk/src/main/java/com/hedera/hashgraph/sdk/Hbar.java
```diff
@@ -64,14 +64,39 @@ public Hbar(BigDecimal amount) {
      */
     public static final Hbar MIN = Hbar.from(-50_000_000_000L);
 
+    private static int findNumber(CharSequence text) {
+        for(int i = 0; i < text.length(); i++) {
```
danielakhterov, line 68: Should just use `String.split()`

sdk/src/main/java/com/hedera/hashgraph/sdk/Hbar.java
```diff
@@ -64,14 +64,39 @@ public Hbar(BigDecimal amount) {
...
+
     /**
      * Converts the provided string into an amount of hbars.
      *
      * @param text The string representing the amount of Hbar
      * @return {@link com.hedera.hashgraph.sdk.Hbar}
      */
     public static Hbar fromString(CharSequence text) {
-        return new Hbar(new BigDecimal(text.toString()), HbarUnit.HBAR);
+        int number_i = findNumber(text);
```
danielakhterov, line 96: Variable names should be camelCase

sdk/src/test/java/com/hedera/hashgraph/sdk/HbarTest.java
```diff
@@ -83,7 +83,9 @@ void constructorWorks() {
 
     @Test
     void fromString() {
-        assertEquals(Hbar.fromString("1").toTinybars(), 100000000);
+        assertEquals(Hbar.fromString("1").toTinybars(), 100_000_000);
+        assertEquals(Hbar.fromString("ℏ 1").toTinybars(), 100_000_000);
```
danielakhterov, line 87: This should be backwards, `1 ℏ` not `ℏ 1`

sdk/src/test/java/com/hedera/hashgraph/sdk/HbarTest.java
```diff
@@ -83,7 +83,9 @@ void constructorWorks() {
...
+        assertEquals(Hbar.fromString("mℏ1").toTinybars(), 100_000);
```
danielakhterov, line 88: Lets require a space between the symbol and number as well. So `1ℏ` would not work, while `1 ℏ` will. A strict format is better than a lose one.

danielakhterov (changes requested):

sdk/src/main/java/com/hedera/hashgraph/sdk/Hbar.java
danielakhterov: Use https://regex101.com/r/6Q5EOm/1

Sean-Tedrow-LB: How about `.split(" ")`?

Let the `BigDecimal` constructor and `getUnit()` do the work of deciding if they're correctly formatted.

danielakhterov: Reason for regex is to validate the input makes sense before we try to do anything with it. What if someone submits ` 1 ℏ` This would split on the first space and cause `BigDecimal` to return `0` (I think) and the `getUnit()` method would error even though the unit wasn't the issue, it was the misformatted input. You _can_ use the regex to validate and then use the `.split(" ")` instead of using the regex groups if you want, but to re-iterate it's mainly for validating input.

danielakhterov (approved):

PR #602, opened 2021-08-11
Refactor: Remove builders

### LINKED ISSUE #566: Do not use protobufs to store values for requests
### Problem

Currently all requests contain some sort of `builder` internally. For some setters/getters we use local variables, while others we encode or decode from the `builder`. For instance `AccountCreateTransaction.[get|set]AutoRenewPeriod()` will write and read the value from the `builder`. Instead we should store the fields locally in the class and only encode the protobufs when we need to construct the request to be sent out.

### Solution

Create private fields for each getter/setter instead of writing to the `builder` and remove `builder` as a field in the class for al requests.

### Alternatives

_No response_

Sean-Tedrow-LB:
**Description**:

Removes builders from transactions and queries.  We no longer use builders to store `Transaction` or `Query` state.
Also fixes misc bugs that I found in the process of performing the refactor, and renames and reorganizes things in `Transaction`, `Query`, and `ChunkedTransaction`.

**Related issue(s)**:

Fixes #566

**Checklist**


sdk/src/main/java/com/hedera/hashgraph/sdk/Transaction.java
```diff
@@ -729,6 +735,13 @@ protected void requireOneNodeAccountId() {
         }
     }
 
+    protected TransactionBody.Builder spawnBodyBuilder() {
```
danielakhterov, line 738: What about `toProtobufTransactionBodyBuilder()` This would be somewhat consistent with the rest of the SDK since we use `toProtobuf()` quite a bit. We just don't use the word `spawn` anywhere in the SDK.

sdk/src/main/java/com/hedera/hashgraph/sdk/LiveHashDeleteTransaction.java
```diff
@@ -18,23 +18,16 @@
  * by either the key of the owning account, or at least one of the keys associated to the livehash.
  */
 public final class LiveHashDeleteTransaction extends Transaction<LiveHashDeleteTransaction> {
-    private final CryptoDeleteLiveHashTransactionBody.Builder builder;
-
     @Nullable
-    AccountId accountId = null;
+    private AccountId accountId = null;
+    private byte[] hash = {};
```
danielakhterov, line 23: We should really be using `ByteString hash` not `byte[] hash` since `ByteString` does not copy.

I'm only going to make this comment here, but know that this should apply to all requests that store `byte[]` fields.

sdk/src/main/java/com/hedera/hashgraph/sdk/Query.java
```diff
@@ -31,16 +31,22 @@
...
+    private Hbar queryPayment = null;
+
+    @Nullable
+    private Hbar maxQueryPayment = null;
+
+    @Nullable
+    private Hbar chosenQueryPayment = null;
 
     Query() {
         builder = com.hedera.hashgraph.sdk.proto.Query.newBuilder();
```
danielakhterov, line 52: Still using a builder here?

sdk/src/main/java/com/hedera/hashgraph/sdk/TokenNftInfoQuery.java
```diff
@@ -152,7 +129,7 @@ void validateChecksums(Client client) throws BadEntityIdException {
 
     @Override
     CompletableFuture<Void> onExecuteAsync(Client client) {
-        int modesEnabled = (isByNft() ? 1 : 0) + (isByToken() ? 1 : 0) + (isByAccount() ? 1 : 0);
+        int modesEnabled = (nftId != null ? 1 : 0) + (tokenId != null ? 1 : 0) + (accountId != null ? 1 : 0);
         if(modesEnabled > 1) {
             throw new IllegalStateException("TokenNftInfoQuery must be one of byNftId, byTokenId, or byAccountId, but multiple of these modes have been selected");
```
danielakhterov, line 134: Technically none of them can also be set.

sdk/src/main/java/com/hedera/hashgraph/sdk/FreezeTransaction.java
```diff
@@ -86,14 +86,30 @@ void validateChecksums(Client client) throws BadEntityIdException {
...
+    void initFromTransactionBody() {
+        var body = sourceTransactionBody.getFreeze();
+        startHour = body.getStartHour();
+        startMinute = body.getStartMin();
+        endHour = body.getEndHour();
+        endMinute = body.getEndMin();
+    }
+
+    FreezeTransactionBody.Builder build() {
+        var builder = FreezeTransactionBody.newBuilder();
```
danielakhterov, line 98: This could be `return FreezeTransactionBody.newBuilder().setStartHour(startHour)...;`

danielakhterov (approved):

PR #605, opened 2021-08-11
fix: fix bug in topicMessageQuery

### LINKED ISSUE #600: Topic message chunk may be lost when a topic message subscription is retried
### Description

Every time a topic subscription is tried, a new `pendingMessages` `HashMap` is created to hold topic message chunks for reassembly.

In case an retriable error happens, a new subscription is tried with an empty `pendingMessages`. As a result,  if before the retry, there are pending chunks, those chunks are lost thus the corresponding topic message is lost.

### Steps to reproduce

1. start a topic message query with chunked topic messages
2. trigger a retry when there are chunks pending reassembly
3. the topic message whose chunks are partially received before the retry will never be able to reassemble

### Additional context

_No response_

### Hedera network

testnet

### Version

v2.0.10

### Operating system

_No response_

Sean-Tedrow-LB:

**Related issue(s)**:

Fixes #600 

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Unreleased
 
+- Fixed bug in `TopicMessageQuery`
 - Fixed bugs in `ChunkedTransaction.getAllTransactionHashesPerNode()`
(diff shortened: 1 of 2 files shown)


danielakhterov (approved):

steven-sheehy:
No test for this scenario was added to the unit test.

Sean-Tedrow-LB:
> No test for this scenario was added to the unit test.

Yes, that's true.  That's something I'm going to work on tomorrow.

Sean-Tedrow-LB:
Added regression test #610

PR #631, opened 2021-08-17
feat: support TLS on previewnet

### LINKED ISSUE #601: Complete TLS support
### Problem

The SDK currently does not have complete TLS support. 

### Solution

Hedera Services will be adding TLS support in the 0.17.0 release to enable complete TLS support.

### Alternatives

_No response_

danielakhterov:

**Description**:

Add support for TLS. 

**Related issue(s)**:

Fixes #601 

**Notes for reviewer**:

**Checklist**


steven-sheehy:
Previewnet will have changed ciphers and certs soon. So please hold off merging until tested against that.

danielakhterov:
@steven-sheehy TLS support was supposed to be part of tonight's SDK release. By soon do you mean by the end of the day, or by the end of the week?

steven-sheehy:
Probably tomorrow. No one is around to confirm status of previewnet. You could always do another release with just TLS and go forward with your current release

sdk/src/main/java/com/hedera/hashgraph/sdk/HederaTrustManager.java
```diff
@@ -0,0 +1,61 @@
...
+            return;
+        }
+
+        var addressBook = Network.NODE_ADDRESS_BOOK.get(networkName);
+        if (addressBook == null) {
+            return;
+        }
+
+        for (var cert : chain) {
+            for (var nodeAddress : addressBook.nodeAddresses) {
```
Sean-Tedrow-LB, line 39: Make nodeAddresses a map

sdk/src/main/java/com/hedera/hashgraph/sdk/HederaTrustManager.java
```diff
@@ -0,0 +1,61 @@
...
+import javax.net.ssl.X509TrustManager;
+import java.nio.charset.StandardCharsets;
+import java.security.cert.CertificateException;
+import java.security.cert.X509Certificate;
+
+class HederaTrustManager implements X509TrustManager {
+    @Nullable
+    private final NetworkName networkName;
+
+    private final AccountId nodeAccountId;
```
Sean-Tedrow-LB, line 16: `HederaTrustManager` should have this node's hash

sdk/src/main/java/com/hedera/hashgraph/sdk/HederaTrustManager.java
```diff
@@ -0,0 +1,61 @@
...
+            for (var nodeAddress : addressBook.nodeAddresses) {
+                if (nodeAddress.accountId.equals(nodeAccountId) && !nodeAddress.certHash.isEmpty()) {
+                    var digest = new SHA384Digest();
+                    var certHash = new byte[digest.getDigestSize()];
+                    digest.update(cert.getEncoded(), 0, cert.getEncoded().length);
+                    digest.doFinal(certHash, 0);
+
+                    var nodeAddressBookCertHash = new String(nodeAddress.certHash.toByteArray(), StandardCharsets.UTF_8);
+
+                    if (!nodeAddressBookCertHash.equals(Hex.toHexString(certHash))) {
```
Sean-Tedrow-LB, line 48: This seems to throw an exception if _any_ certificate in the chain does _not_ match this node's hash.  From what I understand, this is not the behavior you're aiming for.

danielakhterov:
I'm guessing mirror node is down which is why the CI failed. I know this isn't CI, but I ran integration tests on my machine and they all passed.

sdk/src/integrationTest/java/IntegrationTestEnv.java
```diff
@@ -173,6 +173,7 @@ public void nextNode(Map<String, AccountId> outMap) throws Exception {
                     outMap.put(node.getKey(), node.getValue());
                     return;
                 } catch (Exception ignored) {
+                    throw ignored;
```
steven-sheehy, line 176: This is an oxymoron. Shouldn't even catch it if you're just going to re-throw it as is.

sdk/src/main/java/com/hedera/hashgraph/sdk/IPAddress.java
```diff
@@ -0,0 +1,49 @@
...
+
+class IPv4Address {
+    IPv4AddressPart network;
+    IPv4AddressPart host;
+
+    IPv4Address() {
+    }
+
+    static IPv4Address fromProtobuf(ByteString address) {
+        return new IPv4Address()
```
steven-sheehy, line 13: This class and `IPv4AddressPart` are not necessary. Java already has a class that can parse IPv4 addresses:

`InetAddress.getByAddress(address.toByteArray()).getHostAddress()`

sdk/src/main/java/com/hedera/hashgraph/sdk/Network.java
```diff
@@ -36,13 +41,114 @@
...
+        } catch (InterruptedException | TimeoutException e) {
             // This should never occur. The network is empty.
         }
     }
 
+    static Network forNetwork(ExecutorService executor, Map<String, AccountId> network) {
+        return new Network(executor, network);
+    }
+
+    static Network forMainnet(ExecutorService executor) {
```
steven-sheehy, line 53: Add `@SuppressWarnings("java:S1313")` to this and other methods with IPs to suppress the numerous sonar warnings.

sdk/src/main/java/com/hedera/hashgraph/sdk/Network.java
```diff
@@ -36,13 +41,114 @@
...
+                    break;
+            }
+
+            nodeAddressBooks.put(networkName, readAddressBookResource(resource));
+        }
+
+        return this;
+    }
+
+    static Map<AccountId, NodeAddress> readAddressBookResource(String fileName) {
```
steven-sheehy, line 132: It would be nice to be able to provide external address books for non-public environments.

sdk/src/main/java/com/hedera/hashgraph/sdk/NodeAddressBook.java
```diff
@@ -0,0 +1,58 @@
...
+
+import com.google.common.base.MoreObjects;
+import com.google.protobuf.ByteString;
+import com.google.protobuf.InvalidProtocolBufferException;
+
+import java.util.ArrayList;
+import java.util.List;
+
+class NodeAddressBook {
+    List<NodeAddress> nodeAddresses;
```
steven-sheehy, line 11: Would be better as private final so this class is immutable and thread safe.

sdk/src/main/java/com/hedera/hashgraph/sdk/NodeAddress.java
```diff
@@ -0,0 +1,148 @@
...
+import java.util.List;
+
+class NodeAddress {
+    String publicKey;
+    AccountId accountId;
+    long nodeId;
+    ByteString certHash;
+    List<Endpoint> addresses;
+    String description;
+    long stake;
```
steven-sheehy, line 18: Would be better as private and final so this class is immutable and thread-safe

sdk/src/main/resources/addressbook/mainnet.pb
```diff
@@ -0,0 +1,24 @@
+
+�"�308201a2300d06092a864886f70d01010105000382018f003082018a0282018100c4561e3c278cd650e80c413ca44423c1c3c13cf1475f6f6976d597ae432b49ab42086b79b841326054b8b3dcf57d8fcd79bfc058183ca24cd4c1cbc574ed1117e2f5b7b3c63ce7b06d9b4efcf7375637b41fe6f53c811b9de6143f3a52957cdf956775120b33703ff57621407ab9575bc2d35c0d44f0983fc1ef63a4ff5209f070c92af106295601c96bced064ec190197019c6811c4c8dd80cb4f4ac71f9ad76e7ac89456fbf4f011f90abd2d90536e8234651f6bef927e3d5d8b7bf459050983beca3abef2a9d97af345772a7740e9699275b018ea0df286add6ce923ef908fbe762a75f21116862db44d3dca1d44b4d2e8dc1066c5006bb5a7d954ad255d4b603273475e511aeb485d069a067c0ab5c24538c933c06b5a6aefa94005c2915213e4ccdae6c942f6272f9dd5282d6b890f1f20efd2399cd674924fa57046ac6da32e73951a73113e91fc2b7ff29e4851b83ff39f83ba9ec6f08cefdbb6cbbbffabfdfaa91d930f7200da48137c394cbd13e701ecdc2616fd21bad681aa4f0010203010001(2:`6e9a8abcdcde6e1149a3ebe17fd58d8905839a86db72b06a3ea20aa176f868b2548852ae423d7a9c6b7cff9e714629a2
```
steven-sheehy, line 2: Did you verify all these address books contain IP information? Just looking at the binary on GitHub I don't see the IP addresses like I do for the previewnet one. But entirely possible it's just not showing up.

sdk/src/main/java/com/hedera/hashgraph/sdk/Node.java
```diff
@@ -19,6 +27,23 @@
...
+
+    public Node setNetworkName(NetworkName networkName) {
+        this.networkName = networkName;
+        return this;
+    }
+
+    @Override
+    ChannelCredentials getChannelCredentials() {
+        return TlsChannelCredentials.newBuilder()
+            .trustManager(new HederaTrustManager(networkName, accountId))
```
steven-sheehy, line 43: We should have an integration test that verifies TLS works.

sdk/src/main/java/com/hedera/hashgraph/sdk/HederaTrustManager.java
```diff
@@ -0,0 +1,63 @@
...
+    @Nullable
+    private final String certHash;
+
+    HederaTrustManager(@Nullable NetworkName networkName, AccountId nodeAccountId) {
+        if (networkName == null) {
+            this.certHash = null;
+            return;
+        }
+
+        var nodeAddressBook = Network.nodeAddressBooks.get(networkName);
```
steven-sheehy, line 22: Passing around `NetworkName` then using that to look into a static hashmap of `Network` is not a very good abstraction and highly couples these all together. Recommend a `Network` instance encapsulate everything and a `network.getAddressBook()` be added. That will also better support non-public networks.

sdk/src/main/java/com/hedera/hashgraph/sdk/HederaTrustManager.java
```diff
@@ -0,0 +1,63 @@
...
+    @Override
+    public void checkServerTrusted(X509Certificate[] chain, String authType) throws CertificateException {
+        if (certHash == null) {
+            return;
+        }
+
+        for (var cert : chain) {
+            var digest = new SHA384Digest();
+            var certHash = new byte[digest.getDigestSize()];
+            digest.update(cert.getEncoded(), 0, cert.getEncoded().length);
```
steven-sheehy, line 48: Call `cert.getEncoded()` once and reuse as it looks like it clones the byte array internally.

sdk/src/main/java/com/hedera/hashgraph/sdk/HederaTrustManager.java
```diff
@@ -0,0 +1,63 @@
...
+            digest.doFinal(certHash, 0);
+
+            if (this.certHash.equals(Hex.toHexString(certHash))) {
+                return;
+            }
+        }
+
+        throw new CertificateException();
```
steven-sheehy, line 56: Should probably add some helpful message

sdk/src/main/java/com/hedera/hashgraph/sdk/Network.java
steven-sheehy, line 53: Exposing a way to switch these built-in endpoints to the TLS port without requiring clients to configure every IP would be nice.

sdk/src/main/java/com/hedera/hashgraph/sdk/Network.java
```diff
@@ -36,13 +41,114 @@
...
+        try {
+            var contents = ByteStreams.toByteArray(Objects.requireNonNull(Client.class.getClassLoader().getResourceAsStream(fileName)));
```
steven-sheehy, line 134: I get a NullPointerException running this

steven-sheehy: This is due to the use of `./` in the address book file paths. Removing those and it succeeds.

steven-sheehy, line 134: This inputstream is not properly closed in a try with resources. It can also fail due to multiple classloaders. Suggest using Guava:

```java
        try (InputStream inputStream = Resources.getResource(fileName).openStream()) {
            var contents = ByteStreams.toByteArray(inputStream);
```

sdk/src/integrationTest/resources
```diff
@@ -0,0 +1 @@
+sdk/src/main/resources/
```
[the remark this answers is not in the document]
danielakhterov: I'll need to double check if this is still an issue, but I added this link because integration tests would fail otherwise.

sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
```diff
@@ -78,11 +99,15 @@
...
-     * @param network the map of node IDs to node addresses that make up the network.
+     * @param networkMap the map of node IDs to node addresses that make up the network.
      * @return {@link com.hedera.hashgraph.sdk.Client}
      */
-    public static Client forNetwork(Map<String, AccountId> network) {
-        return new Client(network);
+    public static Client forNetwork(Map<String, AccountId> networkMap) {
+        var executor = createExecutor();
+        var network = Network.forNetwork(executor, networkMap);
+        var mirrorNetwork = MirrorNetwork.forMainnet(executor);
```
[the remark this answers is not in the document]
danielakhterov: This was definitely a mistake from copy and pasting. Will fix.

sdk/src/main/java/com/hedera/hashgraph/sdk/Node.java
danielakhterov: When this PR was merged the was no way to "enable" TLS for a previous network which would make creating a TLS integration test difficult; you'd need to set the network yourself using `Client.forNetwork()` where the IPs were `:50212`. Since the IPs change per previewnet/testnet hard coding them into an integration test seems like a bad idea. Nonetheless, once we add support for switching to TLS through some configuration we should definitely add a test for it.

sdk/src/main/java/com/hedera/hashgraph/sdk/Network.java
danielakhterov: Yea, you're definitely right. For this release we wanted to get as basic of TLS support as humanly possible which is why everything looks bad is is hard coded.

sdk/src/main/java/com/hedera/hashgraph/sdk/Network.java
danielakhterov: Huh, yea that's interesting. Will remove.

sdk/src/main/java/com/hedera/hashgraph/sdk/HederaTrustManager.java
steven-sheehy: I'm also worried if this method is implemented for different ssl implementations. A lot of the implementations I saw throw a not implemented exception. Perhaps we can double check if there's some other appropriate method we can use that's guaranteed to be implemented.

sdk/src/integrationTest/resources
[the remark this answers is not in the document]
steven-sheehy: `sdk/src/integrationTest/resources` seems like it would be a folder, not a file with text for another resources folder. Perhaps you just need to add the folder with some dummy file?

sdk/src/main/java/com/hedera/hashgraph/sdk/Network.java
danielakhterov: Thanks for the suggestion, I wasn't aware of this..

sdk/src/integrationTest/resources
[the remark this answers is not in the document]
danielakhterov: I created a soft link to `sdk/src/main/resources` which is why it appears as a file. I hoped git would be able to show hat nicely, but that doesn't seem to be the case.

PR #634, opened 2021-08-17
fix: fixed signTransaction()

Sean-Tedrow-LB:
**Checklist**


steven-sheehy:
Looks like since you're not part of the hashgraph org, your CI can't access repo secrets. We should probably split sonarcloud uploading to a conditional job. 

That means the integration tests would also fail for you since you don't have access to the OPERATOR secrets, right?

Sean-Tedrow-LB:
CI integration tests have been failing for me for a very long time.  We've just been running all of our integration tests locally.

PR #635, opened 2021-08-17
fix: fix gradle file

Sean-Tedrow-LB:
boop

### DIFF
--- a/sdk/build.gradle
+++ b/sdk/build.gradle
@@ -151,7 +151,7 @@ task integrationTest(type: Test) {
     classpath = sourceSets.integrationTest.runtimeClasspath
 
     onlyIf {
-        System.getenv('OPERATOR_ID') != null && System.getenv('OPERATOR_KEY') != null || System.getenv('CONFIG_FILE') != null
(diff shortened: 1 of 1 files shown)


sdk/build.gradle
```diff
@@ -151,7 +151,7 @@ task integrationTest(type: Test) {
     classpath = sourceSets.integrationTest.runtimeClasspath
 
     onlyIf {
-        System.getenv('OPERATOR_ID') != null && System.getenv('OPERATOR_KEY') != null || System.getenv('CONFIG_FILE') != null
+		(System.getenv('OPERATOR_ID') != null && System.getenv('OPERATOR_KEY') != null) || System.getenv('CONFIG_FILE') != null
```
steven-sheehy, line 154: How does this fix anything? `&&` has precedence over `||` and doesn't need the parentheses to be logically correct.

Also, the rest of the file uses 4 spaces for indentation and this is switching to tabs.

Sean-Tedrow-LB: Don't worry about the tabs/spaces, we're about to reformat that.

And apparently you're wrong, because my integration tests did not work before making this change, and they started working after.

PR #637, opened 2021-08-17
chore: reformat source files

Sean-Tedrow-LB:

**Description**:

In addition to reformatting source files, this also removes `finalizedBy JacocoTestReport`.  That line was causing integration tests to be run every time `build` is run.  For our workflow, we want to be able to separate unit tests and integration tests.

**Checklist**


sdk/build.gradle
```diff
@@ -1,234 +1,232 @@
...
-    // jvmArgs "-Dorg.slf4j.simpleLogger.log.com.hedera.hashgraph=trace"
+	// NOTE: Uncomment to enable trace logs in the SDK during tests
+	// jvmArgs "-Dorg.slf4j.simpleLogger.log.com.hedera.hashgraph=trace"
 
-    // this task will fail on the first failed test
-    failFast = true
+	// this task will fail on the first failed test
+	failFast = true
 
-    finalizedBy jacocoTestReport
```
steven-sheehy, line 171: That's fine, but that was to fix code coverage not being reported. Can you look into seeing why the coverage report is not being generated then?

steven-sheehy: Or at least create a ticket so it can be worked on? Capturing code coverage is important so we can continuously improve.

danielakhterov: Created issue for this: https://github.com/hashgraph/hedera-sdk-java/issues/639 If it comes down to it we might want to revert this change and simply do `./gradlew compileJava compileTestJava compileIntegrationTestJava` instead of ./gradlew build`

danielakhterov (approved):

PR #712, opened 2021-09-21
feat: add TLS configuration support

### LINKED ISSUE #672: Add support for toggling TLS connections
### Problem

At the moment the only way to enable TLS is set a network which uses `:50212`.

### Solution

Users should be able to use `Client.forPreviewnet()` and enable TLS through some method.
 - Implement `Client.setTransportSecurity(true)`

### Alternatives

_No response_

### LINKED ISSUE #710: `HederaTrustManager` should log when `certHash` is `null`
### Problem

Currently `HederaTrustManager` will blindly accept any TLS connection if the user uses `Client.forNetwork()`. That should be allowed, but at the very least we should log this in case the user was unaware any TLS connection would succeed.

### Solution

Add a logging statement when returning early from `HederaTrustManager` check verification.

### Alternatives

_No response_

### LINKED ISSUE #711: Simplify  `HederaTrustManger` hash calcuation
### Problem

The code for hashing is more verbose than it needs to be, it'd be nice to clean it up.

### Solution

Use `MessageDigest.getInstance("SHA-384").digest(certBytes)`

### Alternatives

_No response_

danielakhterov:

**Description**:

**Related issue(s)**:

Fixes #710 
Fixes #711 
Fixes #672 

**Notes for reviewer**:

**Checklist**


sdk/src/main/java/com/hedera/hashgraph/sdk/ManagedNodeAddress.java
```diff
@@ -0,0 +1,127 @@
...
+        return name != null;
+    }
+
+    @Override
+    public String toString() {
+        if (name != null) {
+            return name;
+        }
+
+        if (port == null) {
```
steven-sheehy, line 94: Port should not be allowed to be null except for in-process

sdk/src/main/java/com/hedera/hashgraph/sdk/ManagedNodeAddress.java
```diff
@@ -0,0 +1,127 @@
...
+            return address;
+        }
+
+        var s = address;
+
+        switch (port) {
+            case 5600:
+            case 433:
+                s += ":" + (transportSecurity ? 433 : 5600);
```
steven-sheehy, line 103: This is a bit of duplication. You shouldn't have a need to figure out the part if you make the class immutable.

sdk/src/main/java/com/hedera/hashgraph/sdk/HederaTrustManager.java
```diff
@@ -29,16 +35,18 @@ public void checkClientTrusted(X509Certificate[] chain, String authType) throws
     @Override
     public void checkServerTrusted(X509Certificate[] chain, String authType) throws CertificateException {
         if (certHash == null) {
+            logger.warn("skipping certificate check since no cert hash was found");
```
steven-sheehy, line 38: I think instead of just a log, a new option `Client.setVerifyCertificate()` or similar would be more explicit. By default it should be true and throw an error if no cert hash. By setting it to false, people could skip verification check and it would also allow the verification check to work for pre-prod environments.

steven-sheehy (changes requested):

steven-sheehy:
Sorry, been a bit busy as of late. Will re-review this and your sync variant PR soon.

sdk/src/test/java/com/hedera/hashgraph/sdk/ClientTest.java
```diff
@@ -134,21 +138,101 @@ void testFromJson() throws Exception {
...
+    @DisplayName("setNetwork() functions correctly")
+    void setMirrorNetworkWorks() throws Exception {
+        var nodes = new ArrayList<String>();
+        nodes.add("hcs.testnet.mirrornode.hedera.com:5600");
+
+        Client client = Client.forNetwork(new HashMap<>()).setMirrorNetwork(nodes);
+        @Var var mirrorNetwork = new HashSet<>(client.getMirrorNetwork());
+
+        Assertions.assertEquals(1, mirrorNetwork.size());
+        Assertions.assertTrue(mirrorNetwork.contains("hcs.testnet.mirrornode.hedera.com:5600"));
```
steven-sheehy, line 201:
```suggestion
        assertThat(client.getMirrorNetwork()).isEqualTo(nodes);
```

sdk/src/test/java/com/hedera/hashgraph/sdk/ClientTest.java
```diff
@@ -134,21 +138,101 @@ void testFromJson() throws Exception {
...
+
+        client.setMirrorNetwork(nodes);
+        mirrorNetwork = new HashSet<>(client.getMirrorNetwork());
+
+        Assertions.assertEquals(1, mirrorNetwork.size());
+        Assertions.assertTrue(mirrorNetwork.contains("hcs.testnet.mirrornode.hedera.com:5600"));
```
steven-sheehy, line 207:
```suggestion
        assertThat(client.getMirrorNetwork()).isEqualTo(nodes);
```

sdk/src/test/java/com/hedera/hashgraph/sdk/ClientTest.java
```diff
@@ -134,21 +138,101 @@ void testFromJson() throws Exception {
...
+
+        nodes.add("hcs.testnet1.mirrornode.hedera.com:5600");
+
+        client.setMirrorNetwork(nodes);
+        mirrorNetwork = new HashSet<>(client.getMirrorNetwork());
+
+        Assertions.assertEquals(2, mirrorNetwork.size());
+        Assertions.assertTrue(mirrorNetwork.contains("hcs.testnet.mirrornode.hedera.com:5600"));
+        Assertions.assertTrue(mirrorNetwork.contains("hcs.testnet1.mirrornode.hedera.com:5600"));
```
steven-sheehy, line 216:
```suggestion
        assertThat(client.getMirrorNetwork()).isEqualTo(newNodes);
```

sdk/src/test/java/com/hedera/hashgraph/sdk/ClientTest.java
```diff
@@ -134,21 +138,101 @@ void testFromJson() throws Exception {
...
+
+        nodes.clear();
+        nodes.add("hcs.testnet1.mirrornode.hedera.com:5600");
```
steven-sheehy, line 219: Shouldn't reuse variables. It makes it hard to follow the tests. Better to create a new variable with a descriptive name.

```suggestion
        final var deletedNode = List.of("hcs.testnet1.mirrornode.hedera.com:5600");
```

sdk/src/test/java/com/hedera/hashgraph/sdk/ClientTest.java
```diff
@@ -134,21 +138,101 @@ void testFromJson() throws Exception {
...
+
+        client.setMirrorNetwork(nodes);
+        mirrorNetwork = new HashSet<>(client.getMirrorNetwork());
+
+        Assertions.assertEquals(1, mirrorNetwork.size());
+        Assertions.assertTrue(mirrorNetwork.contains("hcs.testnet1.mirrornode.hedera.com:5600"));
```
steven-sheehy, line 225:
```suggestion
        assertThat(client.getMirrorNetwork()).isEqualTo(deleteNode)
```

sdk/src/test/java/com/hedera/hashgraph/sdk/ClientTest.java
```diff
@@ -134,21 +138,101 @@ void testFromJson() throws Exception {
...
+
+        nodes.clear();
+        nodes.add("hcs.testnet.mirrornode.hedera.com:5600");
+
+        client.setMirrorNetwork(nodes);
+        mirrorNetwork = new HashSet<>(client.getMirrorNetwork());
+
+        Assertions.assertEquals(1, mirrorNetwork.size());
+        Assertions.assertTrue(mirrorNetwork.contains("hcs.testnet.mirrornode.hedera.com:5600"));
```
steven-sheehy, line 234:
```suggestion
        final var reAddedNodes = List.of("hcs.testnet1.mirrornode.hedera.com:5600". "hcs.testnet.mirrornode.hedera.com:5600");
        client.setMirrorNetwork(nodes);
        assertThat(client.getMirrorNetwork()).isEqualTo(reAddedNode)
```

sdk/src/main/java/com/hedera/hashgraph/sdk/Network.java
```diff
@@ -159,167 +130,94 @@ NodeAddress getAddressBook(AccountId nodeAccountId) {
...
-                }
-            }
-        }
-
-        nodes.clear();
-        networkNodes.clear();
-        network.clear();
-
-        lock.release();
+    public Node getNodeForAccountId(AccountId nodeAccountId) {
```
steven-sheehy, line 220: Methods in `Network` don't need to be public. By exposing them it makes them part of the permanent API that you can't change.

steven-sheehy, line 220: The `ForAccountId` is implied by the parameter and can be omitted.

```suggestion
    public Node getNode(AccountId nodeAccountId) {
```

steven-sheehy (changes requested):
Still reviewing the network classes. Here's the feedback I have so far.

sdk/src/main/java/com/hedera/hashgraph/sdk/Network.java
danielakhterov: Maybe I'm misunderstanding, but since `Network` is package-private then all the methods are at most package-private as well even when they're declared `public`. The big reason I added an explicit scope to most methods is because I wanted to make it a conscious effort to determine if a method should be accessible by other classes or not. It is far too easy to mistakenly add a new method as package private even when it should be private and/or change a package-private method's behavior without realizing other classes rely on that method's functionality.

danielakhterov: Fair point, but the reason I added `ForAccountId` is because I expect any `getX` methods to not take any parameters as that is the Java convention, no?

sdk/src/test/java/com/hedera/hashgraph/sdk/ClientTest.java
danielakhterov: I will update this, but just wanted to know the rational behind using a single variable. Using a single variable which is manipulated makes sure `setMirroNetwork()` works correctly even when the same list pointer is passed into it several times. Perhaps the entire test shouldn't need to do this, but I feel we should have at least one test to confirms an updated list will change the current network.

steven-sheehy: Sure, you can keep one test doing that. Probably the test that verifies it is unchanged when the same network is passed into it again would make the most sense.

sdk/src/main/java/com/hedera/hashgraph/sdk/Network.java
steven-sheehy: The Java beans standards says that for exposing fields you should name the getter as `getFieldName()`. You're not exposing a field here but are performing a lookup of information so it can be named whatever you like. Omitting the `For` part allows for overriding of the method with different views indexed by a different parameter. For example `getNode(String endpoint)`. It's not a big deal here with just one parameter, but It's really bad when there are multiple parameters and people duplicate the parameter names in the method name (e.g. `getNodeByAccountIdAndEndpoint(AccountId accountId, String endpoint)`.

steven-sheehy: In general, you should always use the smallest scope possible until you have a need to widen it. Since the SDK has every class in one package (not ideal, IMO), everything should be private or package private unless the client directly uses it.

When you have a package private class with public methods, if another class that's public inherits from that class suddenly those methods are now public for use by everyone. The developer who extended the class may not have intended that since he didn't closely examine the other code that he's not touching. Here, you don't currently have another class inherit from `Network` so it's not a problem at this time, but that could change in the future and they may get mistakenly exposed. Better to reduce the scope until needed.

sdk/src/main/java/com/hedera/hashgraph/sdk/ManagedNodeAddress.java
```diff
@@ -0,0 +1,110 @@
...
+        this.name = name;
+        this.address = address;
+        this.port = port;
+    }
+
+    public static ManagedNodeAddress fromString(String string) {
+        var hostAndPortMatcher = HOST_AND_PORT.matcher(string);
+        var inProcessMatcher = IN_PROCESS.matcher(string);
+
+        if (hostAndPortMatcher.find()) {
```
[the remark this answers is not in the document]
steven-sheehy: inprocess should also be fixed

sdk/src/test/java/com/hedera/hashgraph/sdk/ClientTest.java
danielakhterov: A couple things to mention here:
 - The version of JUnit in the SDK does not have `assertThat`; we are on the latest version, but I'm guessing there are different JUnit libraries
 - `isEqualTo` would only work when `deleteNode` is exactly 1 element since `client.getMirrorNetwork()` is not guaranteed to return you a list in the same order

steven-sheehy: The `assertThat()` that I'm recommending to use is from assert4j, which I added as a dependency in one of my PRs.

I'm aware about `isEqualTo()` only working if it's the same size and order. If you don't want to guarantee a stable order of `getMirrorNetwork()` you can use `assertThat(client.getMirrorNetwork()).containsExactlyInAnyOrder(item2, item1, item3)`.

steven-sheehy (dismissed):
LGTM, but need to signoff some commits to pass DCO

steven-sheehy (approved):

steven-sheehy:
Noticed javadocs still have old args: https://github.com/hashgraph/hedera-sdk-java/blob/e2863dfdd0bbdfcd7c4e41fbffb8822cdb024111/sdk/src/main/java/com/hedera/hashgraph/sdk/ManagedNetwork.java#L26

danielakhterov:
Good catch, will create a new PR real quick to fix this.

PR #713, opened 2021-09-21
feat: implement sync variants of execute-like methods

### LINKED ISSUE #585: `Executable.execute()` should not call `Executable.onExecuteAsync()`
### Problem

The SDK currently generates all the variants of `execute` which include the async versions with callbacks, timeoutes, and the sync version. The issue is the sync version is massively slowed down by the fact it calls the async version since the async version uses completable futures.

### Solution

To get around this we should extract all the sync code from `onExecuteAsync` into other methods. Then we can implement a sync version that simply calls those extract methods.

### Alternatives

_No response_

danielakhterov:

**Description**:

**Related issue(s)**:

Fixes #585 

**Notes for reviewer**:

**Checklist**


sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
```diff
@@ -283,6 +283,17 @@ public synchronized Client setNetwork(Map<String, AccountId> network) throws Int
         return network;
     }
 
+    public synchronized void pingSync(AccountId nodeAccountId) {
```
steven-sheehy, line 286: Prefer `ping()` and `pingAll()` over explicit `Sync` suffix. Mostly because I think you should remove the `executeSync`.

sdk/src/main/java/com/hedera/hashgraph/sdk/TransactionResponse.java
```diff
@@ -37,6 +44,15 @@
             .executeAsync(client);
     }
 
+    public TransactionRecord getRecordSync(Client client) throws Exception {
+        getReceiptSync(client);
```
steven-sheehy, line 48: I don't understand why we need to get the receipt before getting a record. The record [contains](https://github.com/hashgraph/hedera-protobufs/blob/e0ee0797bacb0199fb1f2f28c42805ea0889f37c/services/transaction_record.proto#L43) the receipt. You should just be able to get the record by itself.

sdk/src/main/java/com/hedera/hashgraph/sdk/TransactionResponse.java
```diff
@@ -29,6 +29,13 @@
         this.scheduledTransactionId = scheduledTransactionId;
     }
 
+    public TransactionReceipt getReceiptSync(Client client) throws Exception {
```
steven-sheehy, line 32: Again, if we don't end up with an `executeSync` we should probably drop the `Sync` prefix on these for consistency.

sdk/src/main/java/com/hedera/hashgraph/sdk/Query.java
```diff
@@ -138,6 +143,54 @@ boolean isPaymentRequired() {
...
+            throw new IllegalStateException(
+                "`client` must have an `operator` or an explicit payment transaction must be provided");
+        }
+
+        return operator;
+    }
+
+    @Override
+    void onExecute(Client client) throws Exception {
+        initWithNodeIds(client);
```
steven-sheehy, line 159: Same comment about duplication. Should do something similar to how you avoided the duplication in `Transaction.onExecute()`.

steven-sheehy (changes requested):

sdk/src/main/java/com/hedera/hashgraph/sdk/TransactionResponse.java
mehcode: @steven-sheehy This is the same as in the async method. The short reason is that we need to get the confirmed record. It costs nothing to poll a pending receipt until its confirmed. It would cost to poll a pending record.

steven-sheehy: Ok, so it's purely a cost saving mechanism. Got it.

sdk/src/main/java/com/hedera/hashgraph/sdk/Query.java
```diff
@@ -277,6 +292,48 @@ public String toString() {
         return builder.toString();
     }
 
+    private class GrpcCostQuery {
+        Client.Operator operator;
+        Hbar cost;
+        Hbar maxCost;
+        boolean existEarly;
```
Sean-Tedrow-LB, line 299: Should be `exitEarly`

sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
```diff
@@ -156,12 +162,72 @@ public SdkRequestT setNodeAccountIds(List<AccountId> nodeAccountIds) {
...
+
+    public O execute(Client client, Duration timeout) throws TimeoutException, PrecheckStatusException {
+        Throwable lastException = null;
+
+        mergeFromClient(client);
+        onExecute(client);
+        checkNodeAccountIds();
+        setNodesFromNodeAccountIds(client);
+
+        for (int attempt = 1; attempt <= maxAttempts + 1; attempt++) {
```
steven-sheehy, line 187: This is doing one more attempt than it does for async. It's always been a bit confusing if this property is tracking "retry attempts" or "total attempts", with the initial attempt only being counted in the latter. But traditionally the async has started at one since it tracked total attempts.

sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
```diff
@@ -179,45 +245,45 @@ private void setNodesFromNodeAccountIds(Client client) {
...
 
-        var startAt = System.nanoTime();
+        return request;
+    }
 
-        return toCompletableFuture(ClientCalls.futureUnaryCall(call, request)).handle((response, error) -> {
-            var latency = (double) (System.nanoTime() - startAt) / 1000000000.0;
+    private CompletableFuture<O> executeAsync(Client client, int attempt, @Nullable Throwable lastException) {
+        if (attempt > maxAttempts) {
+            return CompletableFuture.<O>failedFuture(new Exception("Failed to get gRPC response within maximum retry count", lastException));
```
steven-sheehy, line 275: This failure differs from the synchronous execute. I like the concrete `MaxAttemptsExceededException` more than the generic `Exception`.

steven-sheehy (changes requested):

sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
```diff
@@ -291,6 +295,19 @@ public synchronized Client setNetwork(Map<String, AccountId> network) throws Int
         return network;
     }
 
+    public Void ping(AccountId nodeAccountId) {
```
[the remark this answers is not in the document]
danielakhterov: The reason we need to return `Void` here is because the originally generated method returned that. Since we override that method we need to maintain return types. If we were to remove those methods from being generated then I think that'd be considered a breaking change.

sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
```diff
@@ -291,6 +295,19 @@ public synchronized Client setNetwork(Map<String, AccountId> network) throws Int
...
+        try {
+            new AccountBalanceQuery()
+                .setAccountId(nodeAccountId)
+                .setNodeAccountIds(Collections.singletonList(nodeAccountId))
+                .execute(this);
+        } catch (Exception e) {
+            logger.debug("pining account {} failed with exception {}", nodeAccountId, e.getMessage());
```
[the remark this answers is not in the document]
danielakhterov: The point of `Client.ping` is not for the user to identify is a node is working, but for the `Network` to de-prioritize the node if it doesn't respond. However, I think adding a logging statement here does make sense. If the user sets `Client.setMaxNodeAttempts(1)` the node will be completely removed from the nodes list if it returns a single gRPC error.

sdk/src/main/java/com/hedera/hashgraph/sdk/TransactionReceiptQuery.java
```diff
@@ -120,24 +120,13 @@ ExecutionState shouldRetry(Status status, Response response) {
             case RECORD_NOT_FOUND:
                 return ExecutionState.Retry;
 
-            case SUCCESS:
```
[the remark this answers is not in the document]
danielakhterov: The previous behavior is technically a bug. This bug occurred when I did the error managed refactor, and the behavior _I thought_ was expected is any `TransactionReceiptQuery` should err if the status was not `SUCCESS`. However, after talking to @mehcode I realized that is incorrect. `TransactionReceiptQuery` itself should not err if invoked directly through `new TransactionReceiptQuery()...execute(client)`, but it should err if invoked through a `*.getReceipt(client)` method call. This also applies to `TransactionRecordQuery`.

sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
```diff
@@ -156,12 +162,72 @@ public SdkRequestT setNodeAccountIds(List<AccountId> nodeAccountIds) {
...
-                throw new IllegalStateException("Request node account IDs were not set before executing");
+    void delay(long delay) {
+        try {
+            Thread.sleep(delay);
+        } catch (InterruptedException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public O execute(Client client) throws TimeoutException, PrecheckStatusException {
```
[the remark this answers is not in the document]
danielakhterov: That is fair, but _I think_ this change is tested by a test you wrote awhile ago https://github.com/hashgraph/hedera-sdk-java/blob/0407027cb0bd665da79ae6ae33f89cc3a9ad24ce/sdk/src/test/java/com/hedera/hashgraph/sdk/MockingTest.java#L24 This tests calls the `AccountBalanceQuery.execute()` method which means it would call the overridden method in this PR. Perhaps we should add a test for the `executeAsync()` since it's been changed.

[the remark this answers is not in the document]
steven-sheehy: I didn't add `MockingTest`. I think Sean did.

But I don't think one test provides good coverage. I see you also added a test for maxattempts, so that's good. But I would like to see the tests ran in an IDE with coverage turned on. It will help you visualize which lines are green with good coverage and those that are not. 

The test you mention provides good coverage for `shouldRetryExceptionally()`, but not much else. Should also test the different outputs of `shouldRetry()` along with the different failures of get cost query.

steven-sheehy (changes requested):
Looking good. Just needs some additional unit testing coverage.

sdk/src/main/java/com/hedera/hashgraph/sdk/Query.java
```diff
@@ -138,55 +144,64 @@ boolean isPaymentRequired() {
...
+
     @Override
-    CompletableFuture<Void> onExecuteAsync(Client client) {
-        initWithNodeIds(client);
+    void onExecute(Client client) throws TimeoutException, PrecheckStatusException {
+        var grpcCostQuery = new GrpcCostQuery(client);
 
-        if ((paymentTransactions != null) || !isPaymentRequired()) {
-            return CompletableFuture.completedFuture(null);
+        if (grpcCostQuery.isRequired()) {
```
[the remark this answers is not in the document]
danielakhterov: Haha, I didn't even notice; you're absolutely right.

danielakhterov:
@steven-sheehy Unfortunately we cannot run tests in the IDE. I haven't touched it in awhile, but _iirc_ there is some error that is thrown related to `ByteBuddy` when running tests from within Intellij which don't happen when running tests from CLI. Nonetheless, will add mocking tests that verify `shouldRetry` works correctly.

steven-sheehy:
Weird, I was able to run Mocking test in Intellij. You can also look at the codecov.io coverage report thought that's slower for iteration.

steven-sheehy (dismissed):

CHANGELOG.md
```diff
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleassed
+
+### Changed
+
+ * Updated `*.[execute|getReceipt|getRecord]()` methods to not use the asynchronous version underneathe
```
Sean-Tedrow-LB, line 12: "underneathe" should be "underneath" or "under the hood"

sdk/src/integrationTest/java/FileAppendIntegrationTest.java
```diff
@@ -94,6 +94,8 @@ void canAppendLargeContentsToFile() throws Exception {
             .execute(testEnv.client)
             .getReceipt(testEnv.client);
 
+        Thread.sleep(10000);
```
Sean-Tedrow-LB, line 97: Why?

sdk/src/main/java/com/hedera/hashgraph/sdk/TransactionRecordQuery.java
```diff
@@ -115,25 +115,13 @@ ExecutionState shouldRetry(Status status, Response response) {
...
-        if (status != Status.OK) {
-            return new PrecheckStatusException(status, transactionId);
-        }
-
-        // has reached consensus but not generated
-        return new ReceiptStatusException(
-            Objects.requireNonNull(transactionId),
-            TransactionReceipt.fromProtobuf(response.getTransactionGetRecord().getTransactionRecord().getReceipt())
-        );
+    PrecheckStatusException mapStatusError(Status status, @Nullable TransactionId transactionId, Response response) {
```
Sean-Tedrow-LB, line 124: This override of `mapStatusError()` can be removed, since it is no different from the default implementation in `Executable`

sdk/src/main/java/com/hedera/hashgraph/sdk/TransactionReceiptQuery.java
```diff
@@ -120,24 +120,13 @@ ExecutionState shouldRetry(Status status, Response response) {
...
-        if (status != Status.OK) {
-            return new PrecheckStatusException(status, transactionId);
-        }
-
-        // has reached consensus but not generated
-        return new ReceiptStatusException(
-            Objects.requireNonNull(transactionId),
-            TransactionReceipt.fromProtobuf(response.getTransactionGetReceipt().getReceipt())
-        );
+    PrecheckStatusException mapStatusError(Status status, @Nullable TransactionId transactionId, Response response) {
```
Sean-Tedrow-LB, line 129: This override of `mapStatusError()` can be removed, since it is no different from the default implementation in `Executable`

steven-sheehy (dismissed):
LGTM

Sean-Tedrow-LB (dismissed):

steven-sheehy (approved):

PR #736, opened 2021-10-14
fix: add keep alive timeout to all grpc connections

### LINKED ISSUE #737: gRPC connections should set a low keep alive timeout in case nodes are unresponsive
### Description

If we don't set a keep alive timeout then unresponsive nodes can cause transaction requests to error with `TRANSACTION_EXPIRED`

### Steps to reproduce

Use Node 10 right now

### Additional context

_No response_

### Hedera network

mainnet

### Version

v2.1.0

### Operating system

Linux

danielakhterov:

**Description**:

**Related issue(s)**:

Fixes #737 

**Notes for reviewer**:

**Checklist**


danielakhterov:
Thinking more on this, we _can_ create two methods for channels, 
`getChannel(): ManagedChannel`
`getChannelAsync(): CompletableFuture<ManagedChannel>`
This would get around both issues I think and it keeps a somewhat similar API to the rest of the SDK where we have sync and async functions for the same functionality.

sdk/src/main/java/com/hedera/hashgraph/sdk/ManagedNode.java
```diff
@@ -228,13 +237,48 @@ synchronized ManagedChannel getChannel() {
...
         return channel;
     }
 
+    boolean channelFailedToConnect() {
+        if (hasConnected) {
+            return false;
+        }
+        @Var var state = getChannel().getState(true);
+        try {
+            for (@Var int i = 0; i < GET_STATE_MAX_ATTEMPTS && state != ConnectivityState.READY; i++) {
```
danielakhterov, line 254: Is this one of those rare case where a do-while loop would work?

sdk/src/main/java/com/hedera/hashgraph/sdk/ManagedNode.java
```diff
@@ -228,13 +237,48 @@ synchronized ManagedChannel getChannel() {
...
+                TimeUnit.MILLISECONDS.sleep(GET_STATE_INTERVAL_MILLIS);
+                state = getChannel().getState(true);
+            }
+        } catch (InterruptedException e) {
+            throw new RuntimeException(e);
+        }
+        hasConnected = (state == ConnectivityState.READY);
```
danielakhterov, line 261: Can probably be moved into the loop and and simply returned at the end there?

sdk/src/main/java/com/hedera/hashgraph/sdk/ManagedNode.java
Sean-Tedrow-LB: Needs to keep track of number of attempts, doesn't it?

danielakhterov:
Can't review as I was the one who started this PR, but I am not the one who finished it. Nonetheless, LGTM.

Sean-Tedrow-LB (approved):

PR #778, opened 2021-11-18
Doc: amend file transaction documentation

### LINKED ISSUE #759: FileAppend docs state it doesn't chunk
### Description

The `FileAppendTransaction` for all the `setData` methods state incorrectly that the user needs to manually chunk, but it internally already chunks automatically as needed. The docs should be updated for all methods to not mention that a `TRANSACTION_OVERSIZE` might occur or that user needs to chunk.

https://github.com/hashgraph/hedera-sdk-java/blob/20bc144e02bb7fb106faf4a0c95e2adcdedfc68c/sdk/src/main/java/com/hedera/hashgraph/sdk/FileAppendTransaction.java#L92

Also, the `FileCreateTransaction` and `FileUpdateTransaction` should be updated to point users to the `FileAppendTransaction` if the file exceeds 6KiB and need automatic chunking support.

### Steps to reproduce

Look at the `FileAppendTransaction` javadocs.

### Additional context

_No response_

### Hedera network

other

### Version

v2.3.0

### Operating system

macOS

Sean-Tedrow-LB:
**Description**:

Amends documentation to reflect that `FileAppendTransaction` now chunks.

**Related issue(s)**:

Fixes #759


sdk/src/main/java/com/hedera/hashgraph/sdk/FileUpdateTransaction.java
```diff
@@ -147,9 +146,8 @@ public FileUpdateTransaction setContents(byte[] bytes) {
      * Note that total size for a given transaction is limited to 6KiB (as of March 2020) by the
      * network; if you exceed this you may receive a  {@link Status#TRANSACTION_OVERSIZE}.
      * <p>
-     * In this case, you will need to break the data into chunks of less than ~6KiB and execute this
-     * transaction with the first chunk and then use {@link FileAppendTransaction} with
-     * {@link FileAppendTransaction#setContents(String)} for the remaining chunks.
+     * In this case, you will need to keep the initial file contents under ~6KiB and
```
steven-sheehy, line 149: This and create are still recommending to manually chunk the first item. Instead, it should just say to use the file append tx to automatically split arbitrarily large files into smaller chunks and submit them to the network.

Sean-Tedrow-LB: I can make that adjustment for `FileCreateTransaction`, but you're going to encounter a bit of an issue with `FileUpdateTransaction`

If I'm reading this correctly, `FileUpdateTransaction` replaces the contents of the file only if the `contents` field is set.  However, the default value of a `bytes` field is "empty bytes" per [the documentation](https://developers.google.com/protocol-buffers/docs/proto3#default), so unless I'm missing something, it's impossible to use a `FileUpdateTransaction` to set the contents of a file to empty bytes, since that's indistinguishable from not-set.

steven-sheehy (approved):

danielakhterov:
Merging as there were no changes made since last approving review.

PR #782, opened 2021-11-19
Fix: Remove streams

### LINKED ISSUE #735: Remove any use of streams in the SDK
### Description

We should not be using any stream API in the SDK because it is not compatible with Java 7 because the earliest stream support in Java came in Java 8. We don't have any tests that show this (and we really should), but at the moment our JDK 7 SDK releases are likely broken because we use streams a few places.

### Steps to reproduce

Use SDK within a JDK 7 context

### Additional context

_No response_

### Hedera network

other

### Version

v2.1.0

### Operating system

Linux

Sean-Tedrow-LB:
**Description**:

Aims to remove the use of streams from the SDK, which would allow us to remove a backport dependency and dramatically shrink the size of the SDK.

**Related issue(s)**:

Fixes #735

**Notes for reviewer**:

Unfortunately, we seem to have inadvertently exposed one of the backports publicly: `TopicMessageQuery.setRetryHandler()` takes a `Predicate`.  We are trying to target Java 7 for the sake of Android development, but `Predicate` is Java 8, so we were using a backport of `Predicate` here.  We need to decide what to do about this before going forward.

**Checklist**

### DIFF
--- a/examples/src/main/java/TopicWithAdminKeyExample.java
+++ b/examples/src/main/java/TopicWithAdminKeyExample.java
@@ -12,7 +12,6 @@ import com.hedera.hashgraph.sdk.TopicUpdateTransaction;
 import com.hedera.hashgraph.sdk.Transaction;
 import com.hedera.hashgraph.sdk.TransactionResponse;
 import io.github.cdimascio.dotenv.Dotenv;
-import java8.util.J8Arrays;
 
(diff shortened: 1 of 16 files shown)


danielakhterov:
Unfortunately it doesn't look like we can actually do this as we discussed before, so closing.

PR #856, opened 2022-01-21
Fix TLS issues with mirror node

### LINKED ISSUE #855: Unable to connect to mainnet-public mirror node
### Description

In a previous release, the SDK detected the presence of port 443 and defaulted to TLS. With the refactoring for `setTransportSecurity()` that behavior was removed or broken. However, there's no way to manually enable transport security for the mirror network. `Client.setTransportSecurity()` only enables it for the main network.

We could enable both mirror and main networks when `Client.setTransportSecurity(true)` but there are scenarios where one is TLS and the other is not. 

### Steps to reproduce

```java
    Client client = Client.forNetwork(Map.of());
    client.setTransportSecurity(true);
    client.setMirrorNetwork(List.of("mainnet-public.mirrornode.hedera.com:443"));
    new TopicMessageQuery()...subscribe(...);
```

### Additional context

_No response_

### Hedera network

other

### Version

2.6.0

### Operating system

macOS

steven-sheehy:
**Description**:
* Fix incorrect port used for TLS detection
* Fix use of hard coded `overrideAuthority("127.0.0.1")` which doesn't work with the CA signed certificates the mirror node uses

**Related issue(s)**:

Fix #855

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/sdk/src/main/java/com/hedera/hashgraph/sdk/ManagedNode.java
+++ b/sdk/src/main/java/com/hedera/hashgraph/sdk/ManagedNode.java
@@ -91,6 +91,10 @@ abstract class ManagedNode<N extends ManagedNode<N, KeyT>, KeyT> implements Comp
         this.useCount = node.useCount;
     }
 
+    protected String getAuthority() {
+        return "127.0.0.1";
+    }
+
     /**
      * Create an insecure version of this node
      *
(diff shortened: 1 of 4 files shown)


steven-sheehy:
This same fix needs to be ported to go and js SDKs

steven-sheehy:
There's 3 problems:

* SDKs use incorrect standard TLS port 433
* SDKs use hard coded `overrideAuthority("127.0.0.1")` which doesn't work with the CA signed  certificates the mirror node uses
* SDKs don't provide a way to enable TLS for mirror nodes if it's not using the standard TLS port

This PR only fixes the first two issues. The third is less important as Hedera mirror nodes have only ever used 443 and 5600 ports.

danielakhterov (approved):

PR #876, opened 2022-02-10
Fix: checksums

### LINKED ISSUE #825: Checksums are broken WRT LedgerIds
### Description

Checksum logic in the SDK is programmed to match the behavior of [this demo](https://codepen.io/lbaird/full/QWKRPGv), which is linked from the [HIP-15 text](https://github.com/hashgraph/hedera-improvement-proposal/blob/2e9cedf24891c473f77400864bdac104085c93e8/HIP/hip-15.md).

According to the HIP:
```
h = unsigned byte array containing the ledger ID followed by 6 zero bytes
```
We currently attempt to find `h` by converting the `LedgerId` to string via `Integer.toString()`, and then concatenating this with twelve 0s, and then interpreting the result as the hex encoding of a byte array.  This seemed good enough when we were working with a `NetworkName` enum which merely stored three hard-coded values, 0 for mainnet, 1 for testnet, 2 for previewnet.

But this actually hid an off-by-one error, and as a consequence, the resulting `h` value for a `LedgerId` of `{0x01}` is the same as the resulting `h` value for a LedgerId of `{0x10}`.

1. `1` gets concatenated with 12 zeros to become `1_00_00_00_00_00_00`
2. `1_00_00_00_00_00_00` is interpreted from left to right as a sequence of pairs of hex digits to become a byte array, `10_00_00_00_00_00_0 -> {0x10, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00}`

### Steps to reproduce

Run [this unit test](https://github.com/hashgraph/hedera-sdk-java/blob/566fdc65dc2d5621db40318b84df4b12f39c7a51/sdk/src/test/java/com/hedera/hashgraph/sdk/AccountIdTest.java#L56-L59) and watch it fail.

### Additional context

_No response_

### Hedera network

other

### Version

v2.5.0

### Operating system

Linux

Sean-Tedrow-LB:
**Description**:
Fixes #825 

**Checklist**


danielakhterov:
The changes look good, but I think we're waiting on making this a sprint task before we merge this.

danielakhterov (dismissed):

danielakhterov (approved):

PR #877, opened 2022-02-11
Feat: Contract create flow

### LINKED ISSUE #839: Streamline contract creation flow
### Problem

Developers wanting to create smart contracts currently have to 
* Create a file (optionally using file append) with the contract's bytecode
* Create a contract from the resulting file

### Solution

Add a builder to the contractCreate transaction that accepts bytecode such that if specified, the createContractTransaction.execute does

* Create file (with admin key for append if necessary) from bytecode
* Appends n times to complete bytecode upload
* Creates the contract
* Upon querying for a receipt or record for the operation, the contractId should be returned. The information for any file append transactions or file create transactions will not be returned. If a user would like to have this information the user can submit the transactions individually. 
* to|from bytes will not be supported

Note: This should continue to support the specification of contract constructor parameters

- [x] Design feature
- [x] Implement

### Alternatives

Continue to use current flow which is less than optimal for developer experience

Sean-Tedrow-LB:
**Description**:
Adds `ContractCreateFlow` and tests for it

**Related issue(s)**:

Fixes #839

**Checklist**


sdk/src/main/java/com/hedera/hashgraph/sdk/ContractCreateFlow.java
```diff
@@ -0,0 +1,342 @@
+package com.hedera.hashgraph.sdk;
```
steven-sheehy, line 1: Can we put this in a `flow` package to encapsulate and aid in discoverability? The sdk package is already big enough.

sdk/src/main/java/com/hedera/hashgraph/sdk/ContractCreateFlow.java
```diff
@@ -0,0 +1,342 @@
...
+     */
+    public ContractCreateFlow setNodeAccountIds(List<AccountId> nodeAccountIds) {
+        Objects.requireNonNull(nodeAccountIds);
+        this.nodeAccountIds = nodeAccountIds;
+        return this;
+    }
+
+    private void splitBytecode() {
+        if(bytecode.length() > FILE_CREATE_MAX_BYTES) {
+            createBytecode = bytecode.substring(0, FILE_CREATE_MAX_BYTES);
```
steven-sheehy, line 244: The bytecode can be arbitrarily large and may not fit in just create + append. This should loop over the bytecode and create as many appends as necessary. I don't think the `createBytecode` and `appendBytecode` fields are appropriate as the chunking and looping should be done inline.

sdk/src/main/java/com/hedera/hashgraph/sdk/ContractCreateFlow.java
```diff
@@ -0,0 +1,342 @@
...
+        var fileCreateTx = new FileCreateTransaction()
+            .setKeys(Objects.requireNonNull(client.getOperatorPublicKey()))
+            .setContents(createBytecode);
+        if (nodeAccountIds != null) {
+            fileCreateTx.setNodeAccountIds(nodeAccountIds);
+        }
+        return fileCreateTx;
+    }
+
+    private FileAppendTransaction spawnFileAppendTransaction(FileId fileId) {
```
steven-sheehy, line 262: I think spawn is the wrong terminology as in technology that's usually used to indicate a process has been forked and started. Just `fileAppendTransaction()` or `getFileAppendTransaction()` would fit better. Same for other spawn methods.

steven-sheehy (changes requested):

sdk/src/main/java/com/hedera/hashgraph/sdk/ContractCreateFlow.java
Sean-Tedrow-LB: The append transaction is chunked.  It will automatically chunk the contents.

sdk/src/main/java/com/hedera/hashgraph/sdk/ContractCreateFlow.java
Sean-Tedrow-LB: How about `make*()`?

Sean-Tedrow-LB: Or `create*()`?

steven-sheehy: That works.

sdk/src/main/java/com/hedera/hashgraph/sdk/ContractCreateFlow.java
danielakhterov: I do agree that this package is big enough, but the reason we put everything in one package to begin with was because of Java's package-private scope issues and that's not an issue I want to resurface.

I just pulled this branch and tried to see if a `flow` package would work in the first place since I had doubts, and it doesn't since we need access to `ContractFunctionParams.toBytes()` which is package-private.

danielakhterov (approved):

Issue #818, opened 2021-12-29
Implement `[Private|Public]Key.is[ED25519|ECDSA]()`


danielakhterov:
### Problem

At the current moment it's a bit cumbersome to determine if a key is `ED25519` or `ECDSA` and it'd be nice if there was an easier way to check which key type it is.

### Solution

Implement the methods.

PR #882, opened 2022-02-15
implement #818 [Private|Public]Key.is[ED25519|ECDSA]()

rickyrodmac:
**Description**:

Problem

At the current moment it's a bit cumbersome to determine if a key is ED25519 or ECDSA and it'd be nice if there was an easier way to check which key type it is.

This PR modifies the PrivateKey & PublicKey classes to add:
* isED25519()
* isECDSA()
* 

**Related issue(s)**:

Fixes #881 

**Notes for reviewer**:

**Checklist**


sdk/src/main/java/com/hedera/hashgraph/sdk/PrivateKey.java
```diff
@@ -133,6 +133,16 @@ public static PrivateKey fromBytesDER(byte[] privateKey) {
         return PrivateKey.fromPrivateKeyInfo(PrivateKeyInfo.getInstance(privateKey));
     }
 
+    public static Boolean isED25519(byte[] privateKey) {
```
danielakhterov, line 136: Hmm, I think it might be better to make these methods abstract and by default return `false` and perhaps they can be updated in `PrivateKeyEd25519` and PrivateKeyECDSA` to return `true` for their variant.

cc @Sean-Tedrow-LB

danielakhterov (changes requested):

PR #892, opened 2022-02-24
Add ExchangeRates + improvements to ExchangeRate

### LINKED ISSUE #891: Add ExchangeRateSet to the SDK so that file 0.0.112 can be parsed
### Problem

The file `0.0.112` contains exchange rate information, it's not possible to parse it with the SDK.

### Solution

Add an `ExchangeRateSet` class to enable parsing of the file

### Alternatives

_No response_

gregscullard:
**Description**:
Adds an `ExchangeRates` class to support parsing file `0.0.112` containing exchange rates.
Also adds a calculated property to `ExchangeRate` for ease of use.

**Related issue(s)**:
Fixes #891

**Checklist**


sdk/src/test/java/com/hedera/hashgraph/sdk/ExchangeRateSetTest.java
```diff
@@ -0,0 +1,32 @@
...
+import com.google.protobuf.InvalidProtocolBufferException;
+import org.bouncycastle.util.encoders.Hex;
+import org.junit.jupiter.api.Test;
+
+import org.threeten.bp.Instant;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class ExchangeRateSetTest {
+    private static final String exchangeRateSetHex = "0a1008b0ea0110b6b4231a0608f0bade9006121008b0ea01108cef231a060880d7de9006";
```
danielakhterov, line 12: Feel like we should use a snapshot test here instead.

danielakhterov (commented):
I'll let @Sean-Tedrow-LB complete this review, but I feel we should rename `ExchabgeRateSet`  since the `Set` suffix really throws me off. Maybe just `ExhangeRates`?

danielakhterov:
Can you target develop instead of main we want all new features to hit develop first.

sdk/src/test/java/com/hedera/hashgraph/sdk/ExchangeRateSetTest.java
gregscullard: Not sure how to go about this, might need some assistance.

gregscullard:
> Can you target develop instead of main we want all new features to hit develop first.

Done

gregscullard:
> I'll let @Sean-Tedrow-LB complete this review, but I feel we should rename `ExchabgeRateSet` since the `Set` suffix really throws me off. Maybe just `ExhangeRates`?

Renamed to `ExchangeRates`

danielakhterov (commented):
Changes look good, but something I only now remembered is our `TransactionReceipt` doesn't decode a `ExchangdRates`. Should this PR update that as well?

gregscullard:
Imo. The current exchange rate for a receipt makes sense. No need to change

Envoyé depuis mon iPhone/Sent from my iPhone

PR #893, opened 2022-02-24
Feat: Verbose Log Node Retries(#801)

rickyrodmac:
Signed-off-by: Ricky Rodriguez <rickyrod@mac.com>

**Description**:
Adds additional log line with Node IP, Timestamp, and Transaction Type.

**Notes for reviewer**:

I have not been able to get the logger configured on my system to capture the output directly.

Any hints are greatly appreciated.

**Checklist**


sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
```diff
@@ -517,6 +517,16 @@ public Node getNode() {
         }
 
         public ClientCall<ProtoRequestT, ResponseT> createCall() {
+            // SECTION 1
+            System.out.format("Node: %s", node.address != null ? node.address.getAddress().toString() : "NULL");
```
danielakhterov, line 521: You should be using `logger.trace` instead of `System.out.format`

danielakhterov, line 521: Since this seems to be copy and pasted a couple times maybe we want to create a logging method instead and just invoke that?

danielakhterov (changes requested):

sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
rickyrodmac: Right below the System.out lines is the logger.trace. I intend to remove the System.out lines once I can get logging on my system working.

rickyrodmac: I agree that would be much better.

danielakhterov (approved):

PR #898, opened 2022-02-25
Feat: Address book query

Sean-Tedrow-LB:
**Related issue(s)**:

Fixes #805

**Notes for reviewer**:

We may want to add mocking tests to ensure that it retries as expected.

**Checklist**

- [ ] Documented (Code comments, README, etc.)
- [X] Tested (unit, integration, etc.)

PR #900, opened 2022-02-26
Implement `MirrorQuery`

danielakhterov:
**Description**:

**Related issue(s)**:

#898 

**Notes for reviewer**:

This primarily de-duplicates code between `AddressBookQuery` and `TopicMessageQuery`. However, it does a little more than that. It merges both of these queries into one which means both will now contain `execute`, `executeAsync`, `subscribe`, `setListener`. Admittedly some of these methods don't make sense to exist on each query; for instance why would you subscribe to `AddressBookQuery`, but I don't see why we can't support such functionality. While I was working on this I often glanced at `Executable`, and ended up with a very similar API to `Executable`, so that brings the question of if we can push this further and create a base class for `Executable` and `MirrorQuery`; although I don't know what we would call that class. That hypothetical class could contain all the backoff logic along with the `makeRequest` and `mapResponse` abstract methods and perhaps more that I'm forgetting about right now.

**Checklist**


danielakhterov:
That's a good point. I'm not really sure on what the API for a bi-directional stream would be though. Is that all we need to consider though: stream, bi-directional stream, and a simple response? If so, do you have any suggestions on API design here?

steven-sheehy:
All gRPC APIs use the `StreamObserver` interface to return responses. For all requests, they use `ClientCall.sendMessage()`. So ideally our APIs would mimic those in some fashion.

That said, I think we can just focus our design on server streaming and unary request/response as the most likely to be implemented. Since request/response is just a specialization of server streaming that only returns one response, we can expose the same concepts as the current TopicMessageQuery while providing simpler `Res execute(Req req)` methods that hide the callback methods.

If we ever add client streaming or bidi, then we would add some interface that mimics parts of `ClientCall` that we return from the execute methods.

rickyrodmac (approved):
In what little experience I have with the SDK these changes look good.

danielakhterov:
After having several discussions with others at LB I'm not too sure if the appoarch this PR takes is the right now. Don't get me wrong, I do enjoy deduplicating code, but it also increases the barrier to entry making it much harder for new employees to understand what's happening and be able to modify it accordingly. I'm not closing this PR just yet, but just wanted to get my thoughts out.

danielakhterov:
Closing this for now, can re-open later if we choose this to be the best option.

PR #904, opened 2022-03-05
chore: update gradle, comment out nexus/maven plugins, make compatible with java17

mehcode:


danielakhterov:
Merging this in without a review since I was able to test it locally and publishing with JDK 17 worked fine.

PR #926, opened 2022-03-19
Fix: Verify body bytes

### LINKED ISSUE #823: fromBytes() should ensure all transactions in a TransactionList message are as expected
### Problem

The current implementation of `Transaction.fromBytes()` _assumes_ that all transactions with matching `TransactionId`s in a given `TransactionList` protobuf message are functionally identical transactions, and it takes a short cut by deserializing only the first one.

This is a security vulnerability: if we use `Transaction.[to|from]Bytes()` to pass a transaction around between client programs for signing, the client programs should be able to inspect the deserialized SDK `Transaction` to know what they're signing.  However, if, for example, the second transaction body in the `TransactionList` does not match the first one (and is perhaps malicious), a client might consent to and `sign()` the first `Transaction` in the `TransactionList`, and in the process unwittingly consent to and `sign()` the second one.

### Solution

`Transaction.fromBytes()` should check whether transactions with the same `TransactionId` in a `TransactionList` are functionally identical, and throw an exception if they aren't.

### Alternatives

_No response_

Sean-Tedrow-LB:
**Related issue(s)**:

Fixes #823

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
+ * `Transaction.fromBytes()` now verifies that transaction bodies in transaction list match.
+
 ## v2.10.0-beta.1
 
 ### Added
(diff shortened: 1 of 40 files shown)


danielakhterov (approved):
Wow, the Java reflection looks horrible, but at least it makes it easier to do this task without implementing verification in each transaction type.

rickyrodmac (commented):
Looks good.

PR #989, opened 2022-05-03
feat: retrieve a record even if a transaction failed

### LINKED ISSUE #969: Enable the SDK to retrieve a record even if a transaction failed
### Problem

Right now, getRecord does a getReceipt first and if the status isn’t SUCCESS, it throws an exception.

There are cases when being able to get a record for a failed transaction does make sense, and it would be desirable to be able to override the exception being thrown.

### Solution

Add support for this feature. 

### Alternatives

_No response_

rickyrodmac:
Signed-off-by: Ricky Rodriguez <ricky.rodriguez@launchbadge.com>

**Description**:

Added:
`forceGetReceipt`
`forceGetReceiptAsync`
`forceGetRecord`
`forceGetRecordAsync`

Fixes #969 

**Notes for reviewer**:
I'm not sure how this can be tested.

**Checklist**


sdk/src/main/java/com/hedera/hashgraph/sdk/TransactionResponse.java
```diff
@@ -62,6 +62,13 @@ public TransactionReceipt getReceipt(Client client) throws TimeoutException, Pre
         return receipt;
     }
 
+    public TransactionReceipt forceGetReceipt(Client client) throws TimeoutException, PrecheckStatusException {
+        return new TransactionReceiptQuery()
+            .setTransactionId(transactionId)
+            .setNodeAccountIds(Collections.singletonList(nodeId))
+            .execute(client);
+    }
+
```
Sean-Tedrow-LB, line 71: You know what?  I am dumb.  I'm pretty sure the async versions aren't necessary.

sdk/src/main/java/com/hedera/hashgraph/sdk/TransactionResponse.java
```diff
@@ -79,6 +93,15 @@ public TransactionRecord getRecord(Client client) throws TimeoutException, Prech
             .execute(client);
     }
 
+    public TransactionRecord forceGetRecord(Client client) throws TimeoutException, PrecheckStatusException {
+        forceGetReceipt(client);
```
Sean-Tedrow-LB, line 97: You don't need to get receipt here.

The only reason we `getReceipt()` in `getRecord()` is because `TransactionReceiptQuery` is free, where `TransactionRecordQuery` is _not_ free.  The intended behavior of `getRecord()` was to throw an exception if the status is not `SUCCESS`, so if we can determine that the status is not `SUCCESS` by using a free query, it makes sense to do that and save the user from paying for a record query in that case.

For `forceGetRecord()`, it doesn't make sense to check the receipt before getting the record, since this is never going to throw the exception; the user wants to pay for the `TransactionRecordQuery` and get the record regardless of the status.  You're basically just getting the receipt and then immediately throwing it away without doing anything with it.

Sean-Tedrow-LB (changes requested):

sdk/src/main/java/com/hedera/hashgraph/sdk/TransactionResponse.java
rickyrodmac: Removed

rickyrodmac (commented):
Removed the async versions and removed the forcedGetRecript from forcedGetRecord

Sean-Tedrow-LB (dismissed):

danielakhterov (approved):

PR #1003, opened 2022-05-23
feat: implement `EthereumTransactionData`

danielakhterov:
Signed-off-by: Daniel Akhterov <daniel@launchbadge.com>**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/sdk/src/main/java/com/hedera/hashgraph/sdk/EthereumFlow.java
+++ b/sdk/src/main/java/com/hedera/hashgraph/sdk/EthereumFlow.java
@@ -44,14 +44,14 @@ public class EthereumFlow implements WithExecute<TransactionResponse> {
     private static FileId createFile(byte[] callData, Client client) throws PrecheckStatusException, TimeoutException {
         try {
             var transaction = new FileCreateTransaction()
-                    .setContents(Arrays.copyOfRange(callData, 0, Math.min(4096, callData.length)))
+                    .setContents(Arrays.copyOfRange(callData, 0, Math.min(FileAppendTransaction.DEFAULT_CHUNK_SIZE, callData.length)))
                     .execute(client);
             var fileId = transaction.getReceipt(client).fileId;
 
-            if (callData.length > 4096) {
+            if (callData.length > FileAppendTransaction.DEFAULT_CHUNK_SIZE) {
                 new FileAppendTransaction()
                         .setFileId(fileId)
-                        .setContents(Arrays.copyOfRange(callData, 4096, callData.length))
+                        .setContents(Arrays.copyOfRange(callData, FileAppendTransaction.DEFAULT_CHUNK_SIZE, callData.length))
                         .execute(client)
                         .getReceipt(client);
             }
@@ -65,14 +65,14 @@ public class EthereumFlow implements WithExecute<TransactionResponse> {
 
     private static CompletableFuture<FileId> createFileAsync(byte[] callData, Client client) {
         return new FileCreateTransaction()
-                .setContents(Arrays.copyOfRange(callData, 0, Math.min(4096, callData.length)))
+                .setContents(Arrays.copyOfRange(callData, 0, Math.min(FileAppendTransaction.DEFAULT_CHUNK_SIZE, callData.length)))
                 .executeAsync(client)
                 .thenCompose((response) -> response.getReceiptAsync(client))
                 .thenCompose((receipt) -> {
-                    if (callData.length > 4096) {
+                    if (callData.length > FileAppendTransaction.DEFAULT_CHUNK_SIZE) {
                         return new FileAppendTransaction()
                                 .setFileId(receipt.fileId)
-                                .setContents(Arrays.copyOfRange(callData, 4096, callData.length))
+                                .setContents(Arrays.copyOfRange(callData, FileAppendTransaction.DEFAULT_CHUNK_SIZE, callData.length))
                                 .executeAsync(client)
                                 .thenApply((r) -> receipt.fileId);
                     } else {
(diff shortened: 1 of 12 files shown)


sdk/src/main/java/com/hedera/hashgraph/sdk/EthereumFlow.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    private static FileId createFile(byte[] callData, Client client) throws PrecheckStatusException, TimeoutException {
+        try {
+            var transaction = new FileCreateTransaction()
+                    .setContents(Arrays.copyOfRange(callData, 0, Math.min(4096, callData.length)))
+                    .execute(client);
+            var fileId = transaction.getReceipt(client).fileId;
+
+            if (callData.length > 4096) {
```
Nana-EC, line 51: Shouldn't this add as many appends as needed?
What if the file since is 9000? You'd need a Create with 2 appends.

Nana-EC: Same for `createFileAsync`

sdk/src/main/java/com/hedera/hashgraph/sdk/EthereumFlow.java
```diff
@@ -0,0 +1,133 @@
...
+
+    public EthereumFlow setMaxGasAllowance(Hbar maxGasAllowance) {
+        this.maxGasAllowance = maxGasAllowance;
+        return this;
+    }
+
+    private static FileId createFile(byte[] callData, Client client) throws PrecheckStatusException, TimeoutException {
+        try {
+            var transaction = new FileCreateTransaction()
+                    .setContents(Arrays.copyOfRange(callData, 0, Math.min(4096, callData.length)))
```
Nana-EC, line 47: Set static fields for `4096` and `5120`
Names should convey the relevance of the byte size threshold

sdk/src/main/java/com/hedera/hashgraph/sdk/EthereumTransactionDataEip1559.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    public static EthereumTransactionDataEip1559 fromBytes(byte[] bytes) {
+        var decoder = RLPDecoder.RLP_STRICT.sequenceIterator(bytes);
+        var rlpItem = decoder.next();
+
+        // typed transaction?
+        byte typeByte = rlpItem.asByte();
+        if (typeByte != 2) {
+            throw new IllegalArgumentException("only EIP1559 is supported");
```
Nana-EC, line 63: nit: note the actual byte to add troubleshooting

sdk/src/main/java/com/hedera/hashgraph/sdk/EthereumTransactionDataLegacy.java
```diff
@@ -0,0 +1,83 @@
...
+        byte[] s
+    ) {
+        super(callData);
+
+        this.nonce = nonce;
+        this.gasPrice = gasPrice;
+        this.gasLimit = gasLimit;
+        this.to = to;
+        this.value = value;
+        this.v = v;
```
Nana-EC, line 40: note: I believe recoveryId may be present and can be extracted from v
```java
        var v = rlpItems.get(6).asBytes();
        BigInteger vBi = new BigInteger(1, v);
        this.recId = vBi.testBit(0) ? 0 : 1
```

Nana-EC (commented):
Some thoughts and suggestions

sdk/src/main/java/com/hedera/hashgraph/sdk/EthereumFlow.java
danielakhterov: `FileAppendTransaction` is implicitly chunked; meaning it will create N `FileAppendTransactions` internally when executing the request. The only reason we did the initial first chunk manually is because we didn't want to create an empty file and *always* submit a `FileAppendTransaction`.

sdk/src/main/java/com/hedera/hashgraph/sdk/EthereumFlow.java
danielakhterov: That's fair, also I should double check the 5120 is correct; for some reason I remember that number, but don't know where it came from.

sdk/src/main/java/com/hedera/hashgraph/sdk/EthereumTransactionDataLegacy.java
danielakhterov: Where would it be present? I went off https://github.com/hashgraph/hedera-services/blob/ac677f00931a672d077069cbd44a8828d54017a5/hapi-utils/src/main/java/com/hedera/services/ethereum/EthTxData.java#L98 which doesn't seem to ever check for a potential `recoveryId`.

sdk/src/main/java/com/hedera/hashgraph/sdk/EthereumFlow.java
Nana-EC: Ah good to know. No-op then

sdk/src/main/java/com/hedera/hashgraph/sdk/EthereumTransactionDataLegacy.java
Nana-EC: Pointing it out in case you want to include both `recoveryId` and `chainId` in your `toString()` like the `EthTxData.java` example you provided does.
If not then this is a no-op

danielakhterov: I seem to have misunderstood what you meant, I went ahead and added `recoveryId` and `chainId` to `*Legacy`.

PR #1013, opened 2022-06-01
Simplifying our publishing procedure to eliminate the need to switch branches

danielakhterov:
These changes allow us to run `just publish` to publish both jdk7 and jdk9 artifacts through the use of search replace. Tested manually that this works.


danielakhterov:
Merging this in without a review because I was able to test this locally, and it seems to be publishing correctly to sonatype.

PR #1040, opened 2022-06-29
WIP: implement RngTransaction

### LINKED ISSUE #1035: HIP-351 : RandomGenerate transaction
### Problem

The SDK needs add support for HIP-351 : RandomGenerate transaction.

Protobuf PR: https://github.com/hashgraph/hedera-protobufs/pull/209/files

Add reference implementation to SDK repo

### Solution

Add support for HIP-351 based on reference implementation.

### Alternatives

_No response_

rickyrodmac:
Signed-off-by: Ricky Rodriguez <ricky.rodriguez@launchbadge.com>

**Description**:
Implement the RngTransaction

**Related issue(s)**:

Fixes #1035, #1035

**Notes for reviewer**:
I need help with this.

**Checklist**


sdk/src/main/java/com/hedera/hashgraph/sdk/RngTransaction.java
```diff
@@ -0,0 +1,91 @@
...
+    // TODO: Need to implement the following methods
+
+    @Override
+    void onFreeze(TransactionBody.Builder bodyBuilder) {
+
+    }
+
+    @Override
+    void onScheduled(SchedulableTransactionBody.Builder scheduled) {
+
```
danielakhterov, line 76:
```suggestion
        throw new UnsupportedOperationException("cannot schedule RngTransaction");
```

sdk/src/main/java/com/hedera/hashgraph/sdk/RngTransaction.java
```diff
@@ -0,0 +1,91 @@
...
+ * Random Number Generator Transaction.
+ */
+public final class RngTransaction extends com.hedera.hashgraph.sdk.Transaction<RngTransaction> {
+
+    /**
+     * If provided and is positive, returns a 32-bit pseudorandom number from the given range in the transaction record.
+     * If not set or set to zero, will return a 384-bit pseudorandom number in the record.
+     */
+    @Nullable
+    public Integer range = null;
```
danielakhterov, line 40: Should be private

sdk/src/main/java/com/hedera/hashgraph/sdk/RngTransaction.java
```diff
@@ -0,0 +1,91 @@
...
+    public RngTransaction setRange(Integer range) {
+        this.range = range;
+        return this;
+    }
+
+    // TODO: Need to implement the following methods
+
+    @Override
+    void onFreeze(TransactionBody.Builder bodyBuilder) {
+
```
danielakhterov, line 71: I don't know the exact method, but it'll be something like
```java
        bodyBuilder.setRandomGenerate(build());
```
But then you'll also need to implement `build()` which should look something like
```java
    RandomGenerateTransactionBody.Builder build() {
        var builder = RandomGenerateTransactionBody.newBuilder();

        if (range != null) {
            builder.setRange(range);
        }

        return builder;
    }
```

sdk/src/main/java/com/hedera/hashgraph/sdk/RngTransaction.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @Override
+    void validateChecksums(Client client) throws BadEntityIdException {
+        // validateChecksums(client);
+    }
+
+    @Override
+    MethodDescriptor<com.hedera.hashgraph.sdk.proto.Transaction, TransactionResponse> getMethodDescriptor() {
+        return null;
```
danielakhterov, line 86: Should look something like
```suggestion
        return UtilService.getRandomGenerateMethod();
```
To show where I got this information check `util_service.proto`
Speaking of which, it looks like you may need to update protobufs again?

sdk/src/main/java/com/hedera/hashgraph/sdk/RngTransaction.java
```diff
@@ -0,0 +1,91 @@
...
+import javax.annotation.Nullable;
+
+/**
+ * Random Number Generator Transaction.
+ */
+public final class RngTransaction extends com.hedera.hashgraph.sdk.Transaction<RngTransaction> {
+
+    /**
+     * If provided and is positive, returns a 32-bit pseudorandom number from the given range in the transaction record.
+     * If not set or set to zero, will return a 384-bit pseudorandom number in the record.
```
steven-sheehy, line 37:
```suggestion
     * If not set or set to zero, will return a 384-bit pseudorandom data in the record.
```

steven-sheehy (changes requested):

sdk/src/main/java/com/hedera/hashgraph/sdk/RngTransaction.java
```diff
@@ -0,0 +1,91 @@
...
+import com.hedera.hashgraph.sdk.proto.TransactionBody;
+import com.hedera.hashgraph.sdk.proto.TransactionResponse;
+import io.grpc.MethodDescriptor;
+
+import javax.annotation.Nullable;
+
+/**
+ * Random Number Generator Transaction.
+ */
+public final class RngTransaction extends com.hedera.hashgraph.sdk.Transaction<RngTransaction> {
```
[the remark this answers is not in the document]
danielakhterov: The reason we often rename things in the SDK is to provide clarity. In my opinion `RandomGenerateTransaction` does not make sense; randomly generate _what_? I'm not saying `RngTransaction` is written in stone, but _to me_ `RngTransaction` at least conveys this transaction will generate random numbers in some form or another.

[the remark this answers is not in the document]
steven-sheehy: Then you should be more involved in the HIP or protobuf process to effect change there. This is the naming we decided upon and the SDKs should match it to be consistent.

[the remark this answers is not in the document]
rbair23: I brought this up in a previous discussion too, in that I find it difficult to use the SDK when it names things differently. It would be better to have consistency across these APIs. Not a lot of people use the HAPI directly, so in some sense the SDK is the "true" API, but these transaction names are exposed also in mirror nodes and explorers and record files and other places. People might get lost looking for an RngTransaction in the records and wonder why we couldn't have more consistency.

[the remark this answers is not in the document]
rbair23: We are discussing this right now, along with a few other changes to this HIP. We are proposing to use PrngTransaction as the name for it (and updating the method used in the smart contracts too)

danielakhterov (commented):
Changes look good, but we should add a test.

rickyrodmac:
There are 2 shouldSerialize tests in the RngTransactionTest.java file.

PR #1071, opened 2022-08-06
test: overhaul unit tests

### LINKED ISSUE #1016: Add unit tests
### Problem

Identify areas to add unit tests and increase the test coverage.

### Solution

Add units tests.

### Alternatives

_No response_

### LINKED ISSUE #1072: Deprecate proxyAccountId fields in transactions
### Problem

`proxyAccountId` fields in `AccountCreateTransaction` and `ContractCreateTransaction` are deprecated in the protobufs.

### Solution

The `setProxyAccountId()` and `getProxyAccountId()` methods in `AccountCreateTransaction` and `ContractCreateTransaction` should be deprecated to match.

### Alternatives

_No response_

Sean-Tedrow-LB:

**Related issue(s)**:

Fixes #1016
Fixes #1072


sdk/src/test/java/com/hedera/hashgraph/sdk/AccountBalanceQueryTest.snap
```diff
@@ -1,8 +1,8 @@
 com.hedera.hashgraph.sdk.AccountBalanceQueryTest.shouldSerializeWithAccountId=[
-  "\ncryptoget_account_balance {\n  account_i_d {\n    account_num: 5005\n    realm_num: 0\n    shard_num: 0\n  }\n  header {\n  }\n}"
+  "# com.hedera.hashgraph.sdk.proto.Query@b01fa486\ncryptoget_account_balance {\n  account_i_d {\n    account_num: 5005\n    realm_num: 0\n    shard_num: 0\n  }\n  header {\n  }\n}"
```
danielakhterov, line 2: I'm pretty sure this test will be inconsistent because the object pointer is being serialized.

Sean-Tedrow-LB: Oh, that's right.  I need to implement it the same way `Tx.toString()` is implemented, and I do recall that it does something special with that.

danielakhterov (dismissed):

danielakhterov (dismissed):

danielakhterov:
@Sean-Tedrow-LB Changes look great, was about to merge before I looked at SonarCloud's report more carefully, and it does have some valid points. I don't know yet what we should be putting in `ForRemoval` of the deprecation, but we can set `since` to be as of next release which should be 2.17.2? Also, it contains about some of our string comparisons it seems.

danielakhterov (approved):

steven-sheehy:
How can I see the code coverage reports? I don't see a Github comment from codecov.io like in other projects. Is that something you can add?

danielakhterov:
I'm not too familiar with the codecov GitHub comment, but I'm assuming we have to enable their app on this repo.

PR #1091, opened 2022-08-15
fix: thread leak

### LINKED ISSUE #1084: Thread leak in Client
### Description

The `Client` class contains an `Executor` that it creates and manages but does not shutdown in its `close()` method. This causes a thread leak since threads grow but never decrease as clients come and go. This is more noticeable if you create a lot of temporary clients like the mirror node does to do periodic validation of consensus nodes.

### Steps to reproduce

```java
for (int i = 0; i < 10000; ++i) {
  try (Client client = Client.forMainnet()) {
    // Use it
  }
}
```

### Additional context

Example shutdown code from the `ExecutorService` javadocs:
```java
   try {
       pool.shutdown();
       if (!pool.awaitTermination(30, TimeUnit.SECONDS)) {
           pool.shutdownNow();
           if (!pool.awaitTermination(60, TimeUnit.SECONDS)) {
               log.warn("Pool did not terminate");
           }
       }
   } catch (InterruptedException ex) {
       pool.shutdownNow();
       Thread.currentThread().interrupt();
   }
```

### Hedera network

mainnet

### Version

v2.17.1

### Operating system

_No response_

Sean-Tedrow-LB:

**Related issue(s)**:

Fixes #1084

**Notes for reviewer**:

**Checklist**


sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
```diff
@@ -1004,7 +1005,15 @@ public synchronized void close() throws TimeoutException {
         try {
             network.close();
             mirrorNetwork.close();
+            executor.shutdown();
```
steven-sheehy, line 1008: Doesn't fix a similar problem in `close(Duration)`. Should combine the two implementations so that one calls the other with a default timeout.

danielakhterov:
cc @steven-sheehy

sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
```diff
@@ -1004,7 +1005,15 @@ public synchronized void close() throws TimeoutException {
...
+            if (!executor.awaitTermination(30, TimeUnit.SECONDS)) {
+                executor.shutdownNow();
+                if (!executor.awaitTermination(60, TimeUnit.SECONDS)) {
```
steven-sheehy, line 1011: Waiting up until 90 seconds by default is a bit long. After combining the two methods should set a reasonable overall timeout like 30s and decrement time spent between the two `awaitTermination()` methods.

sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
```diff
@@ -1004,7 +1005,15 @@ public synchronized void close() throws TimeoutException {
...
+                    logger.warn("Executor did not terminate");
+                }
+            }
         } catch (InterruptedException e) {
+            executor.shutdownNow();
             throw new RuntimeException(e);
```
steven-sheehy, line 1017: Shouldn't rethrow `InterruptedException`. Call `Thread.currentThread().interrupt()` instead.

steven-sheehy (changes requested):

sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
```diff
@@ -1004,7 +1005,15 @@ public synchronized void close() throws TimeoutException {
         try {
             network.close();
             mirrorNetwork.close();
+            executor.shutdown();
+            if (!executor.awaitTermination(30, TimeUnit.SECONDS)) {
```
danielakhterov, line 1009: Should we be using `closeTimeout` here?

danielakhterov (approved):

steven-sheehy (approved):

sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
```diff
@@ -108,14 +113,7 @@ public final class Client implements AutoCloseable, WithPing, WithPingAll {
...
     static ExecutorService createExecutor() {
-        var threadFactory = new ThreadFactoryBuilder()
-            .setNameFormat("hedera-sdk-%d")
-            .setDaemon(true)
-            .build();
-
-        return Executors.newFixedThreadPool(
-            Runtime.getRuntime().availableProcessors(),
-            threadFactory);
+        return ForkJoinPool.commonPool();
```
Sean-Tedrow-LB, line 116: @limechain

Revert this change, keep the rest of the changes in this PR.

sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
```diff
@@ -1020,11 +1013,19 @@ public synchronized void close() throws TimeoutException {
...
-            mirrorNetwork.close(timeout);
-        } catch (InterruptedException e) {
-            throw new RuntimeException(e);
+        var closeDeadline = Instant.now().plus(timeout);
+        network.beginClose();
+        mirrorNetwork.beginClose();
+
+        var networkError = network.awaitClose(closeDeadline, null);
+        var mirrorNetworkError = mirrorNetwork.awaitClose(closeDeadline, networkError);
+
```
Sean-Tedrow-LB, line 1022: @limechain

Close the executor here

PR #1093, opened 2022-08-16
fix: update network

### LINKED ISSUE #1082: Update Mainnet Address book
### Problem

The mainnet address book used by the SDK is quite old. This can result in transactions failing due to them being sent to old or invalid nodes.

### Solution

The address book should be updated. 

Additionally, I'd suggest adding a step to the release process to always pull the latest address book from the network at release time. That way, whenever the SDK ships, it will have the most up-to-date address book possible.

I've attached the current mainnet address book in both json and protobuf formats. Please let me know if I can assist in any other way.

[node-details-mainnet.102.json.zip](https://github.com/hashgraph/hedera-sdk-java/files/9310740/node-details-mainnet.102.json.zip)
[mainnet.pb.zip](https://github.com/hashgraph/hedera-sdk-java/files/9310732/mainnet.pb.zip)

### Alternatives

_No response_

Sean-Tedrow-LB:

**Related issue(s)**:

Fixes #1082


danielakhterov (approved):

danielakhterov (approved):

sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
```diff
@@ -301,7 +301,7 @@ public static Client fromConfig(Reader json) throws Exception {
                 String mirror = config.mirrorNetwork.getAsString();
                 switch (mirror) {
                     case "mainnet":
-                        client.setMirrorNetwork(Lists.of("hcs.mainnet.mirrornode.hedera.com:5600"));
+                        client.setMirrorNetwork(Lists.of("mainnet-public.mirrornode.hedera.com:5600"));
```
steven-sheehy, line 304: Did you test this? There is no port 5600 on mainnet-public.

danielakhterov: I thought we cleared this up before. However, yea there is port 5600 for `mainnet-public`, but neither is there for `hcs.mainnet`? It seemed reasonable to change this to an endpoint that would work for users?

Sean-Tedrow-LB: Yeah, that is kind of a problem, but I think it's just a reflection of a deeper problem that I think may require a separate PR: our SDK was designed with the assumption that there is a TLS endpoint and a non-TLS endpoint for each node, and that all of the networks and nodes can be switched between the non-TLS and the TLS endpoints using the same procedure, and we defaulted to the non-TLS connection mode for the client.

So it's not clear what to do here.  If the user is connected to the consensus network in non-TLS mode, do we want to connect to the mirror node in TLS mode on mainnet?  That would be a break from previous behavior.  What about testnet and previewnet, where non-TLS endpoints _do_ exist?  When the `Client` is set to non-TLS mode, should it connect to the non-TLS endpoints for the testnet and previewnet mirror nodes to match current behavior, or should it connect to the TLS endpoints to match the behavior of mainnet?

I kind of think we just need to have a separate issue and PR to sort out all of the TLS stuff.  I think that refactoring, and adjustments to the interface, are in order.

As it stands, you're simply required to enable TLS on the `Client` if you want to make mirror node queries.

steven-sheehy: I thought we did clear this up, but apparently I was misunderstood. There is **no** port 5600 on mainnet-public.mirrornode.hedera.com. 

```bash
$ grpcurl -plaintext -d '{"file_id": {"fileNum": 102}, "limit": 0}' mainnet-public.mirrornode.hedera.com:5600 com.hedera.mirror.api.proto.NetworkService/getNodes
Failed to dial target host "mainnet-public.mirrornode.hedera.com:5600": dial tcp 34.102.244.153:5600: connect: connection refused
```

It is better to revert this change until you can refactor your code to support a TLS only mainnet-public. I don't see a problem with connecting to TLS only mainnet-public mirror node regardless of config option. There's no downside to using TLS with mirror nodes since it uses a hostname with a CA signed certificate. Unlike consensus nodes that use IPs and self-signed certificates that people might not want to trust.

PR #1101, opened 2022-08-24
doc: add precompile example

### LINKED ISSUE #1094: Add HIP-514 example
### Problem

HIP-514 introduces new token precompiles that Hedera supports. Add examples using the new precompiles are supported.
HIP: https://hips.hedera.com/hip/hip-514

The following are not yet support from the list:
- getTokenExpiryInfo(address)
- updateTokenExpiryInfo(address, Expiry),
- updateTokenInfo(address, HederaToken), 
- isToken(address token)
- getTokenType(address token)

### Solution

Add example

### Alternatives

_No response_

### LINKED ISSUE #1118: Functions missing from `ContractCreateFlow`
### Problem

The following methods are missing from `ContractCreateFlow`:

- `.[get|set]AutoRenewAccountId()`
- `.[get|set]MaxChunks()`

### Solution

add them

### Alternatives

_No response_

### LINKED ISSUE #1119: Support for handling raw response codes
### Problem

The solidity precompile functions return raw response code integers.

### Solution

Add functions for converting raw response code integers to and from `Status`.

- `Status.[to|from]ResponseCode()`

### Alternatives

_No response_

Sean-Tedrow-LB:

**Description**:

**Related issue(s)**:

Fixes #1094
Fixes  #1118
Fixes #1119

**Notes for reviewer**:

**Checklist**


danielakhterov (approved):

examples/src/main/java/SolidityPrecompileExample.java
```diff
@@ -0,0 +1,126 @@
...
+            // step 2 mints it
+            // step 3 associates Alice with it
+            // step 4 transfers it to Alice.
+            // step 5 approves an allowance of the fungible token with operator as the owner and Alice as the spender [NOT WORKING]
+            // steps 6 - 10 test misc functions on the fungible token (see PrecompileExample.sol for details).
+            // step 11 creates an NFT token with a custom fee, and with the admin and supply set to Alice's key
+            // step 12 mints some NFTs
+            // step 13 associates Alice with the NFT token
+            // step 14 transfers some NFTs to Alice
+            // step 15 approves an NFT allowance with operator as the owner and Alice as the spender [NOT WORKING]
```
nikolovyanko, line 111: It's highly possible that the feature from [PR](https://github.com/hashgraph/hedera-services/pull/3638) for approveNft isn't available in testnet and previewnet yet, because the txn's behavior is as the approveNft function is missing.

examples/src/main/java/SolidityPrecompileExample.java
```diff
@@ -0,0 +1,126 @@
...
+                .addSignerForStep(13, alicePrivateKey)
+                // Alice must sign to burn the token because her key is the supply key
+                .addSignerForStep(16, alicePrivateKey);
+
+            // step 0 tests pseudo random number generator (PRNG)
+            // step 1 creates a fungible token
+            // step 2 mints it
+            // step 3 associates Alice with it
+            // step 4 transfers it to Alice.
+            // step 5 approves an allowance of the fungible token with operator as the owner and Alice as the spender [NOT WORKING]
```
nikolovyanko, line 105: Step 5 function is failing because the created fungible token is associated only with the spender (aliseAccount), but it needs to be also associated with the provider of allowance, which in this case is the calling contract.

examples/src/main/resources/precompile-example/PrecompileExample.sol
```diff
@@ -0,0 +1,237 @@
...
+            100, // amount (applicable to fungible tokens only)
+            new bytes[](0) // metadatas (applicable to NFT tokens only)
+        );
+
+        require(newTotalSupply == 100 + 100);
+    }
+
+    function step3() external returns (int responseCode) {
+        require(msg.sender == owner);
+
```
nikolovyanko, line 78: You can associate the token to the contract address directly here in the contract as a quick fix suggestion to the issue with step5, e.g.  responseCode = associateToken(address(this), fungibleToken) .

examples/src/main/resources/precompile-example/IHederaTokenService.sol
```diff
@@ -0,0 +1,763 @@
...
+    /// Allow or reaffirm the approved address to transfer an NFT the approved address does not own.
+    /// Only Applicable to NFT Tokens
+    /// @param token The Hedera NFT token address to approve
+    /// @param approved The new approved NFT controller.  To revoke approvals pass in the zero address.
+    /// @param serialNumber The NFT serial number  to approve
+    /// @return responseCode The response code for the status of the request. SUCCESS is 22.
+    function approveNFT(
+        address token,
+        address approved,
+        int64 serialNumber
```
stoqnkpL, line 531:
```suggestion
        uint256 serialNumber
```
like it is in [IHederaTokenService.sol](https://github.com/hashgraph/hedera-smart-contracts/blob/8546133775642df03ab63d7d7141ef560c41da29/contracts/hts-precompile/IHederaTokenService.sol#L526) in the smart-contracts repo (the last arg has a different type and this changes the function signature).

examples/src/main/java/SolidityPrecompileExample.java
Sean-Tedrow-LB: I would have thought the provider of the allowance would be whoever's ID is in the transaction ID of the `ContractExecuteTransaction`.  Does this mean that the owner of an allowance can only ever be the contract itself?  Is there any way for a contract function to approve an allowance with a different owner?

danielakhterov (approved):

examples/src/main/java/SolidityPrecompileExample.java
stoqnkpL: The allowance provider is the `sender` in the message frame when we reach the precompiled contract. 
For example, if when have this setup:
User A --calls-> Contract C --calls-> Precompile, the `sender` would be Contract C.
> Does this mean that the owner of an allowance can only ever be the contract itself?

No, we could also have this setup (notice the delegatecall):
User A --calls-> Contract C --delegatecalls-> Precompile, then the `sender` would be User A.
We have such a contract and test in the services repo in [ApproveAllowanceSuite](https://github.com/hashgraph/hedera-services/blob/bb302309e61eb0b0fc11de967266433951b6bb31/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/ApproveAllowanceSuite.java#L209) which uses the [HtsApproveAllowance.sol](https://github.com/hashgraph/hedera-services/blob/bb302309e61eb0b0fc11de967266433951b6bb31/test-clients/src/main/resource/contract/contracts/HtsApproveAllowance/HtsApproveAllowance.sol#L18) contract.

PR #1225, opened 2022-11-02
Fix: Client close timeout while subscribed to mirror node topic query - #1182

### LINKED ISSUE #1182: Client.close() times out if subscribing to mirror gRPC API
### Description

Attempting to close a `Client` in the middle of a mirror node gRPC API streaming query will block for `closeTimeout` then throw a timeout exception. Since you can't have a subscription without the `Channel` that the `Client` manages for it, it's expected that the SDK would register any subscriptions and close them gracefully on `close()`. Requiring the user to manually unsubscribe from the gRPC query is awkward and error-prone since it can only unsubscribe during the happy path or if close is called manually. In the below, it can't just unsubscribe once in a finally block.

### Steps to reproduce

```java
try (Client client = Client.forMainnet()) {
  ...
  new TopicMessageQuery().setTopicId(topicId).subscribe(client, r -> {});
}
```

### Additional context

```
2022-10-15T19:12:06.518-0600 ERROR ForkJoinPool-2-worker-9 i.g.i.ManagedChannelImpl [Channel<13>: (previewnet.mirrornode.hedera.com:443)] Uncaught exception in the SynchronizationContext. Panic! io.grpc.StatusRuntimeException: UNKNOWN: Uncaught exception in the SynchronizationContext. Re-thrown.
	at io.grpc.Status.asRuntimeException(Status.java:530)
	at io.grpc.internal.RetriableStream$1.uncaughtException(RetriableStream.java:75)
	at io.grpc.SynchronizationContext.drain(SynchronizationContext.java:97)
	at io.grpc.SynchronizationContext.execute(SynchronizationContext.java:127)
2022-10-15T19:12:06.522-0600 ERROR ForkJoinPool-2-worker-9 c.h.m.t.e.a.c.SDKClient startup probe:  java.util.concurrent.TimeoutException: Failed to properly shutdown all channels
	at com.hedera.hashgraph.sdk.BaseNetwork.awaitClose(BaseNetwork.java:605)
	at com.hedera.hashgraph.sdk.Client.close(Client.java:1305)
	at com.hedera.hashgraph.sdk.Client.close(Client.java:1271)
```

### Hedera network

mainnet

### Version

v2.18.0

### Operating system

_No response_

nikola-genov:
**Description**:

Fix Client close timeout while subscribed to mirror node topic query

This PR modifies
* Client to support tracking of open topic queries to mirror node
* Client to automatically unsubscribe those topics on close
* Related integration test added

Fixes #1182 

**Notes for reviewer**:
The MirrorNodeCloseClientIntegrationTest was run 1000 times in a row to make sure no race conditions arise on SubscriptionHandle.unsubscribe() (which is asynchronous). This does not give us 100% guarantee that such issues will never happen on production. In the future the entire Client.close() logic may be revised if needed.

**Checklist**


sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
```diff
@@ -106,6 +101,8 @@ public final class Client implements AutoCloseable {
     @Nullable
     private CompletableFuture<Void> networkUpdateFuture;
 
+    private Set<SubscriptionHandle> subscriptions = new HashSet<>();
```
steven-sheehy, line 104: This is not thread safe. Also, there's not much reason it should be a set since SubscriptionHandle does not implement equals/hashCode and its addition/removal is handled internally. 
```suggestion
    private Collection<SubscriptionHandle> subscriptions = Collections.synchronizedList(new ArrayList<>());
```

sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
```diff
@@ -404,6 +401,25 @@ private void cancelScheduledNetworkUpdate() {
         }
     }
 
+    private void cancelAllSubscriptions() {
+        ArrayList<SubscriptionHandle> handles = new ArrayList<>(subscriptions);
+        for (SubscriptionHandle handle : handles) {
+            handle.unsubscribe();
+        }
```
steven-sheehy, line 408: No need for a copy. This is handled within a synchronized close.

```suggestion
    private void cancelAllSubscriptions() {
        subscriptions.forEach(SubscriptionHandle::unsubscribe);
```

sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
```diff
@@ -404,6 +401,25 @@ private void cancelScheduledNetworkUpdate() {
...
+    }
+
+    void trackSubscription(SubscriptionHandle subscriptionHandle) {
+        if (!subscriptions.contains(subscriptionHandle)) {
+            subscriptions.add(subscriptionHandle);
+        }
```
steven-sheehy, line 414: Redundant to check contains before adding with a set. But since switching to a list can just add.

sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
```diff
@@ -404,6 +401,25 @@ private void cancelScheduledNetworkUpdate() {
...
+    }
+
+    void untrackSubscription(SubscriptionHandle subscriptionHandle) {
+        if (subscriptions.contains(subscriptionHandle)) {
+            subscriptions.remove(subscriptionHandle);
+        }
```
steven-sheehy, line 420:
```suggestion
        subscriptions.remove(subscriptionHandle);
```

sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
```diff
@@ -404,6 +401,25 @@ private void cancelScheduledNetworkUpdate() {
         }
     }
 
+    private void cancelAllSubscriptions() {
+        ArrayList<SubscriptionHandle> handles = new ArrayList<>(subscriptions);
+        for (SubscriptionHandle handle : handles) {
+            handle.unsubscribe();
```
steven-sheehy, line 407: Should enhance `SubscriptionHandle::unsubscribe` to set `onUnsubscribe = null` after `run()` so it only runs once even if invoked by user and Client

steven-sheehy (changes requested):

sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
nikola-genov: The copy is needed to avoid ConcurrentModificationException after the first item is removed from the set/collection (that happens in onUnsubscribe - client.untrackSubscription(subscriptionHandle)).

sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
nikola-genov: ArrayList would allow for duplicate records of the same instance to be added to the collection. The default Object.hashCode is all we need to avoid duplicates in a Set. I suggest that we settle for:

private Set<SubscriptionHandle> subscriptions = Collections.synchronizedSet(new HashSet<>());

Or is there any reason why duplicates should be allowed?

steven-sheehy: I think it's fine to use set, just didn't think it was necessary since there shouldn't be duplicates because the object is always internally added/removed and not by the user. If we want to use a concurrent set, then I think we should use `ConcurrentHashMap.newKeySet()`. It's way more performant and has concurrent iterator support so we don't need to copy the collection to iterate. Only caveat is it's available since Java 8 and I'm not sure what the baseline Java version the SDK supports.

CHANGELOG.md
```diff
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+ - `MirrorNodeCloseClientIntegrationTest`
```
steven-sheehy, line 11: I don't think we should document test changes in the changelog. It's assumed fixes would have a corresponding test.

sdk/src/integrationTest/java/MirrorNodeCloseClientIntegrationTest.java
```diff
@@ -0,0 +1,51 @@
...
+            .setTopicMemo("[e2e::TopicCreateTransaction]")
+            .execute(testEnv.client);
+
+        var topicId = Objects.requireNonNull(response.getReceipt(testEnv.client).topicId);
+
+        new TopicInfoQuery()
+            .setTopicId(topicId)
+            .execute(testEnv.client);
+
+        Thread.sleep(3000);
```
steven-sheehy, line 28: Thread sleeps are always a code smell and should be removed. Internally, the TopicInfoQuery and TopicMessageQuery should already retry the request. If need be, that retry mechanism could be adjusted instead of sleeping.

sdk/src/integrationTest/java/MirrorNodeCloseClientIntegrationTest.java
```diff
@@ -0,0 +1,51 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.Objects;
+
+import static org.assertj.core.api.Assertions.*;
+
+public class MirrorNodeCloseClientIntegrationTest {
+    @Test
+    @DisplayName("Can close the Client while subscribed to mirror node topic queries")
+    void canCloseClient() throws Exception {
+        var testEnv = new IntegrationTestEnv(1);
```
steven-sheehy, line 15: This test should not be an integration test. Tests that hit a remote system are brittle and should be avoided. There's already a stub-based test for `TopicMessageQuery` that should be enhanced for this scenario.

sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
steven-sheehy: Think we should use `ConcurrentHashMap.newKeySet()` then we can avoid this copy.

steven-sheehy (changes requested):

steven-sheehy (approved):
LGTM

dikel (approved):
LGTM

PR #1232, opened 2022-11-10
Fix: Executable.execute to set gRPC deadline #1226

### LINKED ISSUE #1226: Execute timeout should use gRPC deadline
### Problem

Follow up from https://github.com/hashgraph/hedera-sdk-java/issues/1175.

The timeout in `Executable.execute(Client client, Duration timeout)` is only used between retry attempts and not the underlying gRPC request. For that there is a `grpcDeadline` property which is unset by default. This means that if I set a timeout for 2s but the grpc request takes 30 seconds then the client will stop after 2s but the underlying grpc request will continue running in the background.

### Solution

The SDK should set this by default if a timeout is provided. The timeout should be the overall timeout for all attempts, but on each attempt it should calculate the timeout remaining and pass that as the `grpcDeadline` in the gRPC CallOptions. Perhaps the deadline could also have max duration like 10 seconds.

### Alternatives

_No response_

nikola-genov:
Signed-off-by: Nikola Genov <nikola.genov@limechain.tech>

**Description**:

* `Executable.execute(Client client, Duration timeout)` now sets gRPC deadline to the underlying gRPC request
* new executableShouldUseGrpcDeadline() added to ExecutableTest.java

* **Related issue(s)**:

Fixes #1226 

**Checklist**

### DIFF
--- a/sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
+++ b/sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
@@ -44,6 +44,7 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.util.*;
 import java.util.concurrent.*;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Supplier;
 
 import static com.hedera.hashgraph.sdk.BaseNodeAddress.PORT_NODE_PLAIN;
@@ -84,7 +85,7 @@ public final class Client implements AutoCloseable {
 
     private Duration requestTimeout = DEFAULT_REQUEST_TIMEOUT;
     private Duration closeTimeout = DEFAULT_CLOSE_TIMEOUT;
-    private volatile Duration grpcDeadline = DEFAULT_GRPC_DEADLINE;
+    private AtomicReference<Duration> grpcDeadline = new AtomicReference(DEFAULT_GRPC_DEADLINE);
 
     private int maxAttempts = DEFAULT_MAX_ATTEMPTS;
 
@@ -1252,7 +1253,7 @@ public final class Client implements AutoCloseable {
         justification = "A Duration can't actually be mutated"
     )
     public Duration getGrpcDeadline() {
-        return grpcDeadline;
+        return grpcDeadline.get();
     }
 
     /**
@@ -1266,7 +1267,7 @@ public final class Client implements AutoCloseable {
         justification = "A Duration can't actually be mutated"
     )
     public Client setGrpcDeadline(Duration grpcDeadline) {
-        this.grpcDeadline = Objects.requireNonNull(grpcDeadline);
+        this.grpcDeadline.set(Objects.requireNonNull(grpcDeadline));
         return this;
     }
 
--- a/sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
+++ b/sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
@@ -85,7 +85,7 @@ public final class Client implements AutoCloseable {
 
     private Duration requestTimeout = DEFAULT_REQUEST_TIMEOUT;
     private Duration closeTimeout = DEFAULT_CLOSE_TIMEOUT;
-    private AtomicReference<Duration> grpcDeadline = new AtomicReference(DEFAULT_GRPC_DEADLINE);
+    private final AtomicReference<Duration> grpcDeadline = new AtomicReference(DEFAULT_GRPC_DEADLINE);
 
     private int maxAttempts = DEFAULT_MAX_ATTEMPTS;
 
--- a/sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
+++ b/sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
@@ -60,6 +60,7 @@ public final class Client implements AutoCloseable {
     static final Duration DEFAULT_MIN_NODE_BACKOFF = Duration.ofSeconds(8L);
     static final Duration DEFAULT_CLOSE_TIMEOUT = Duration.ofSeconds(30L);
     static final Duration DEFAULT_REQUEST_TIMEOUT = Duration.ofMinutes(2L);
+    static final Duration DEFAULT_GRPC_DEADLINE = Duration.ofSeconds(10L);
     static final Duration DEFAULT_NETWORK_UPDATE_PERIOD = Duration.ofHours(24);
     // Initial delay of 10 seconds before we update the network for the first time,
     // so that this doesn't happen in unit tests.
@@ -83,6 +84,7 @@ public final class Client implements AutoCloseable {
 
     private Duration requestTimeout = DEFAULT_REQUEST_TIMEOUT;
     private Duration closeTimeout = DEFAULT_CLOSE_TIMEOUT;
+    private Duration grpcDeadline = DEFAULT_GRPC_DEADLINE;
 
     private int maxAttempts = DEFAULT_MAX_ATTEMPTS;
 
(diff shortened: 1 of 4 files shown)


sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
```diff
@@ -353,10 +353,13 @@ public O execute(Client client, Duration timeout) throws TimeoutException, Prech
                 throw new MaxAttemptsExceededException(lastException);
             }
 
-            if (Instant.now().isAfter(timeoutTime)) {
+            Duration deadline = Duration.between(Instant.now(), timeoutTime);
+            if (deadline.isNegative() || deadline.isZero()) {
                 throw new TimeoutException();
             }
 
+            this.setGrpcDeadline(deadline);
```
steven-sheehy, line 361: There's a `Executable.setGrpcDeadline()` that should still be respected. Here's its value is being overwritten so users can't customize it.

sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
```diff
@@ -353,10 +353,13 @@ public O execute(Client client, Duration timeout) throws TimeoutException, Prech
                 throw new MaxAttemptsExceededException(lastException);
             }
 
-            if (Instant.now().isAfter(timeoutTime)) {
+            Duration deadline = Duration.between(Instant.now(), timeoutTime);
```
steven-sheehy, line 356: I think the deadline should be lower than the timeout by default to allow room for retries to other nodes. Here, if timeout is `30s` then the deadline consumes all `30s`.

Instead, we should add a default `grpcDeadline = Duration.ofSeconds(10L)` to `Client`. Then each `Executable.grpcDeadline` will default to `Client.grpcDeadline`. Then here we should continue to check `Instant.now().isAfter(timeoutTime)` and also calculate a min of `executable.grpcDeadline` and the timeout remaining.

sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
steven-sheehy, line 361: I don't think we should modify the `Executable` properties set by the user. Instead, keep it as the user provided value and calculate the min deadline with time remaining and pass to `GrpcRequest`. Otherwise if you modify the original you can't calculate the min against timeout on the next retry.

steven-sheehy (changes requested):

sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
nikola-genov: Yes, the remaining timeout was already calculated, I implemented the rest of the suggestions.

sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
```diff
@@ -1240,6 +1242,34 @@ public Client setCloseTimeout(Duration closeTimeout) {
...
+    /**
+     * Maximum amount of time a gRPC request can run
+     *
+     * @return                          the gRPC deadline value
+     */
+    @SuppressFBWarnings(
+        value = "EI_EXPOSE_REP",
+        justification = "A Duration can't actually be mutated"
+    )
+    public synchronized Duration getGrpcDeadline() {
```
steven-sheehy, line 1254: Remove synchronized on get as it will be a performance bottleneck. Can also remove it on set if you use volatile or AtomicReference for the field

sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
```diff
@@ -573,7 +578,7 @@ private void executeAsyncInternal(
             return;
         }
 
-        GrpcRequest grpcRequest = new GrpcRequest(client.network, attempt);
+        GrpcRequest grpcRequest = new GrpcRequest(client.network, attempt, null);
```
steven-sheehy, line 581: Why are we not setting deadline for async? We definitely should set deadline for all flows.

sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
```diff
@@ -721,11 +729,15 @@ class GrpcRequest {
...
-            if (Executable.this.grpcDeadline != null) {
-                return options.withDeadlineAfter(Executable.this.grpcDeadline.toMillis(), TimeUnit.MILLISECONDS);
-            } else {
+            if (this.grpcDeadline == null && Executable.this.grpcDeadline == null ) {
                 return options;
             }
+
+            long deadline = Math.min(
+                this.grpcDeadline == null ? Long.MAX_VALUE : this.grpcDeadline.toMillis(),
+                Executable.this.grpcDeadline == null ? Long.MAX_VALUE : Executable.this.grpcDeadline.toMillis());
```
steven-sheehy, line 738: How can either of these be null? Client has a default and we don't allow null. Executable inherits from client if not set and we don't allow null. GrpcRequest inherits from Executable. Should remove all nullable logic in GrpcRequest.

sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
```diff
@@ -353,11 +357,12 @@ public O execute(Client client, Duration timeout) throws TimeoutException, Prech
                 throw new MaxAttemptsExceededException(lastException);
             }
 
-            if (Instant.now().isAfter(timeoutTime)) {
+            Duration deadline = Duration.between(Instant.now(), timeoutTime);
```
steven-sheehy, line 360: It's a bit misleading to call this deadline, since the actual deadline will be under 10s. This is more a timeout for all requests whereas deadline is per request. So recommend renaming to `timeout`.

steven-sheehy (changes requested):

sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
nikola-genov: Makes sense. However getRequestTimeout() is also synchronized, is that on purpose?

sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
nikola-genov: GrpcRequest constructor currently allows null and the null check logic was already in place so I kept it. But yes, we can ensure no null values are permitted and clean-up the logic.

sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
nikola-genov: I renamed it to 'currentTimeout' to differentiate from the method parameter which is already called timeout.

sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
nikola-genov: I switched to AtomicRef for the grpcDeadline. Probably in a separate work item we'll need to refactor the rest of the Client's members for consistency...

sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
```diff
@@ -83,6 +85,7 @@ public final class Client implements AutoCloseable {
 
     private Duration requestTimeout = DEFAULT_REQUEST_TIMEOUT;
     private Duration closeTimeout = DEFAULT_CLOSE_TIMEOUT;
+    private AtomicReference<Duration> grpcDeadline = new AtomicReference(DEFAULT_GRPC_DEADLINE);
```
steven-sheehy, line 88: nit: Should be final

steven-sheehy (approved):
LGTM

steven-sheehy (approved):

dikel (approved):
LGTM

PR #1253, opened 2022-11-22
BIP39 - mnemonic passphrase normalize to NFKD

### LINKED ISSUE #1240: Bip39 passphrase must be normalized
Current implementation allows any unicode string as a passphrase, but the [BIP39 spec](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki#generating-the-mnemonic) states that it must be normalized to NFKD 

Open for discussion: Should there be any additional limitations on the allowed passphrases? The above change will bring the SDK in line with the BIP39 spec, but it seems very common for wallets to limit the user to specifically ASCII characters, with a length cap of 100 characters. Should we follow suit, to avoid the situation where a user cannot recover a key on specific wallets?

nikola-genov:
Signed-off-by: Nikola Genov <nikola.genov@limechain.tech>

**Description**:

This PR modifies Mnemonic seed generation when a passphrase is used. In order to support BIP-39 the passphrase is normalized to NFKD.

**Related issue(s)**:

Fixes #1240 

**Notes for reviewer**:
Should we normalize the whole mnemonic and not just the passphrase?

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  - Transaction sometimes being reported as duplicate when submitting large number of transactions
  - `RejectedExecutionException` under heavy load
  - `nodes` not clearing when reusing transaction
+ - BIP-39 - unicode mnemonic passphrases are normalized to NFKD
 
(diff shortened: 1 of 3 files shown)


dikel (approved):
LGTM

PR #1309, opened 2023-01-24
Refactor key derivation

### LINKED ISSUE #1263: `fromMnemonic` returns a private key which shouldn't be used
This isn't a bug, but rather a way that an SDK user can easily misuse the `PrivateKey::fromMnemonic` function.

By default, calling `fromMnemonic` returns the key at derivation path `m/44'/3030'/0'/0'` (`A` in the figure below). If a programmer isn't familiar with derivation paths, they will probably just use the resulting private key without a further thought. This behavior seems to be enforced by the [online documentation](https://docs.hedera.com/guides/docs/sdks/keys/recover-keys-from-a-mnemonic-phrase), which shows simple usage of `fromMnemonic`

```
//Use the mnemonic to recover the private key
PrivateKey privateKey = PrivateKey.fromMnemonic(mnemonic);
PublicKey publicKey = privateKey.publicKey();

//v2.0.0
```

However, the private key created from `fromMnemonic` should *not* be used directly - the programmer ought to go one layer deeper in the derivation path, to `m/44'/3030'/0'/0'/0'`, `m/44'/3030'/0'/0'/1'`, etc.. (`B` in the figure below). This requirement *is* documented, but only briefly, in the [function documentation](https://github.com/hashgraph/hedera-sdk-java/blob/develop/sdk/src/main/java/com/hedera/hashgraph/sdk/PrivateKey.java)

```
     * @return the recovered key; use {@link #derive(int)} to get a key for an account index (0
     * for default account)
```

Screenshot 2022-11-29 at 10 49 44 AM

My understanding is that receiving to `A` is out of line with expected BIP44 behavior, and doing so might result in different wallets not finding existing funds. It seems worth it to me to make it harder for a wallet developer to mistakenly receive here.

### LINKED ISSUE #1277: Provide test vectors described in SLIP10, BIP32, BIP39
### Description

All test vectors described in [SLIP10](https://github.com/satoshilabs/slips/blob/master/slip-0010.md#test-vectors), [BIP32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki), and [BIP39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki#generating-the-mnemonic) should be checked for all key types.

### Steps to reproduce

Provide tests.

### Additional context

_No response_

### Hedera network

mainnet, testnet, previewnet

### Version

Current Version

### Operating system

None

dikel:
**Description**:
This PR refactors the derivation of private keys:
- Adds `toStandard[Ed25519|ECDSAsecp256k1]PrivateKey()` which use the correct derivation path
- Deprecates `Mnemonic.toPrivateKey()` and `PrivateKey.fromMnemonic()`
- Adds support for ECDSA secp256k1 derivation

**Related issue(s)**:

Fixes #1277, Fixes #1263 

**Checklist**


sdk/src/main/java/com/hedera/hashgraph/sdk/PrivateKeyECDSA.java
```diff
@@ -118,12 +124,72 @@ public PrivateKey legacyDerive(long index) {
...
     @Override
     public boolean isDerivable() {
-        return false;
+        return this.chainCode != null;
     }
 
     @Override
     public PrivateKey derive(int index) {
-        throw new IllegalStateException("ECDSA secp256k1 keys do not currently support derivation");
+        if (this.chainCode == null) {
```
nikola-genov, line 132: I think it's better to use isDerivable() here.

nikola-genov (approved):
Generally looks good to me. One minor improvement could be to use the isDerivable() instead of explicit 'chainCode == null' check.

nikola-genov (approved):

sdk/src/main/java/com/hedera/hashgraph/sdk/Utils/Bip32Utils.java
```diff
@@ -0,0 +1,23 @@
...
+
+public class Bip32Utils {
+    /**
+     * Harden the index
+     *
+     * @param index         the derivation index
+     * @return              the hardened index
+     */
+    public static int toHardenedIndex(int index) {
+        return index | 0x80000000;
```
nikola-genov, line 11: This can be a constant

nikola-genov (approved):
Looks good. Minor improvement would be to extract 0x80000000 as a constant.

sdk/src/main/java/com/hedera/hashgraph/sdk/PrivateKey.java
```diff
@@ -75,6 +72,27 @@ public static PrivateKey generateECDSA() {
...
+     *
+     * @param seed  the seed
+     * @return      the ECDSA private key
+     */
+    public static PrivateKey fromSeedECDSAsecp256k1(byte[] seed) {
+        return PrivateKeyECDSA.fromSeed(seed);
+    }
+
+    /**
+     * @deprecated use {@link Mnemonic#toStandardEd25519PrivateKey(String, int)} ()} or {@link Mnemonic#toStandardECDSAsecp256k1PrivateKey(String, int)} (String, int)} instead
```
alittley, line 95: I think there needs to be a description somewhere about why this was deprecated. At least, the doc should mention that this function uses the incomplete derivation path

sdk/src/test/java/com/hedera/hashgraph/sdk/MnemonicTest.java
```diff
@@ -19,6 +19,8 @@
  */
 package com.hedera.hashgraph.sdk;
 
+import com.hedera.hashgraph.sdk.utils.Bip32Utils;
+import org.bouncycastle.util.encoders.Hex;
```
alittley, line 23: Since this is the first implementation of the new standard, I think it would make sense to include a new set of standard "hedera" test vectors. The list I had come up with is detailed in the `Tests` section of [the standards issue](https://github.com/hashgraph/hedera-sdk-reference/issues/73#issue-1485459503), bullet #2

In addition to defining new test vectors, this would also involve renaming test variables `MNEMONIC3_STRING`, `MNEMONIC_LEGACY_STRING`, `MNEMONIC_STRING`, and `MNEMONIC_PRIVATE_KEY`, so that it is clear what these actually are.

alittley (approved):
Overall LGTM. FYI - I didn't have a chance to verify the mathy parts of this PR (though of course the unit tests testify to correct math). Other reviewers more involved in this SDK should verify separately that the logic makes sense

sdk/src/main/java/com/hedera/hashgraph/sdk/PrivateKey.java
```diff
@@ -75,6 +72,27 @@ public static PrivateKey generateECDSA() {
     }
 
     /**
+     * Extract the ED25519 private key from a seed.
+     *
+     * @param seed  the seed
+     * @return      the ED25519 private key
+     */
+    public static PrivateKey fromSeedED25519(byte[] seed) {
```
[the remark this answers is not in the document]
dikel: I think it's better to leave them to be consistent with the other methods `generateED25519` calls `PrivateKeyED25519.generateInternal` and `generateECDSA` calls `PrivateKeyECDSA.generateInternal`

sdk/src/test/java/com/hedera/hashgraph/sdk/MnemonicTest.java
dikel: We've already added the test vectors (`toStandardED25519PrivateKey` and `toStandardECDSAsecp256k1PrivateKey` methods in `MnemonicTest`

sdk/src/main/java/com/hedera/hashgraph/sdk/Mnemonic.java
```diff
@@ -554,4 +542,48 @@ private byte[] wordsToLegacyEntropy2() throws BadMnemonicException {
 
         return entropy;
     }
+
```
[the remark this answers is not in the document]
alittley: I've fine with that logic, but this really needs to be folded into the [standard proposal](https://github.com/hashgraph/hedera-sdk-reference/issues/73).

For example, I know the C++ SDK effort has already implemented the proposal as it _currently_ exists, with default parameters.

sdk/src/main/java/com/hedera/hashgraph/sdk/PrivateKey.java
```diff
@@ -78,7 +81,14 @@ public static PrivateKey generateECDSA() {
      * @return      the ED25519 private key
      */
     public static PrivateKey fromSeedED25519(byte[] seed) {
-        return PrivateKeyED25519.fromSeed(seed);
+        var hmacSha512 = new HMac(new SHA512Digest());
```
alittley: Putting code particular to a key algorithm into the base class doesn't seem right to me. I'll leave path forward up to you since I'm not involved in the architecture, but IMO this needs a refactor

sdk/src/test/java/com/hedera/hashgraph/sdk/MnemonicTest.java
alittley: It's important that these particular test vectors are tracked somewhere so that all SDKs use the same ones. That will ensure that the `Standard` mnemonic->key functions are consistent across SDKs

PR #1332, opened 2023-02-17
Fix integration tests

### LINKED ISSUE #1324: Fix failing integration tests
### Problem

Some integration tests are failing. 

### Solution

Fix failling tests

### Alternatives

_No response_

nikola-genov:
**Description**:

This PR modifies the integration tests to run on local node.

**Related issue(s)**:

Fixes #1324 

**Checklist**


.github/workflows/build.yml
```diff
@@ -66,14 +67,49 @@ jobs:
             ${{ runner.os }}-gradle-
       - name: Build
```
ochikov, line 68: Let's name it PR Checks

.github/workflows/build.yml
```diff
@@ -66,14 +67,49 @@ jobs:
...
         run: ./gradlew build
-      - name: Generate Jacoco report
-        run: ./gradlew jacocoTestReport
+
+  test:
```
ochikov, line 71: Let's name it: Integration Tests

ochikov (changes requested):

.github/workflows/build.yml
```diff
@@ -5,6 +5,7 @@ on:
     branches:
       - main
       - develop
+      - fix/1324-integration-tests
```
dikel, line 8: Can we remove this?

.github/workflows/build.yml
dikel: I think "test" is better because we are running both the integration and the unit tests

.github/workflows/build.yml
```diff
@@ -66,14 +67,49 @@ jobs:
...
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-
+
+      - name: Start the local node
+        run: npx @hashgraph/hedera-local start -d --network local
+
+      - name: Run Integration Tests
```
dikel, line 96: Like the above comment, I think this should be named "Generate Jacoco report" or "Run Tests"

.github/workflows/build.yml
ochikov: Niki is testing with this branch. We can remove it yes.

ochikov (changes requested):

.github/workflows/build.yml
nikola-genov: removed

.github/workflows/build.yml
nikola-genov: Note: we discussed this and decided to leave it unchanged

.github/workflows/build.yml
nikola-genov: Yes, this is for both the integration and the unit tests

.github/workflows/build.yml
nikola-genov: Agree. Renamed to 'Run Tests'

dikel (approved):
LGTM

PR #1345, opened 2023-02-27
Fix/1324 integration tests flow

nikola-genov:
**Description**:
Test PR - do not merge. 

TODO - delete when no longer necessary... 


.github/workflows/build.yml
```diff
@@ -6,6 +6,7 @@ on:
       - main
       - develop
       - release/*
+      - fix/1324-integration-tests-flow
```
dikel, line 9: Can we remove this?

sdk/src/test/resources/client-config-ci.json
```diff
@@ -0,0 +1,12 @@
+{
```
dikel, line 1: Do we need this file?

PR #1405, opened 2023-05-26
Remove sdk-jdk7 artifact

### LINKED ISSUE #1402: Deprecate the `sdk-jdk7` artifact
### Problem

Currently we have an artifact `sdk-jdk7` which is used by Android apps. This was needed because the old Android versions didn't support Java 8. We can deprecate it if we decide we only need to support devices with API  version >= 26 (which is more than 90% of the devices)

### Solution

- delete `jdk7.gradle`
- update the publish task in `Taskfile.yml`
- delete the unnecessary dependencies

### Alternatives

_No response_

dikel:
**Description**:

Removes the sdk-jdk7 artifact.
Additional changes:
- removed `ThreeTen` dependency
- removed `Jabel` dependency
- removed `future-converter-java8-guava` dependency
- changed the minimum supported Android version to 8
- changed the required Java version to 17+

**Related issue(s)**:

Fixes #1402 

**Notes for reviewer**:

**Checklist**


dikel:
Hey @steven-sheehy, we bumped the minimal Android version to 8 after a [twitter poll](https://twitter.com/hedera/status/1661757375269978112?cxt=HHwWgIDS_ejh348uAAAA) and removed the `sdk-jdk7` artifact because it was mainly needed for older Android versions.

example-android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
+#Fri May 19 11:15:00 EEST 2023
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.5.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.0-bin.zip
```
steven-sheehy, line 4: Might as well use 8.1.1 if you're upgrading.
```suggestion
distributionUrl=https\://services.gradle.org/distributions/gradle-8.1.1-bin.zip
```

steven-sheehy, line 4: You need to run `./gradlew wrapper --gradle-version 8.1.1` to upgrade. Just bumping `gradle-wrapper.properties` does not update the `gradlew` scripts or jar. Those need to be checked in as well.

examples/build.gradle
```diff
@@ -1,17 +1,15 @@
...
     implementation "io.github.cdimascio:java-dotenv:5.3.1"
     implementation "com.google.errorprone:error_prone_core:2.19.1"
+    implementation project(path: ':sdk')
 }
 
 java {
-    sourceCompatibility = org.gradle.api.JavaVersion.VERSION_1_8
-    targetCompatibility = org.gradle.api.JavaVersion.VERSION_1_8
+    sourceCompatibility = 17
+    targetCompatibility = 17
```
steven-sheehy, line 12: Should only need to set these in root `build.gradle` and remove from all sub projects.

example-android/gradle/wrapper/gradle-wrapper.properties
steven-sheehy: Actually, why is there a separate gradle wrapper in a sub-folder instead of using the root gradle wrapper? And why does its version now differ?

dikel: Thank you. Fixed

dikel: The wrapper in the example-android folder is because of Android Studio. Usually you only import the sub-folder so it needs a separate gradle wrapper

steven-sheehy (approved):

PR #1474, opened 2023-07-10
Add more unit tests for `TokenInfo`

### LINKED ISSUE #1469: More Unit tests for `TokenInfo`
### Problem

`TokenInfoTest ` is missing some Unit tests.

### Solution

To add the following Unit tests in `TokenInfoTest.java`:

- `FromProtobuf`
- `FromBytes`
- `ToProtobuf`
- `ToBytes`

For reference see [TokenInfoUnitTests.cc](https://github.com/hashgraph/hedera-sdk-cpp/blob/main/sdk/tests/unit/TokenInfoUnitTests.cc) from the [C++ SDK](https://github.com/hashgraph/hedera-sdk-cpp).

### Alternatives

N/A

thenswan:
**Description**:
- added more unit tests for `TokenInfo`;
- reformatted and rearranged code;
- regenerated test `.snap` file.

**Related issue(s)**:
Fixes #1469 

**Notes for reviewer**:
When using `PrivateKey.fromSeedECDSAsecp256k1(byte[] seed)` method to initialize `PrivateKey`, different public keys are generated each time tests are run on different machines. This can cause `io.github.jsonSnapshot.SnapshotMatcher` to fail tests. To avoid this issue, `PrivateKey` should be initialized using `fromString()` method to get the same key each time. Hence, the `toProtobuf` test use `getEd25519()` method to assert equality, which is different from C++ SDK `toProtobuf` unit test.

**Checklist**


sdk/src/test/java/com/hedera/hashgraph/sdk/TokenInfoTest.java
```diff
@@ -19,69 +19,89 @@
...
+    private static final Duration testAutoRenewPeriod = Duration.ofHours(10);
+    private static final Instant testExpirationTime = Instant.ofEpochSecond(1554158542);
+
     @BeforeAll
     public static void beforeAll() {
         SnapshotMatcher.start();
     }
 
-    @AfterClass
+    @AfterAll
```
rokn, line 94: Why do we need this change?

thenswan: @rokn newer annotation comes from newer junit (jupiter) version, but it is pretty same thing and that's it.

PR #1565, opened 2023-08-21
Add `populateAccountNum()` to `AccountId` and `ContractId` classes

### LINKED ISSUE #1537: Add populateAccountNum() to AccountId/ContractId classes
### Problem
Following [this discussion](https://github.com/hashgraph/hedera-sdk-reference/issues/128):
Whenever you generate `AccountId/ContractId.fromEvmAddress()` the `num` field is set to `0` as a default value as there is no cryptographic relation to the evmAddress and you cannot get the actual `num`

### Solution

We need to add `populateAccountNum()` method to those classes, which queries the mirror node REST API and updates the `num` field of the `AccountId/ContractId`. The method should also return the object itself

### Alternatives

_No response_

thenswan:
**Description**:
- added checks inside `fromSolidityAddress()` and, based on the result, pass the flow to the proper method ([Reference PR](https://github.com/hashgraph/hedera-sdk-js/pull/1741/files));
- added `populateAccountNum()` to `AccountId` and `ContractId` classes ([Reference PR](https://github.com/hashgraph/hedera-sdk-go/pull/784/files)).

**Related issue(s)**:
Fixes #1537 

**Checklist**

### DIFF
--- a/sdk/src/main/java/com/hedera/hashgraph/sdk/AccountId.java
+++ b/sdk/src/main/java/com/hedera/hashgraph/sdk/AccountId.java
@@ -22,9 +22,10 @@ package com.hedera.hashgraph.sdk;
 import com.google.protobuf.ByteString;
 import com.google.protobuf.InvalidProtocolBufferException;
 import com.hedera.hashgraph.sdk.proto.AccountID;
-import java.io.IOException;
 import java.util.Arrays;
 import java.util.Objects;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 import java.util.regex.Pattern;
 import javax.annotation.Nonnegative;
 import javax.annotation.Nullable;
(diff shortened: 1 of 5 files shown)


sdk/src/main/java/com/hedera/hashgraph/sdk/AccountId.java
```diff
@@ -206,14 +207,18 @@ public static AccountId fromEvmAddress(EvmAddress evmAddress) {
...
+        } else {
+            return new AccountId(
+                shard,
+                realm,
+                0,
+                null,
+                null,
+                evmAddress
+            );
+        }
```
rokn, line 221: I think we should revert this as well

thenswan: @rokn updated PR

rokn (approved):

rokn (approved):

PR #1570, opened 2023-08-24
Support for issue #1569

### LINKED ISSUE #1569: Better support for sending transactions to a specific proxies and custom networks
### Problem

There are certain times where we want to send a transaction to a particular node. For example to understand if a particular node is having problems, sending a test query to that node is something we are required to do.

I imagine this is also a problem for targeting local instances? In my use-case I want to hit specific proxies for previewnet, testnet, and mainnet.

The SDK makes this difficult to do. You can craft a network like so:

```
  Map<String, AccountId> myNetwork = new HashMap<>();
  myNetwork.put("127.0.0.1:50211", new AccountId(3));
  Client client = Client.forNetwork(myNetwork)
```

There are some very big downsides to this approach: for starters, it does not support TLS. That is because no attempt to load the address book is loaded, as that path is not executed. Client provides a `setTransportSecurity()` flag but this appears to have no effect.

If I didn't want to disable the transport security and provide certificate, I have no way of doing that either. I might be able to do it by loading from a json config but there is no documentation or examples on the structure of that configuration, so I don't know how to use that method.

### Solution

Provide methods to initialize a network like mainnet and then restrict it to a single proxy?

Or alternatively change a lot of the accessors? 

Client is a final class, for example. There's very useful methods to load from an address book I could probably use to accomplish my use-case, but that are private or package protected so I can't use them.

Similarly, the Network class which does a lot of the heavily lifting of Client is also package protected, so I can't use it.

### Alternatives

There's not alot honestly. Short of patching the SDK and cutting my own library I don't see a way to accomplish what I'm trying to do.

thenswan:
**Description**:
Support for issue #1569.

**Related issue(s)**:
Fixes #1569

### DIFF
--- a/examples/src/main/java/IssueExample.java
+++ b/examples/src/main/java/IssueExample.java
@@ -1,0 +1,139 @@
+/*-
+ *
+ * Hedera Java SDK
+ *
+ * Copyright (C) 2020 - 2023 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+import com.google.errorprone.annotations.Var;
+import com.hedera.hashgraph.sdk.AccountId;
+import com.hedera.hashgraph.sdk.Client;
+import com.hedera.hashgraph.sdk.PrecheckStatusException;
+import com.hedera.hashgraph.sdk.PrivateKey;
+import io.github.cdimascio.dotenv.Dotenv;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Map;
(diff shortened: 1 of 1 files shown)


examples/src/main/java/IssueExample.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    // However here we can set network map with only specific nodes populated from address book and it would work.
+    public static void TLSWithoutNodeAddresses2() throws PrecheckStatusException, TimeoutException, InterruptedException {
+        Client client = Client.forName(HEDERA_NETWORK).setTransportSecurity(true).setVerifyCertificates(true);
+        client.setOperator(OPERATOR_ID, OPERATOR_KEY);
+
+        var network = client.getNetwork();
+
+        Entry<String, AccountId> firstEntry = network.entrySet().iterator().next();
```
dalvizu, line 131: Is this confirmed to be a TLS node and this passes?

I don't know order that is returned by this set, and I believe order of set is not guaranteed? But if it returns first element mirrornode returns it [will be non tls endpoint](https://testnet.mirrornode.hedera.com/api/v1/network/nodes). Would be good to know. If this works, this can be considered a work-around

thenswan: Hey @dalvizu, so:
- I can confirm the order of the set is not guaranteed;
- When setting `.setTransportSecurity(true).setVerifyCertificates(true)` -- `Client#getNetwork` returns `:50212` endpoints, which are TLS ones;

(You can confirm it in the debugger by checking the `network` variable 🙂)

Also, the mirror node shouldn’t be part of those networks. You can choose whichever account you want to test from the `network`.

PR #1587, opened 2023-09-04
Fix generated transaction IDs are not unique issue.

### LINKED ISSUE #1557: Generated transaction IDs are not unique
### Description

I generate many transactions in parallel from the same account (it's a setup stage in a benchmark built on top of the Java Client SDK; the transactions are `AccountCreateTransaction`). Occasionally, the same ID is generated for two different transactions. If those transactions are submitted to the same node I get the `DUPLICATE_TRANSACTION` error for the last one. If, however, they are submitted to different nodes, it becomes possible to get a receipt for a different transaction that is accepted as valid. In particular, `AccountCreateTransaction` receives a wrong account ID that can not be used with the key provided with the transaction.
The problem is with the `TransactionId.generate()` method:
```
    public static TransactionId generate(AccountId accountId) {
        Instant instant = Clock.systemUTC().instant().minusNanos((long) (Math.random() * 5000000000L + 8000000000L));
        return new TransactionId(accountId, instant);
    }
```
The randomization works OK at a low speed but at a high TPS (up to 25K in my case) the probability of generating the same value becomes significant. A strictly monotonic clock source is required here. Unfortunately, existing Java clock sources do not provide such guarantee in multithreaded environment. 

Thanks to the possibility of setting my own transaction ID, I have been able to implement the following scheme that works reliably in my benchmark:
```
    private final long adjustment = System.currentTimeMillis() * 1_000_000L - System.nanoTime();
    private final AtomicLong monotonicTime = new AtomicLong();

    private TransactionId generateTxnId(Client client) {
        long currentTime;
        long lastTime;
        do {
            currentTime = System.nanoTime() + adjustment;
            lastTime = monotonicTime.get();
            if (currentTime <= lastTime) currentTime = lastTime + 1;
        } while (!monotonicTime.compareAndSet(lastTime, currentTime));
        return new TransactionId(client.getOperatorAccountId(), Instant.ofEpochSecond(0, currentTime));
    }
```
Something along those lines should be implemented in Client SDKs.

### Steps to reproduce

I used [Load Generator](https://github.com/hashgraph/network-load-generator) to run `CryptoTransferLoadTest` against a 7-node network.

### Additional context

_No response_

### Hedera network

other

### Version

v2.24.0

### Operating system

Linux

thenswan:
**Description**:
Fix generated transaction IDs are not unique issue.

**Related issue(s)**:
Fixes #1557 

**Checklist**

### DIFF
--- a/sdk/src/main/java/com/hedera/hashgraph/sdk/TransactionId.java
+++ b/sdk/src/main/java/com/hedera/hashgraph/sdk/TransactionId.java
@@ -63,7 +63,12 @@ public final class TransactionId implements Comparable<TransactionId> {
     @Nullable
     private Integer nonce = null;
 
+    private static final long MICROSECONDS_PER_MILLISECOND = 1_000_000L;
+
+    private static final long CLOCK_DRIFT_THRESHOLD = 1_000L;
+
     private static final AtomicLong monotonicTime = new AtomicLong();
+
 
     /**
      * No longer part of the public API. Use `Transaction.withValidStart()` instead.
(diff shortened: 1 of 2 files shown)


sdk/src/main/java/com/hedera/hashgraph/sdk/TransactionId.java
```diff
@@ -97,8 +98,19 @@ public static TransactionId withValidStart(AccountId accountId, Instant validSta
...
+        long currentTime;
+        long lastTime;
+
+        do {
+            currentTime = System.currentTimeMillis() * 1_000_000L;
+            lastTime = monotonicTime.get();
+            if (currentTime <= lastTime) currentTime = lastTime + 1000L;
+        }
+
+        while (!monotonicTime.compareAndSet(lastTime, currentTime));
```
rokn, line 110: Can we add a comment here and maybe extract the consts for more code clarity

thenswan: @rokn please check it now :)

rokn (approved):

sdk/src/main/java/com/hedera/hashgraph/sdk/TransactionId.java
```diff
@@ -64,6 +63,13 @@ public final class TransactionId implements Comparable<TransactionId> {
     @Nullable
     private Integer nonce = null;
 
+    private static final long MICROSECONDS_PER_MILLISECOND = 1_000_000L;
```
OlegMazurov, line 66: The name is not right, should be `NANOSECONDS_PER_MILLISECOND`.

sdk/src/main/java/com/hedera/hashgraph/sdk/TransactionId.java
```diff
@@ -64,6 +63,13 @@ public final class TransactionId implements Comparable<TransactionId> {
...
+
+    private static final long CLOCK_DRIFT_THRESHOLD = 1_000L;
```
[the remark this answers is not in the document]
thenswan: Please check #1605

sdk/src/main/java/com/hedera/hashgraph/sdk/TransactionId.java
thenswan: Please check #1605

PR #1732, opened 2024-02-19
build: Kotlin DSL / Convention Plugins / Java Modules

jjohannes:
**Description**:

This PR does the full package of **Gradle + Java Module** updates similar to what we did in [hedera-services](https://github.com/hashgraph/hedera-services):

- All Gradle config is now in Kotlin DSL
- Gradle config done in _convention plugins_
- Latest Gradle Version 8.8
- Update `io.grpc:*` dependencies to 1.64.0 (to be Java Module compatible)
- Configuration Cache enabled
- Integration tests and example classes are moved into _packages_ to make them Java Module compatible
- Integration tests and samples run on Module Path when executed via Gradle (locally or on CI)
- `module-info.java` files added to `sdk/main`,  `sdk/integrationTest` and `examples`
- Dependencies are declared in `module-info.java` and scope is defined according to `checkAllModuleInfo`
- Versions are centrally declared in separate block in `sdk-dependency-versions/build.gradle.kts`
- Building the **sdk-full** variant is now fully done in Gradle through a separate `sdk-full` project that has a separate `module-info` (exchanging `requires grpc.protobuf.lite` for `requires grpc.protobuf`) but otherwise re-uses sources and proto buffs from the `sdk/main` sources folder.
- The `sdk` and `sdk-full` are now both published by the Gradle build in the one publishing run
- Cleanup: remove config files for no longer used build/ci solutions (Built, GitLab)

Side-effects:
- `io.grpc:grpc-netty-shaded` in client project is automatically aligned with the `grpc-api` version we use

Manual Testing (for things not covered by the automatic tests):
- I tested that the Android app example still builds and runs
- I tested the publishing by temporarily adding a local repository

Follow ups:
- https://github.com/hashgraph/hedera-sdk-java/issues/1794
- https://github.com/hashgraph/hedera-services/issues/13505 We should adjust the _code style rules and tools_ to what ist done in _hedera-services_ as well. E.g. remove sonar, introduce spotless, decide if you want to keep _error prone_ etc. Since there are some decissions to be made, I would do that as a later follow up  - maybe once this is "stable" in  _hedera-services_.


examples/src/main/java/com/hedera/hashgraph/sdk/examples/ContractHelper.java
```diff
@@ -59,7 +59,7 @@ public class ContractHelper {
     final Map<Integer, AccountId> stepFeePayers = new HashMap<>();
 
     public static String getBytecodeHex(String filename) throws IOException {
-        try (InputStream jsonStream = ContractHelper.class.getClassLoader().getResourceAsStream(filename)) {
+        try (InputStream jsonStream = ContractHelper.class.getResourceAsStream(filename)) {
```
jjohannes, line 62: Production code change: resources need to be loaded "inside the module" when this runs on the module path.

sdk/src/test/java/com/hedera/hashgraph/sdk/AccountAllowanceDeleteTransactionTest.java
```diff
@@ -43,7 +42,7 @@ public static void beforeAll() {
         SnapshotMatcher.start();
     }
 
-    @AfterClass
+    @AfterAll
```
jjohannes, line 45: This was wrong in many places - missed because JUnit4 was still one the classpath.

nathanklick:
@jjohannes Please rebase this PR when you get a chance. The team is ready to review and merge.

jjohannes:
@nathanklick rebased and ready for review

thenswan:
I merged [PR related to the docs update](https://github.com/hashgraph/hedera-sdk-java/pull/1621) (recommend to check it out), and we have [a doc on creating a fat/uber JAR](https://github.com/hashgraph/hedera-sdk-java/blob/main/docs/java-app/create-jar-guide.md) based on the user's request. Would like to ask if is it possible for jar creation to be done as a convention plugin (like in hedera-services)?

thenswan:
Proposing change `com.hedera.hashgraph.sdk.integration.test` to `com.hedera.hashgraph.sdk.test.integration`.

gradle/plugins/src/main/kotlin/com.hedera.gradlebuild.patch-modules.gradle.kts
```diff
@@ -0,0 +1,154 @@
...
+    module("io.github.json-snapshot:json-snapshot", "json.snapshot")
+    module("junit:junit", "junit")
+    module("org.mockito:mockito-core", "org.mockito")
+    module("org.mockito:mockito-junit-jupiter", "org.mockito.junit.jupiter")
+    module("org.objenesis:objenesis", "org.objenesis")
+}
+
+dependencies {
+    javaModulesMergeJars("com.google.code.findbugs:jsr305:3.0.2")
+}
```
thenswan, line 154: Can you please explain purpose of this, when we have above
```
module("javax.annotation:javax.annotation-api", "java.annotation") {
        exportAllPackages()
        mergeJar("com.google.code.findbugs:jsr305")
}
```

thenswan:
I am seeing a following error, when running an example via Intellij (works fine when running via cmd, maybe it can be fixed by updating gradle, see [this discussion](https://github.com/gradle/gradle/issues/21364#issuecomment-2040972594)):
```
FAILURE: Build failed with an exception.

* What went wrong:
Configuration cache problems found in this build.

1 problem was found storing the configuration cache.
- Task `:examples:TransferCryptoExample.main()` of type `org.gradle.api.tasks.JavaExec`: cannot serialize object of type 'java.io.PipedInputStream', a subtype of 'java.io.InputStream', as these are not supported with the configuration cache.
  See https://docs.gradle.org/8.6/userguide/configuration_cache.html#config_cache:requirements:disallowed_types

See the complete report at file:///Users/hedera-sdk-java/build/reports/configuration-cache/4ektp07xks1k5f40qn2zxqbfv/2hvheo6cndbpywiqj8ir9e180/configuration-cache-report.html

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Get more help at https://help.gradle.org.

BUILD FAILED in 23s
```

thenswan:
Running integration tests [using configuration file](https://github.com/hashgraph/hedera-sdk-java/blob/main/docs/sdk/developer-guide.md#using-configuration-file) is not working anymore.

thenswan:
I am not able to run an Android app (but was able to run it from the `main` branch). Build fail with:
```
Could not determine the dependencies of task ':example-android:app:compileDebugJavaWithJavac'.
> Could not resolve all task dependencies for configuration ':example-android:app:debugCompileClasspath'.
   > Could not find com.google.protobuf:protobuf-javalite:.
     Required by:
         project :example-android:app > project :sdk
   > Could not find com.esaulpaugh:headlong:.
     Required by:
         project :example-android:app > project :sdk
   > Could not find io.grpc:grpc-stub:.
     Required by:
         project :example-android:app > project :sdk
   > Could not find io.helidon.grpc:io.grpc:.
     Required by:
         project :example-android:app > project :sdk
   > Could not find org.bouncycastle:bcprov-jdk18on:.
     Required by:
         project :example-android:app > project :sdk
```

sdk-full/src/main/java/module-info.java
```diff
@@ -0,0 +1,22 @@
+module com.hedera.hashgraph.sdk.full {
+    requires transitive com.google.common;
+    requires transitive com.google.protobuf;
+    requires transitive headlong;
+    requires transitive io.grpc.stub;
+    requires transitive io.grpc;
+    requires transitive org.bouncycastle.provider;
+    requires transitive org.slf4j;
```
thenswan, line 8: Why those are transitive?

sdk/src/main/java/module-info.java
```diff
@@ -0,0 +1,22 @@
+module com.hedera.hashgraph.sdk {
+    requires transitive com.google.common;
+    requires transitive com.google.protobuf;
+    requires transitive headlong;
+    requires transitive io.grpc.stub;
+    requires transitive io.grpc;
+    requires transitive org.bouncycastle.provider;
+    requires transitive org.slf4j;
```
thenswan, line 8: Why those are transitive?

thenswan:
Would be great to add the docs and update current docs which is in the repo. Given the new setup it can be hard for new collaborators to catch up, so docs and guides should be super helpful, for example:
- What is the algorithm for adding new dependency, considering potential changes to [gradle/plugins/src/main/kotlin/com.hedera.gradlebuild.patch-modules.gradle.kts](https://github.com/hashgraph/hedera-sdk-java/pull/1732/files#diff-35a2955e16a68ec66a7fb9a0f4a11d902a09d723f4cb20b892d2a1592a15cdbf)
- etc.

thenswan:
Can you please share how did you perform manual testing (Android app and release process)? 🙏🏻

thenswan:
Please let me know when there will be a time to re-visit having a separate project for the versions only. Now I am thinking about simplicity (in the context of this project) and not having a separate project for the versions only.

thenswan:
Will get back later on the `com.google.code.findbugs:jsr305`.

thenswan (changes requested):
Also branch needs to be updated.

gradle/plugins/src/main/kotlin/com.hedera.gradlebuild.publish.gradle.kts
```diff
@@ -0,0 +1,83 @@
...
+            }
+        }
+
+        scm {
+            url = "https://github.com/hashgraph/hedera-sdk-java"
+            connection = "scm:git:https://github.com/hashgraph/hedera-sdk-java.git"
+            developerConnection = "scm:git:ssh://github.com:hashgraph/hedera-sdk-java.git"
+        }
+
+        developers {
```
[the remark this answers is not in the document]
agadzhalov: definitely

nathanklick:
@jjohannes Please see comments and questions from the Limechain team. Please rebase this PR. The team is ready to review and merge as soon as it is rebased and comments addressed.

.github/workflows/release-artifacts.yml
```diff
@@ -95,7 +95,7 @@ jobs:
           cache-read-only: false
 
       - name: Compile SDK & Javadoc
-        run: ./gradlew assemble :sdk:javadoc -Dfile.encoding=UTF-8 --scan
+        run: ./gradlew assemble :sdk:javadoc --scan
 
       - name: Nexus Release
-        run: ./gradlew publishToSonatype closeAndReleaseSonatypeStagingRepository --no-parallel -Dfile.encoding=UTF-8 --scan -PsonatypeUsername=${{ secrets.SONATYPE_USERNAME }} -PsonatypePassword=${{ secrets.SONATYPE_PASSWORD }}
```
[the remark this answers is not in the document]
jjohannes: It is sufficient to call `closeAndReleaseSonatypeStagingRepository` now. Everything else is wired up in Gradle. Concretely, the following makes sure **sdk** and **sdk-full** are published before the `closeSonatypeStagingRepository` which in turn is done before the `closeAndReleaseSonatypeStagingRepository` (this relationship is already setup by the plugin).

```
tasks.named("closeSonatypeStagingRepository") {
    // The publishing of all components to Maven Central is automatically done before close
    // (which is done before release).
    dependsOn(":sdk:publishToSonatype")
    dependsOn(":sdk-full:publishToSonatype")
}
```

I know this is not very straight forward. We have a similar setup in https://github.com/hashgraph/pbj where we tried different things and ended up with his.

sdk/build.gradle.kts
```diff
@@ -0,0 +1,56 @@
...
+    version("org.assertj.core", "3.24.2")
+    version("org.apache.commons.lang3", "3.14.0")
+    version("grpc.netty.shaded", "1.57.2")
+}
+
+testModuleInfo {
+    requires("org.assertj.core")
+    requires("org.junit.jupiter.api")
+    requires("org.junit.jupiter.params")
+    requires("org.mockito")
```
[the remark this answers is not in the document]
jjohannes: For tests I only added the minimum:
- `jupiter` does not need a version, because Gradle uses one of the latest as default
- `json.snapshot` brings in versions for "mockito"

More details: https://scans.gradle.com/s/awnnigj7pabpu/dependencies?configurationFilter=WyJ0ZXN0UnVudGltZUNsYXNzcGF0aCJd&expandAll&focusedDependencyView=versions&projectFilter=WyI6c2RrIl0&toggled=W1swLDAsWzE2NV1dLFswLDAsWzE2NSw0NjhdXV0

(You can run `./gradlew sdk:dependencies --scan` to get such a build scan with all the dependencies)

If you prefer, we can add versions for these dependencies as well.

build.gradle
```diff
@@ -1,111 +0,0 @@
...
-            property "sonar.projectKey", "hashgraph_hedera-sdk-java"
-            property "sonar.organization", "hashgraph"
-            property "sonar.host.url", "https://sonarcloud.io"
-            property "sonar.exclusions", "examples/**"
-        }
-    }
-
-    tasks.withType(Test) {
-        apply plugin: "jacoco"
-    }
```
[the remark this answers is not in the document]
jjohannes: This configuration was ill placed. It should not be inside a _tasks_ configuration. `apply plugin: "jacoco"` (old style) is the same as `id("jacoco")` in the `plugins` block in the top of the file. This one should also have been at the top.

It is now in the `plugins {}` block of `com.hedera.gradle.java-base.gradle.kts`.

gradle/plugins/src/main/kotlin/com.hedera.gradlebuild.java-base.gradle.kts
```diff
@@ -0,0 +1,117 @@
...
+        excludedPaths = ".*generated.*"
+    }
+}
+
+tasks.withType<AbstractArchiveTask>().configureEach {
+    isPreserveFileTimestamps = false
+    isReproducibleFileOrder = true
+    fileMode = 436 // octal: 0664
+    dirMode = 509 // octal: 0775
+}
```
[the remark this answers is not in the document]
jjohannes: This is to get reproducible builds (should be a Gradle default, but unfortunately is not - https://github.com/gradle/gradle/issues/28806). Without his, Zip/Jar files built from the same commit may differ depending on build time or operating system.

gradle/plugins/src/main/kotlin/com.hedera.gradlebuild.java.gradle.kts
```diff
@@ -0,0 +1,153 @@
...
+    }
+    plugins {
+        id("grpc") {
+            artifact = "io.grpc:protoc-gen-grpc-java:1.58.0"
+        }
+    }
+}
+tasks.generateProto {
+    plugins { plugins.register("grpc") }
+}
```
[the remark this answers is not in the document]
jjohannes: This was here before: https://github.com/hashgraph/hedera-sdk-java/blob/11733c2cd0820c4f2affb1a52ce003bbf960ee20/sdk/protobuf.gradle#L12

(With a "shorthand hard to read" Groovy syntax)

I am not an expert on the probuf stuff and preserved what we had. My understanding is that this adds the "protobuf plugin" (not a Gradle plugin, but a plugin for protobuf code generation itself) for "grpc" code generation.

gradle/plugins/src/main/kotlin/com.hedera.gradlebuild.java-base.gradle.kts
```diff
@@ -0,0 +1,117 @@
...
+    }
+}
+
+jacoco {
+    toolVersion = "0.8.8"
+}
+
+javaModuleDependencies {
+    versionsFromPlatformAndConsistentResolution(":sdk", ":sdk")
+}
```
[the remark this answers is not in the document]
jjohannes: This is configuring the **central** and **consistent** version management. It says:

- Arg 1: Versions can be defined in the `sdk` project (here we could choose a separate project for versions).
- Arg 2: `sdk` is also the main thing we build - all versions should be consistent with the `runtimeClasspath` of `sdk`. Meaning, also when we compile code or run tests, the same versions should be used. Otherwise, versions can differ if transitive dependencies that are sometimes hidden (e.g. at compile time) cause version upgrades.

gradle/plugins/src/main/kotlin/com.hedera.gradlebuild.java.gradle.kts
```diff
@@ -0,0 +1,153 @@
...
+
+tasks.jar {
+    exclude("**/*.proto")
+    includeEmptyDirs = false
+}
+
+sourceSets.all {
+    configurations[getTaskName("", "compileProtoPath")].extendsFrom(configurations["internal"])
+}
```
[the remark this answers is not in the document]
jjohannes: The Gradle prtotobuf plugin has its own "classpath" for the protobuf compiler tool. There is one for each "sourceSet" (e.g for _protobufs_ in main and protobufs in _test_). They are named after the source set. E.g. "compileProtoPath" for "main" or "testCompileProtoPath" for "test".

In Gradle, all classpaths (or module paths) are represented by the Gradle concept `Configuration`. Here we grep the configurations the protobuf uses and say they should inherit everything from the "internal" `Configuration`. Which is the one that has all the _versions_ attached we define centrally in `sdk/build.gradle.kts`

Unfortunately this kind of "glue" is often needed because version management is _complicated_ (sometimes you want everything to be consistent, sometimes you want different versions on different claspaths etc). And there are different solutions and no one-fit-all for all projects. That why some things in this area do not "just work".

sdk/src/main/java/module-info.java
jjohannes: The `checkAllModuleInfo` dependency scope check tells us if a type of these dependencies is used in public API. Then the dependency should be transitive.

You can run `./gradlew :sdk:reason --id com.google.guava:guava` (or another `--id`) for details and then search the code base for the places. Maybe some things are unintentional.

sdk-full/src/main/java/module-info.java
jjohannes: See comment below. 

This is an almost exact copy of the file in `sdk` except for exchanging 
`requires grpc.protobuf.lite;` and `requires grpc.protobuf;`

The dependency scope check will make sure that you do not forget to change dependencies in both files if you do in the future. It will do the same analysis for both.

gradle/plugins/src/main/kotlin/com.hedera.gradlebuild.patch-modules.gradle.kts
jjohannes: This is needed to define a version for `jsr305`. The above only contains IDs pointing at Jars. Before that configuration can be used, Jars need to be fetched by their version. Usually all versions are ind `sdk/build.gradle.kts` now. But this one does not "really exist" because it is merged. So it does not have a "Module Name" to address it.

But this is super complicated 😞 and we should get rid of this. It's all only here because both Jars contain the `javax.annotation` package. Ideally we could check the code if we can change something about which annotations are used to get rid of this.

I'll put this into a follow up issue.

sdk/build.gradle.kts
```diff
@@ -0,0 +1,56 @@
...
+
+moduleInfo {
+    version("com.github.spotbugs.annotations", "4.7.3")
+    version("com.google.common", "33.0.0-jre")
+    version("com.google.errorprone.annotations", "2.21.1")
+    version("com.google.gson", "2.10.1")
+    version("grpc.protobuf.lite", "1.46.0")
+    version("grpc.protobuf", "1.46.0")
+    version("headlong", "10.0.0")
+    version("io.grpc", "3.2.1")
```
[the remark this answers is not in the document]
jjohannes: These are both updates for the _Java Module_ setup. 
- Guava (com.google.common) is transitively on the path already, but we want a newer version. 
- `io.grpc` is the modularized replacement for several `io.grpc:*` Jars (com.hedera.gradle.patch-modules.gradle.kts and commetns in the file). We do the same in _hedera-services_.

One follow up could be, when you have this merged with the `checkAllModuleInfo` dependency analysis task, to check if some dependencies can be removed altogether (Guava for example).

sdk/build.gradle.kts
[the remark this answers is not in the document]
thenswan: I am good with this setup for now, thanks 🙂

gradle/plugins/src/main/kotlin/com.hedera.gradlebuild.java-base.gradle.kts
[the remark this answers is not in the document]
thenswan: Thanks @jjohannes. As a follow up on this -- do we need to do something similar ([services PR](https://github.com/hashgraph/hedera-services/pull/13108)) in this repo?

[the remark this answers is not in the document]
jjohannes: Yes we should. Ideally, we would use the exact same setup in both repos.

I would prefer to get this PR merged now as it is so huge. And then move more things from "services" over afterwards.

sdk/src/main/java/module-info.java
thenswan: @jjohannes I think to leave same transitive dependencies as we have now and add new transitive dependencies as a follow up item later along with communicating it to the community as it might potentially break apps.

sdk-full/src/main/java/module-info.java
thenswan: My latest comment below affects this conversation as well.

thenswan (commented):
Please explain why this change was introduced

sdk/src/main/java/module-info.java
jjohannes: That makes sense.

I changed this to only have `com.google.protobuf` (com.google.protobuf:protobuf-javalite) as _transitive_ (api) dependency as it currently is on `main`.

For the analysis, I added the corresponding configuration in [build.gradle.kts](https://github.com/hashgraph/hedera-sdk-java/blob/gradle-updates/build.gradle.kts#L7-L16) you can then decide yourself what to do for each case (or just leave things as they are).

sdk-full/src/main/java/module-info.java
jjohannes: Correponding change done here as well.

sdk/src/main/java/module-info.java
thenswan: Thanks @jjohannes 

I think we should only have protobuf dependency as transitive now without exposing others like in [build.gradle.kts](https://github.com/hashgraph/hedera-sdk-java/blob/gradle-updates/build.gradle.kts#L7-L16) 
So its 1:1 like in `main`

thenswan:
`Taskfile.yml`: `publish` command is irrelevant -- we can remove it.

thenswan:
How can we address this:
```
> Configure project :sdk
The 'sonarqube' task depends on compile tasks. This behavior is now deprecated and will be removed in version 5.x. To avoid implicit compilation, set property 'sonar.gradle.skipCompile' to 'true' and make sure your project is compiled, before analysis has started.
The 'sonar' task depends on compile tasks. This behavior is now deprecated and will be removed in version 5.x. To avoid implicit compilation, set property 'sonar.gradle.skipCompile' to 'true' and make sure your project is compiled, before analysis has started.

> Configure project :sdk-full
The 'sonarqube' task depends on compile tasks. This behavior is now deprecated and will be removed in version 5.x. To avoid implicit compilation, set property 'sonar.gradle.skipCompile' to 'true' and make sure your project is compiled, before analysis has started.
The 'sonar' task depends on compile tasks. This behavior is now deprecated and will be removed in version 5.x. To avoid implicit compilation, set property 'sonar.gradle.skipCompile' to 'true' and make sure your project is compiled, before analysis has started.
```

thenswan:
A couple of questions based on analysis of the output of `./gradlew sdk:dependencies --scan`:
1. Does it make sense for `compileProtoPath` and `testIntegrationCompileProtoPath` to have all of those dependencies? 
2. It is OK that `project :sdk (*)` is included twice in: `testIntegrationCompileClasspath`, `testIntegrationCompileProtoPath` and `testIntegrationRuntimeClasspath`?
3. Is it OK to have FAILED here:
```
mainRuntimeClasspath
+--- project :sdk (*)
+--- com.google.guava:guava FAILED
+--- com.google.protobuf:protobuf-javalite FAILED
+--- com.esaulpaugh:headlong FAILED
+--- io.grpc:grpc-stub FAILED
+--- io.helidon.grpc:io.grpc FAILED
+--- org.bouncycastle:bcprov-jdk18on FAILED
+--- org.slf4j:slf4j-api FAILED
+--- com.google.code.gson:gson FAILED
+--- io.grpc:grpc-protobuf-lite FAILED
\--- org.bouncycastle:bcpkix-jdk18on FAILED
```
4. and here:
```
errorprone - Error Prone dependencies, will be extended by all source sets' annotationProcessor configurations
+--- com.google.errorprone:error_prone_core:2.21.1
|    +--- com.google.errorprone:error_prone_annotation:2.21.1
|    |    \--- com.google.guava:guava:32.1.1-jre
|    |         +--- com.google.guava:guava-parent:32.1.1-jre
|    |         |    +--- com.google.code.findbugs:jsr305:3.0.2 (c)
|    |         |    +--- org.checkerframework:checker-qual:3.33.0 (c)
|    |         |    \--- com.google.errorprone:error_prone_annotations:2.18.0 -> 2.21.1 (c)
|    |         \--- com.google.guava:failureaccess:1.0.1
|    +--- com.google.errorprone:error_prone_type_annotations:2.21.1
|    +--- com.google.errorprone:error_prone_check_api:2.21.1
|    |    +--- com.google.errorprone:error_prone_annotation:2.21.1 (*)
|    |    +--- com.google.code.findbugs:jsr305:3.0.2
|    |    +--- io.github.eisop:dataflow-errorprone:3.34.0-eisop1
|    |    +--- io.github.java-diff-utils:java-diff-utils:4.0
|    |    |    \--- org.eclipse.jgit:org.eclipse.jgit:4.4.1.201607150455-r
|    |    +--- com.google.auto.value:auto-value-annotations:1.9
|    |    +--- com.google.errorprone:error_prone_annotations:2.21.1
|    |    +--- com.github.kevinstern:software-and-algorithms:1.0
|    |    +--- com.github.ben-manes.caffeine:caffeine:3.0.5
|    |    |    +--- org.checkerframework:checker-qual:3.19.0 -> 3.33.0
|    |    |    \--- com.google.errorprone:error_prone_annotations:2.10.0 -> 2.21.1
|    |    \--- com.google.inject:guice:5.1.0
|    |         +--- javax.inject:javax.inject:1
|    |         +--- aopalliance:aopalliance:1.0
|    |         \--- com.google.guava:guava:30.1-jre -> 32.1.1-jre (*)
|    +--- org.pcollections:pcollections:3.1.4
|    +--- com.google.guava:guava:32.1.1-jre (*)
|    +--- com.google.auto:auto-common:1.2.1
|    |    \--- com.google.guava:guava:31.0.1-jre -> 32.1.1-jre (*)
|    +--- com.google.code.findbugs:jsr305:3.0.2
|    +--- io.github.eisop:dataflow-errorprone:3.34.0-eisop1
|    +--- com.google.auto.value:auto-value-annotations:1.9
|    +--- com.google.errorprone:error_prone_annotations:2.21.1
|    +--- com.google.protobuf:protobuf-java:3.19.6
|    +--- com.google.auto.service:auto-service-annotations:1.0.1
|    +--- javax.inject:javax.inject:1
|    \--- javax.annotation:javax.annotation-api FAILED
+--- com.uber.nullaway:nullaway:0.10.14
|    +--- org.checkerframework:dataflow-nullaway:3.38.0
|    \--- com.google.guava:guava:24.1.1-jre -> 32.1.1-jre (*)
\--- io.grpc:grpc-java-api-checker:1.1.0
```

thenswan:
Android example still not working for me.

- When I am trying to build it as is - I get a lot of `Duplicate class io.grpc` errors:
```
FAILURE: Build completed with 2 failures.

1: Task failed with an exception.
-----------
* What went wrong:
Execution failed for task ':app:checkDebugDuplicateClasses'.
> A failure occurred while executing com.android.build.gradle.internal.tasks.CheckDuplicatesRunnable
   > Duplicate class io.grpc.Attributes found in modules grpc-api-1.61.1.jar -> grpc-api-1.61.1 (io.grpc:grpc-api:1.61.1) and io.grpc-3.2.1.jar -> io.grpc-3.2.1 (io.helidon.grpc:io.grpc:3.2.1)
     Duplicate class io.grpc.Attributes$1 found in modules grpc-api-1.61.1.jar -> grpc-api-1.61.1 (io.grpc:grpc-api:1.61.1) and io.grpc-3.2.1.jar -> io.grpc-3.2.1 (io.helidon.grpc:io.grpc:3.2.1)
     Duplicate class io.grpc.Attributes$Builder found in modules grpc-api-1.61.1.jar -> grpc-api-1.61.1 (io.grpc:grpc-api:1.61.1) and io.grpc-3.2.1.jar -> io.grpc-3.2.1 (io.helidon.grpc:io.grpc:3.2.1)
     Duplicate class io.grpc.Attributes$Key found in modules grpc-api-1.61.1.jar -> grpc-api-1.61.1 (io.grpc:grpc-api:1.61.1) and io.grpc-3.2.1.jar -> io.grpc-3.2.1 (io.helidon.grpc:io.grpc:3.2.1)
     Duplicate class io.grpc.BinaryLog found in modules grpc-api-1.61.1.jar -> grpc-api-1.61.1 (io.grpc:grpc-api:1.61.1) and io.grpc-3.2.1.jar -> io.grpc-3.2.1 (io.helidon.grpc:io.grpc:3.2.1)
     Duplicate class io.grpc.BindableService found in modules grpc-api-1.61.1.jar -> grpc-api-1.61.1 (io.grpc:grpc-api:1.61.1) and io.grpc-3.2.1.jar -> io.grpc-3.2.1 (io.helidon.grpc:io.grpc:3.2.1)
     Duplicate class io.grpc.CallCredentials found in modules grpc-api-1.61.1.jar -> grpc-api-1.61.1 (io.grpc:grpc-api:1.61.1) and io.grpc-3.2.1.jar -> io.grpc-3.2.1 (io.helidon.grpc:io.grpc:3.2.1)
     ...
```
- When I am commenting `implementation(platform("com.hedera.hashgraph:sdk")) ...` and trying to use `implementation("com.hedera.hashgraph:sdk")` (btw can I use it with version here?):
```
FAILURE: Build failed with an exception.

* What went wrong:
Could not determine the dependencies of task ':app:compileDebugJavaWithJavac'.
> Could not resolve all task dependencies for configuration ':app:debugCompileClasspath'.
   > Could not find com.google.protobuf:protobuf-javalite:.
     Required by:
         project :app > project :hedera-sdk-java:sdk
```

thenswan:
Integration tests.

I get following error once, when changing some code and then trying to run integration test. After trying to run integration test the second time - nothing really happens. Sample output:
```
> Task :sdk:compileTestIntegrationJava FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':sdk:compileTestIntegrationJava'.
> Could not resolve all files for configuration ':sdk:testIntegrationCompileClasspath'.
   > Failed to transform sdk-2.32.0-beta.1.jar (project :sdk) to match attributes {artifactType=jar, javaModule=true, org.gradle.category=library, org.gradle.dependency.bundling=external, org.gradle.jvm.version=17, org.gradle.libraryelements=jar, org.gradle.usage=java-api}.
      > Execution failed for ExtraJavaModuleInfoTransform: /Users/nikitalebedev/Work/LimeChain/projects-workzone/hedera-sdk-java/sdk/build/libs/sdk-2.32.0-beta.1.jar.
         > Not a module and no mapping defined: sdk-2.32.0-beta.1.jar

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Get more help at https://help.gradle.org.

BUILD FAILED in 7s
19 actionable tasks: 6 executed, 13 up-to-date
Configuration cache entry stored.
```

thenswan:
**Docs**

- Can you please update following docs based on the changes you introduced 🙏🏻 
  - [Integration tests](https://github.com/hashgraph/hedera-sdk-java/blob/main/docs/sdk/developer-guide.md#integration-tests)
  - [Android](https://github.com/hashgraph/hedera-sdk-java/blob/main/docs/sdk/developer-guide.md#integration-tests) 
- Also would like to ask to make following item included in this PR (NOT as a follow up): 
> Would be great to add the docs and update current docs which is in the repo. Given the new setup it can be hard for new collaborators to catch up, so docs and guides should be super helpful, for example:

gradle/plugins/src/main/kotlin/com.hedera.gradle.repositories.gradle.kts
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+repositories {
+    mavenCentral()
+    maven("https://jitpack.io") {
+        content { includeModule("io.github.cdimascio", "java-dotenv") }
```
thenswan, line 20: Please help me to understand why this change was introduced

thenswan (changes requested):
Thanks for an amazing job @jjohannes 
Would like to align couple more things before merging

gradle/plugins/src/main/kotlin/com.hedera.gradlebuild.java-base.gradle.kts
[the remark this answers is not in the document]
thenswan: Makes sense to me, just need to track it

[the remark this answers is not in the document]
jjohannes: I created this issue https://github.com/hashgraph/hedera-services/issues/13505 to track it in this list which we use to organize the work I am doing on the builds https://github.com/hashgraph/hedera-services/issues/11230

sdk/src/main/java/module-info.java
jjohannes: Yes that is the case after my changes. Only [`com.google.protobuf` is transitive](https://github.com/hashgraph/hedera-sdk-java/blob/gradle-updates/sdk/src/main/java/module-info.java#L2).

The dependency analysis plugin configuration in [build.gradle.kts](https://github.com/hashgraph/hedera-sdk-java/blob/gradle-updates/build.gradle.kts#L7-L16) are required to tell the dependency analysis to ignore that certain classes expose types from dependencies in their API although we do not declare those dependencies as transitive.

jjohannes:
> How can we address: The 'sonarqube' task depends on compile tasks. This behavior is now deprecated ...

By not using sonarqube anymore or waiting for them to update their Gradle plugin. 😄 I would suggest to remove sonarqube in the follow up https://github.com/hashgraph/hedera-services/issues/13505 and use the other solutions we now use in 'hedera-services' (Codecov and Codacy).

gradle/plugins/src/main/kotlin/com.hedera.gradlebuild.java-base.gradle.kts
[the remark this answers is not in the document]
thenswan: Thanks @jjohannes

thenswan:
>I would suggest to remove sonarqube in the follow up https://github.com/hashgraph/hedera-services/issues/13505 and use the other solutions we now use in 'hedera-services' (Codecov and Codacy).

Agree with aligning it with `hedera-services` 👍🏻

thenswan:
@jjohannes kindly pinging to figure out if you saw [this comment](https://github.com/hashgraph/hedera-sdk-java/pull/1732#issuecomment-2117684849) and others I left on May 17th 🙂

jjohannes:
@thenswan yes, saw your comments. This is next on my list when I have a bit more time again for Hedera. Moste likely in the beginning of next week.

gradle/plugins/src/main/kotlin/com.hedera.gradle.repositories.gradle.kts
jjohannes: This instructs Gradle to **only** look for `java-dotenv` in the "jitpack.io" repo. And it is documentation for us to remember why we need that repo – in addition to `mavenCentral()` – at all.

But your question makes me realize that we should change the order. Which I did now. Gradle goes through the repositories from top to bottom. Now, it will only look for `java-dotenv` in JitPack (and find it). It will not make any other requests to that repo, because of the the `content { ... }` filter. All other requests will go directly to mavenCentral(). With that, we do not have any unnecessary requests to any of the repositories.

jjohannes:
@thenswan worked on this today, and it was more than expected. The first point below is something we need to think about carefully. The other things should be addressed. I have two open TODOs, that I'll do later (could also be done as follow up). You can have a look at everything else. See details below.

#### Issue with Android App / Open Question

The problem with the Android App made me realize something we have not discussed and which might be a problem:
In the new setup, we replace `io.grpc:grpc-api` with `io.helidon.grpc:io.grpc`. Otherwise, things do not work as Java Module due to split package problems. We will then publish the dependency to `io.helidon.grpc:io.grpc`. Which is a difference to the current release. Is this acceptable at all?

In any case, I made sure that the publishing is sound now so that if you depend on `com.hedera.hashgraph:sdk`, you will only get `io.helidon.grpc:io.grpc` transitively and not the overlapping `io.grpc:*` modules. This is done using `exclude` which is written to the pom/module metadata. The Android sample now works.

If we do not want to do this change, we cannot deliver a working `module-info`. If you are unsure about this, maybe we need to discuss this directly together with @nathanklick. 

#### Documentation (guides)

I updated the docs on integration tests and Android.

#### Integration Tests issue

I am not able to reproduce this problem:
```
> Task :sdk:compileTestIntegrationJava FAILED
...
   > Failed to transform sdk-2.32.0-beta.1.jar (project :sdk) to match attributes {artifactType=jar, javaModule=true, org.gradle.category=library, org.gradle.dependency.bundling=external, org.gradle.jvm.version=17, org.gradle.libraryelements=jar, org.gradle.usage=java-api}.
      > Execution failed for ExtraJavaModuleInfoTransform: /Users/nikitalebedev/Work/LimeChain/projects-workzone/hedera-sdk-java/sdk/build/libs/sdk-2.32.0-beta.1.jar.
         > Not a module and no mapping defined: sdk-2.32.0-beta.1.jar
```

I know this issue in general, but don't see it in this project. Also works on CI. Can you try again, and if you still get it, run again with `--scan` and send me the link.

#### Questions about dependencies 

> 1. Does it make sense for **compileProtoPath** and **testIntegrationCompileProtoPath** to have all of those dependencies?

It's the same as on main. This setup comes from Google's protobuf plugin and it inherits all the dependencies that the "compileClasspath" has. If this is necessary, I don't know. But I won't touch this standard setup if there is no good reason to change it.
Note that on this branch, everything is shown twice, because the version (tagged with `constraint` or `(c)`) is shown as a separate entry.

> 2. It is OK that project :sdk (*) is included twice in: testIntegrationCompileClasspath, testIntegrationCompileProtoPath and testIntegrationRuntimeClasspath?

It's not a problem if things are duplicated. But it's interesting why this is. Looks like the first entry is added by Gradle automatically when we setup the tests by `register<JvmTestSuite>("testIntegration")` (by this I am surprised, I'll do some more research is it can be turned off). The second is added, because the "intgration tests" run as Java Module and we have an explicit dependency defined through the `requires com.hedera.hashgraph.sdk;`.

> 3./4. Is it OK to have FAILED here: mainRuntimeClasspath / errorprone

Does not seem to be a problem, but it's not nice. Put it on my TODO list to check.


#### Broken integration tests

I disabled two integration tests on this branch, because they are failing. When I test locally, they also fail on "main" for me. I am not sure if this has something to do with the changes on the branch or if they are already broken and somehow did not run on CI with the old setup. Looks to me, as if the assertions in the tests need to be a bit different. Can you or someone else from the team check if there is already an issue with the tests on main?

These are the tests: https://github.com/hashgraph/hedera-sdk-java/pull/1732/commits/7880fd20233b4f278cbbe88aedf9e93dbb132462

#### TODOs 
 
- [ ] TODO, More comments (docs) in .gradle.kts filesThis is independent of everything else.
- [ ] TODO, check why "mainRuntimeClasspath FAILED" and "errorprone FAILED" and improve

jjohannes:
I moved the versions into a separate file `sdk-dependency-versions/build.gradle.kts` – in alignment with what we do now in _hedera-services_. Sorry for the back-and-forth on this, but I now think it is the better solution as the Dependabot integration keeps working flawlessly this way and the setup in Gradle is also a bit clearer. The versions are also used by the examples and the _sdk-full_ project (and not just the _sdk_ project).

thenswan:
Thank you for detailed answers and updates @jjohannes 
I addressed tests, fix is already on the `main` branch 🙂 

Re Android - will check it more in detail.

thenswan:
**Integration tests**

Seems to be working at this point and I am not able to reproduce same error after new changes 👍🏻

jjohannes:
@thenswan good news on the topic of:

> Issue with Android App / Open Question
> The problem with the Android App made me realize something we have not discussed and which might be a problem:
In the new setup, we replace io.grpc:grpc-core with io.helidon.grpc:io.grpc. Otherwise, things do not work as Java Module due to split package problems. We will then publish the dependency to io.helidon.grpc:io.grpc. Which is a difference to the current release. Is this acceptable at all?

We are updating 'io.grpc' in the services repo (https://github.com/hashgraph/hedera-services/compare/13679-tm-upgrade-netty) with the newest version. Turns out, the replacement stuff is no longer necessary and we can keep using `grpc-api` (or `grpc-core`) directly. Then this whole problematic change goes away.
I will update this PR accordingly and let you know when I am done. Then you can have another look.

jjohannes:
@thenswan I did the changes and tested the PR again. I did some more tweaks to the docs for the examples and Android example. I also added building the Android example to the GH Actions pipeline.

This PR now includes the version upgrade of `io.grpc:*` to **1.64.0** (which is the latest version). This is necessary as described in my comment above. It allows us to use the `io.grpc` Jars directly as modules.

I apologies in advance if I still should have missed something. Please take another look. Maybe we can finally get this integrated.

thenswan (approved):
LGTM 🎉

agadzhalov (approved):
LGTM, tyvm @jjohannes and @thenswan 🙏🚀

0xivanov (approved):

gradle/plugins/src/main/kotlin/com.hedera.gradle.java.gradle.kts
```diff
@@ -0,0 +1,170 @@
...
+    reportsDir = layout.buildDirectory.dir("reports/spotbugs")
+    //includeFilter = file('spotbugs-include.xml')
+    //excludeFilter = file('spotbugs-exclude.xml')
+    onlyAnalyze = listOf("com.hedera.hashgraph.sdk.*")
+    //projectName = name
+    //release = version
+    //extraArgs = [ '-nested:false' ]
+    //jvmArgs = [ '-Duser.language=ja' ]
+    //maxHeapSize = '512m'
+}
```
rbarkerSL, line 148: Why are we leaving all these commented out things in here? Just shortcuts for various debugging scenarios?

rbarkerSL (approved):
Looks good to me, just one minor comment on some comments.

gradle/plugins/src/main/kotlin/com.hedera.gradle.java.gradle.kts
thenswan: @rbarkerSL this is legacy things as of now. Most probably we will remove spotbugs completely, but we will review it as a follow up, because this thing is huge already:)

PR #1786, opened 2024-04-22
feat: method to calculate the recoveryId for ECDSA signatures

### LINKED ISSUE #1784: Add recovery id calculation when signing ethereum call transactions
Currently there is no way to calculate recovery id (v) when we sign ethereum transactions from java SDK.

The recovery ID is an index from 0 to 3, which indicates which of the 4 possible public keys is the correct one. Because the key recovery operation yields multiple potential public keys, the correct key must either be stored alongside the signature or you must be willing to try each recovery ID in turn until you find one that outputs the key you are expecting.

Recovery operation is described in [SEC1v2, section 4.1.6](https://www.secg.org/sec1-v2.pdf).

JivkoKelchev:
**Description**:

This PR adds a method to calculate the `recoveryId` for ECDSA signatures. 
The correct `recoveryId` value is needed on the consensus node side in order to deterministically recover the signer's public key and corresponding EVM address. 
To calculate recovery, we try to recover the public key with all possible `recoveryId` values (For secp256k1 cureve - 0 and 1). After that, we compare public key candidates with the expected one.
This algorithm is described in [SEC1v2 section 4.1.6.](https://www.secg.org/sec1-v2.pdf)

**Related issue(s)**:

Fixes #1784

**Notes for reviewer**:

**Checklist**


0xivanov (commented):
LGTM, there are some sonar issues thought

vtronkov (dismissed):
Awesome job, Jivko! It's a really hard bug to find and fix 💯 Thank you!

agadzhalov (approved):
LGTM

PR #1827, opened 2024-05-27
chore: deprecate TokenRelationship.symbol field due to lack of support by consensus nodes and mirror node

### LINKED ISSUE #1821: Deprecate `TokenRelationship.symbol`
After updating queries in #1698, SDK clients lost `symbol` field data for `AccountInfoQuery` and `ContractInfoQuery` as this field was not added to mirror node REST API. So since the mirror node does not return value for the `symbol` of the token, suggest marking it as deprecated [in the `TokenRelationship` class](https://github.com/hashgraph/hedera-sdk-java/blob/4301900b75fdc96413f2b35fd6833b64de1a5dfb/sdk/src/main/java/com/hedera/hashgraph/sdk/TokenRelationship.java#L42).

thenswan:
This PR deprecates `TokenRelationship.symbol` field due to lack of support by consensus nodes and mirror node.

Closes #1821

### DIFF
--- a/sdk/src/main/java/com/hedera/hashgraph/sdk/TokenRelationship.java
+++ b/sdk/src/main/java/com/hedera/hashgraph/sdk/TokenRelationship.java
@@ -37,6 +37,10 @@ public class TokenRelationship {
     public final TokenId tokenId;
     /**
      * The Symbol of the token
+     *
+     * @deprecated Not supported by consensus nodes (from hedera-services tag v0.50.x).
+     * Although the Mirror Node REST APIs still contain this feature, there is no straightforward way of integration,
+     * leading to this field being deprecated.
      */
     public final String symbol;
     /**


agadzhalov (dismissed):

sdk/src/main/java/com/hedera/hashgraph/sdk/TokenRelationship.java
```diff
@@ -37,6 +37,10 @@ public class TokenRelationship {
     public final TokenId tokenId;
     /**
      * The Symbol of the token
+     *
+     * @deprecated Not supported by consensus nodes (from hedera-services tag v0.50.x).
+     * Although the Mirror Node REST APIs still contain this feature, there is no straightforward way of integration,
+     * leading to this field being deprecated.
```
0xivanov, line 43: Can be marked with `@Deprecated` Annotation. Also can add in the javadoc that the symbol can be extracted from `TokenInfo`

0xivanov (approved):
LGTM

PR #1842, opened 2024-06-10
feat: custom derivation paths in Mnemonic ECDSA private key derivation

bguiz:
**Description**:

## What

- Add custom derivation path capability

## Why

- Compatibility/ parity with EVM dev tools + wallets. Specifically:
  - When working Metamask, you need to use the hardcoded derivation path in it, and in SDK there's also a (different) hardcoded derivation path.
  - In EVM dev tools - web3.js/ethers.js/viem etc - they allow you to specify a custom derivation path, and do not force you to use the Ethereum one.
- The lack of support for custom derivation paths is a friction point for EVM developers onboarding into Hedera development
  - This may have the effect of effectively discouraging use of the SDK, in favour of using 3rd party libraries that do support custom derivation paths
- we should aim to reduce the reliance on 3rd party dependencies in general, within code that is written using the SDK
  - The following examples demonstrate that (1) additional code is needed, and (2) additional 3rd party dependencies need to be brought in:
  - example of extra code required in JS: https://github.com/hedera-dev/scratch-hello-world/blob/60fa4007/js-hello-world/00-create-fund-account/script-create-fund-account.js#L19-L23
  - example of extra code required in Java: https://github.com/hedera-dev/scratch-hello-world/blob/60fa4007/java-hello-world/00-create-fund-account/src/main/java/createfundaccount/ScriptCreateFundAccount.java#L30-L55
  - example of extra code required in Go: https://github.com/hedera-dev/scratch-hello-world/blob/60fa4007/golang-hello-world/00-create-fund-account/script-create-fund-account.go#L37-L61

**Related issue(s)**:

Fixes: https://github.com/hashgraph/hedera-sdk-java/issues/1263#issuecomment-1330953933

**Notes for reviewer**:

- Related: https://github.com/hedera-dev/scratch-hello-world/pull/1
- Discussion with @thenswan + @0xivanov: https://swirldslabs.slack.com/archives/C0453UVBR42/p1717898573711519?thread_ts=1717061977.081859&cid=C0453UVBR42
- Related: https://github.com/hashgraph/hedera-sdk-go/pull/958
- Related: https://github.com/hashgraph/hedera-sdk-js/pull/2341

**Checklist**


sdk/src/main/java/com/hedera/hashgraph/sdk/Mnemonic.java
```diff
@@ -557,28 +559,101 @@ public PrivateKey toStandardEd25519PrivateKey(String passphrase, int index) {
...
+     * as per BIP-44:
+     * `m / purpose' / coin_type' / account' / change / address_index`
+     * Takes into account `'` for hardening as per BIP-32,
+     * and does not prescribe which components should be hardened.
+     *
+     * @param derivationPath  the derivation path in BIP-44 format,
+     *                        e.g. "m/44'/60'/0'/0/0"
+     * @return an array of integers designed to be used with PrivateKey#derive
+     */
+    private int[] calculateDerivationPathValues(String derivationPath) {
```
0xivanov, line 574: Could benefit from additional validation and exception handling to cover invalid input formats
 
```java
    private int[] calculateDerivationPathValues(String derivationPath) throws IllegalArgumentException {
        if (derivationPath == null || derivationPath.isEmpty()) {
            throw new IllegalArgumentException("Derivation path cannot be null or empty");
        }

        Pattern pattern = Pattern.compile("m/(\\d+'?)/(\\d+'?)/(\\d+'?)/(\\d+'?)/(\\d+'?)");
        Matcher matcher = pattern.matcher(derivationPath);

        if (!matcher.matches()) {
            throw new IllegalArgumentException("Invalid derivation path format. Expected format: m/44'/60'/0'/0/0");
        }

        int[] numbers = new int[5];
        boolean[] isHardened = new boolean[5];

        try {
            // Extract numbers and determine if hardened
            for (int i = 1; i <= 5; i++) {
                String value = matcher.group(i);
                if (value.endsWith("'")) {
                    isHardened[i - 1] = true;
                    value = value.substring(0, value.length() - 1);
                } else {
                    isHardened[i - 1] = false;
                }
                numbers[i - 1] = Integer.parseInt(value);
            }
        } catch (NumberFormatException e) {
            throw new IllegalArgumentException("Invalid number format in derivation path", e);
        }

        int[] values = new int[5];
        for (int i = 0; i < numbers.length; i++) {
            values[i] = (isHardened[i] ? Bip32Utils.toHardenedIndex(numbers[i]) : numbers[i]);
        }

        return values;
    }

```

bguiz: thanks for the suggestion - will look into this!

bguiz: @0xivanov I have implemented this in [8f7f0e6](https://github.com/hashgraph/hedera-sdk-java/pull/1842/commits/8f7f0e6d8ec19a13ae08b3edc15d941ae3378905)

Note that I'm not sure if the `NumberFormatException` scenario can occur, thanks to the Regex that would catch it upon executing `matcher.matches()`.
Do you prefer if I leave that check in anyway?
Or should I remove it from the code?
Or is there something I'm missing (something that could slip past the Regex pattern match test)?

0xivanov: You are right that `NumberFormatException` is unlikely to occur, however, for defensive programming and to ensure robustness, it's still a good practice to handle any unexpected exceptions that might arise.

0xivanov (commented):
LGTM, would like @thenswan to also have a look

sdk/src/main/java/com/hedera/hashgraph/sdk/Mnemonic.java
bguiz: Ok got it.
I will leave it in then.
But I won't have test coverage for that `catch` - I assume that this is OK.

0xivanov (approved):

bguiz:
❓ Has this made it into a release yet?

thenswan:
Hey @bguiz, yes, you can use it in stable v2.35.1 👍🏻

bguiz:
Thanks! 🙌 
> Hey @bguiz, yes, you can use it in stable v2.35.1 👍🏻

PR #1871, opened 2024-06-18
fix: properly handle `PLATFORM_NOT_ACTIVE` status code

thenswan:
**Description**:
This PR improves handling of `PLATFORM_NOT_ACTIVE` Status for `getReceipt()` and `getRecord()`.

### DIFF
--- a/sdk/src/main/java/com/hedera/hashgraph/sdk/TransactionReceiptQuery.java
+++ b/sdk/src/main/java/com/hedera/hashgraph/sdk/TransactionReceiptQuery.java
@@ -194,18 +194,17 @@ public final class TransactionReceiptQuery
     @Override
     ExecutionState getExecutionState(Status status, Response response) {
         switch (status) {
-//            case BUSY:
             case UNKNOWN:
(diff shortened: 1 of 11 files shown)


sdk/src/main/java/com/hedera/hashgraph/sdk/TransactionReceiptQuery.java
```diff
@@ -204,7 +204,8 @@ ExecutionState getExecutionState(Status status, Response response) {
                 break;
 
             default:
-                return ExecutionState.REQUEST_ERROR;
+//                return ExecutionState.REQUEST_ERROR;
+                return super.getExecutionState(status, response);
```
0xivanov, line 208: We have the same logic in `TransactionRecordQuery.java`, shall we change it there too to handle `PlatformNotActive`

thenswan: I think yes, we just need to sync to discuss the approach

0xivanov (approved):
LGTM

PR #1901, opened 2024-07-16
feat: update security parameters when updating the addressbook

### LINKED ISSUE #1900: I would like the client to update security parameters when updating the addressbook so that updates will affect TLS
### Problem

I'd like to test using TLS with a network other than {mainnet, testnet, previewnet}.

### Solution

- Have `Client::setNetworkFromAddressBook` update the `Network::addressBook` and `Node::addressBookEntry` which contain the security parameters.

### Alternatives

_No response_

matteriben:
**Description**:

Update security parameters when updating the addressbook

Fixes #1900


thenswan:
Hey @matteriben, thank you for your contribution here once again 🙂 
Can you please explain use-case(s) for this feature more in detail? It would be great if you provide some examples of how you are planning to use it.

matteriben:
> Hey @matteriben, thank you for your contribution here once again 🙂 Can you please explain use-case(s) for this feature more in detail? It would be great if you provide some examples of how you are planning to use it.

I'd like to test using TLS with a network other than {mainnet, testnet, previewnet}. To do so, I need a way to provide the `certHash` in addition to the address string and `accountId` when configuring the client for my network.

matteriben:
@thenswan This change may be problematic as it stands currently since it overwrites `certHash` (needed for TLS), and it appears that hedera-service may not provide `certHash` meaning with this change updating the client with the current addressbook could effectively remove `certHash` (by overwriting with nothing).

matteriben:
@thenswan I updated this PR and so that the old `certHash` will be used in case the new addressbook is missing `certHash`, which currently appears to be the case when retrieving the addressbook from hedera-service directly.

I also updated the test to verify this case.

matteriben:
> @matteriben thank you for provided description and example. I reviewed everything and have a question: Can you please take a look at [this example](https://github.com/hashgraph/hedera-sdk-java/pull/1570) that I made some time ago and let me know if this will suit your needs (or how your use-case is different from that)?

@thenswan issue #1569 is related, but not the same issue. In issue #1569 the primary concern is configure the client to use a specific node which is already exists in the predefined address book. Having the ability to load a custom address book (including security parameters) would solve both issue #1569 and issue #1900.

PR #1570 demonstrates how to load a predefined address book then narrow the network to a single node. This is sufficient for issue #1569 but not #1900 because issue #1569 is dealing with nodes that exist in one of the predefined networks, where issue #1900 is dealing with a custom network that is not predefined.

sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
```diff
@@ -416,6 +416,7 @@ public synchronized Client setNetworkFromAddressBook(NodeAddressBook addressBook
             addressBook.nodeAddresses,
             isTransportSecurity() ? PORT_NODE_TLS : PORT_NODE_PLAIN
         ));
```
matteriben, line 418: Note: `Client::setNetworkFromAddressBook` is called from `Client::scheduleNetworkUpdate` for periodic background addressbook refresh. `Client::setNetworkFromAddressBook` currently convert the addressbook to `Map<String, AccountId>` (where string is endpoint.toString()) and passes that to `Network::setNetwork`. In other words, current behavior updates only the endpoints from the latest addressbook, not the security parameters.

If the line below which is added by this PR is only called conditionally, with the condition set to `false` during refresh, the existing behavior will be preserved. It seems to me using the latest security parameters makes sense, however I understand that there is reluctance to change the existing behavior of this code.

0xivanov: I agree, it might look something like this, keeping the original interface and adding the flag:

```java
    public synchronized Client setNetworkFromAddressBook(NodeAddressBook addressBook, boolean updateAddressBook)
        throws InterruptedException, TimeoutException {
        network.setNetwork(Network.addressBookToNetwork(
            addressBook.nodeAddresses,
            isTransportSecurity() ? PORT_NODE_TLS : PORT_NODE_PLAIN
        ));
        if (updateAddressBook) {
            network.setAddressBook(addressBook);
        }
        return this;
    }

    /**
     * Replace all nodes in this Client with the nodes in the Address Book
     *
     * @param addressBook A list of nodes and their metadata
     * @return {@code this}
     * @throws InterruptedException when a thread is interrupted while it's waiting, sleeping, or otherwise occupied
     * @throws TimeoutException     when shutting down nodes
     */
    public synchronized Client setNetworkFromAddressBook(NodeAddressBook addressBook)
        throws InterruptedException, TimeoutException {
        return setNetworkFromAddressBook(addressBook, false);
    }
```

sdk/src/main/java/com/hedera/hashgraph/sdk/Network.java
```diff
@@ -158,6 +160,44 @@ private Network setLedgerIdInternal(@Nullable LedgerId ledgerId, @Nullable Map<A
...
     }
 
+    void setAddressBook(NodeAddressBook addressBook) {
+        Map<AccountId, NodeAddress> newAddressBook = addressBook.getNodeAddresses().stream()
+            .filter(nodeAddress -> Objects.nonNull(nodeAddress.getAccountId()))
+            .collect(Collectors.toMap(NodeAddress::getAccountId, Function.identity(),
+                /*
+                * Here we index by AccountId ignoring any subsequent entries with the same AccountId.
+                *
+                * Currently, this seems to be needed when reloading predefined address book for testnet which contains
```
0xivanov, line 170: It's weird that there are multiple entries with the same AccountId, can you give more information why this is happening if you found something during your testing?

0xivanov: What do you think about simplifying this logic:

```java
 void setAddressBook(NodeAddressBook addressBook) {
        var newNodeAddresses = addressBook.getNodeAddresses();

        if (this.addressBook != null) {
            for (NodeAddress newNodeAddress : newNodeAddresses) {
                NodeAddress previousNodeAddress = this.addressBook.get(newNodeAddress.getAccountId());

                if (previousNodeAddress != null) {
                    ByteString certHash = newNodeAddress.getCertHash();
                    if (certHash == null || certHash.isEmpty()) {
                        newNodeAddress.setCertHash(previousNodeAddress.getCertHash());
                    }
                }
                this.addressBook.put(newNodeAddress.getAccountId(), newNodeAddress);
            }
        } else {
            this.addressBook = new HashMap<>();
            for (NodeAddress newNodeAddress : newNodeAddresses) {
                this.addressBook.put(newNodeAddress.getAccountId(), newNodeAddress);
            }
        }

        for (var node : nodes) {
            node.setAddressBookEntry(this.addressBook.get(node.getAccountId()));
        }
    }
```

Would this work in all cases (unit test is passing)?

matteriben: > It's weird that there are multiple entries with the same AccountId, can you give more information why this is happening if you found something during your testing?

@0xivanov I believe I saw this in the predefined address book for testnet, I'm not sure why.

matteriben: > What do you think about simplifying this logic like this:

It looks pretty similar to what I had, the only real significant difference I noticed just visually comparing is that this version doesn't seem to remove entries in the addressbook in the case when the account-id is missing from the latest addressbook.

By the way, as far as I can tell aside from this method which updates `Network::addressbook` then loops through it to update nodes, `Network::addressbook` appears to be unused, except for two tests. You might be able to remove `Network::addressbook` in that case, if we still want to try and preserve old TLS parameters, you can check the address-book-entry in the node before overwriting it.

0xivanov: > the only real significant difference I noticed just visually comparing is that this version doesn't seem to remove entries in the addressbook in the case when the account-id is missing from the latest addressbook.

What piece of logic are you referring to?

matteriben: > What piece of logic are you referring to?

This version only doesn't seem to remove entries from `this.addressBook` in the case when the account-id is missing from the latest `addressBook`.

```
 void setAddressBook(NodeAddressBook addressBook) {
        var newNodeAddresses = addressBook.getNodeAddresses();

        if (this.addressBook != null) {
            for (NodeAddress newNodeAddress : newNodeAddresses) {
                // ...
                this.addressBook.put(newNodeAddress.getAccountId(), newNodeAddress);
            }
        }
        // ...
    }
```

Compared with this version:

```
    void setAddressBook(NodeAddressBook addressBook) {
        Map<AccountId, NodeAddress> newAddressBook = addressBook.getNodeAddresses().stream()
            .filter(nodeAddress -> Objects.nonNull(nodeAddress.getAccountId()))
            .collect(Collectors.toMap(NodeAddress::getAccountId, Function.identity(),
                (a, b) -> a));
        // ...
        this.addressBook = newAddressBook;
    }
```

0xivanov: I see what you mean,
I wrote a unit test for this flow and even `network.setNetwork()` does not handle the scenario when the accountID is null. It populates the nodes list with this invalid accountID and the test fails.
I am not sure if it is even possible to have a node without account id and if it is necessary to handle this case.

0xivanov: Ok, will keep your implementation.

0xivanov (approved):
LGTM

agadzhalov (approved):
LGTM

PR #1905, opened 2024-07-18
fix: allow certHash to contain upper or lower case characters

### LINKED ISSUE #1904: certHash doesn't work with uppercase letters
### Description

`certHash` is a hex string and should be compared in a case insensitive manor.

### Steps to reproduce

1. Try using a `certHash` that contains uppercase letters and TLS won't work.

### Additional context

_No response_

### Hedera network

_No response_

### Version

v2.35.1

### Operating system

None

matteriben:
**Description**:
Allow certHash to contain upper or lower case characters

**Related issue(s)**:

Fixes #1904

### DIFF
--- a/sdk/src/main/java/com/hedera/hashgraph/sdk/HederaTrustManager.java
+++ b/sdk/src/main/java/com/hedera/hashgraph/sdk/HederaTrustManager.java
@@ -108,7 +108,7 @@ class HederaTrustManager implements X509TrustManager {
                 throw new IllegalStateException("Failed to find SHA-384 digest for certificate hashing", e);
             }
 
-            if (this.certHash.equals(Hex.toHexString(certHashBytes))) {
+            if (this.certHash.equalsIgnoreCase(Hex.toHexString(certHashBytes))) {
                 return;
             }
         }


sdk/src/main/java/com/hedera/hashgraph/sdk/HederaTrustManager.java
```diff
@@ -108,7 +108,7 @@ public void checkServerTrusted(X509Certificate[] chain, String authType) throws
                 throw new IllegalStateException("Failed to find SHA-384 digest for certificate hashing", e);
             }
 
-            if (this.certHash.equals(Hex.toHexString(certHashBytes))) {
+            if (this.certHash.equalsIgnoreCase(Hex.toHexString(certHashBytes))) {
```
0xivanov, line 111: The hashes should be identical, I don't see the point in comparing them case-insensitive.

matteriben: Hexadecimal is not case sensitive.

agadzhalov:
@matteriben can you share more details what is the purpose of this change? Is ti a chore or is ti actually a fix?

sdk/src/main/java/com/hedera/hashgraph/sdk/HederaTrustManager.java
0xivanov: I see, makes sense. Though I am still cautious about changing this part of the logic, because
1. It's quite old and I do not have the whole context for it
2. It seems to be involving the security of the sdk
3. I checked this piece of logic in other sdks and there it's the same as the current implementation

matteriben:
> @matteriben can you share more details what is the purpose of this change? Is ti a chore or is ti actually a fix?

I think its more correct to ignore the case when comparing hex strings, but it's not causing any problem for me.

0xivanov:
Closing since it's not mandatory change.

PR #1930, opened 2024-08-01
feat: support ports that are not well known for use with port forwarding

### LINKED ISSUE #1929: support ports that are not well known for use with port forwarding
### Problem

I'd like to connect using port-forwarding which means for example addresses such as:
```
node1 -> localhost:10001
node2 -> localhost:10002
node3 -> localhost:10003
```

### Solution

- Don't ignore ports that are not well known.
- Support TLS with ports that are not well known by tracking state of transport security inside node address.

### Alternatives

_No response_

matteriben:
**Description**:
Support ports that are not well known for use with port forwarding.

**Related issue(s)**:

Fixes #1929


sdk/src/main/java/com/hedera/hashgraph/sdk/Network.java
```diff
@@ -207,9 +211,14 @@ private static Map<AccountId, NodeAddress> getAddressBookForLedger(@Nullable Led
...
         var network = new HashMap<String, AccountId>();
+        boolean wellKnownPorts = addressBook.stream()
+            .map(NodeAddress::getAddresses)
+            .flatMap(Collection::stream)
+            .map(Endpoint::getPort)
+            .allMatch(Set.of(PORT_NODE_TLS, PORT_NODE_PLAIN)::contains);
         for (var nodeAddress : addressBook) {
             for (var endpoint : nodeAddress.addresses) {
-                if (endpoint.port == desiredPort) {
+                if (!wellKnownPorts || endpoint.port == desiredPort) {
```
0xivanov, line 221: I guess we could remove `int desiredPort` and remove this if statement as a whole. WDYT?

matteriben: It's ok with me, but it changes the behavior of things. Currently this API only allows a single port through, if you remove desiredPort it will allow any and all ports through.

Also, this change should be made in conjunction with the change to BaseNodeAddress. Because this change will allow non-standard port numbers to get through, but without the corresponding change in BaseNodeAddress requests to use transport security on non-standard ports will be silently ignored.

0xivanov: `desiredPort` right now can be either 50212 or 50211 and given this information, the current filtering does not make sense for me. 
If we want to allow not well known ports, it makes sense for me to allow any port.
At the end of the day, the user can set whatever network they want given the addition of dynamic address book feature.

0xivanov (approved):
LGTM

PR #1932, opened 2024-08-02
fix: examples

### LINKED ISSUE #1630: Fix some Java examples
- [x] `ZeroTokenOperationsExample` - doesn’t work
- [x] `SolidityPrecompileExample` - doesn’t work with `hedera-local-node`
- [x] `ConsensusPubSubChunkedExample` - doesn’t work on CI

thenswan:
**Description**:
Fix examples and enable them in CI:
- `ZeroTokenOperationsExample`
- `SolidityPrecompileExample`
- `ConsensusPubSubChunkedExample`

Closes #1630


examples/src/main/java/com/hedera/hashgraph/sdk/examples/ConsensusPubSubChunkedExample.java
```diff
@@ -122,7 +113,7 @@ public static void main(String[] args) throws TimeoutException, PrecheckStatusEx
         // get the receipt to ensure there were no errors
         transaction.execute(client).getReceipt(client);
 
-        boolean largeMessageReceived = largeMessageLatch.await(60, TimeUnit.SECONDS);
+        boolean largeMessageReceived = largeMessageLatch.await(180, TimeUnit.SECONDS);
```
0xivanov, line 116: This example can be flaky on localhost, i suggest running it at least 10 times to make sure it works

examples/src/main/java/com/hedera/hashgraph/sdk/examples/ZeroTokenOperationsExample.java
```diff
@@ -77,14 +93,14 @@ public static void main(String[] args) throws Exception {
...
         // step 5 wipe the token by passing a zero value
-
         contractHelper.executeSteps(/* from step */ 0, /* to step */ 5, client);
 
         // step 6 use SDK and transfer passing a zero value
         // Create Fungible Token
         System.out.println("Attempting to execute step 6");
 
         TokenCreateTransaction tokenCreateTransaction = new TokenCreateTransaction()
+            .setMaxTransactionFee(new Hbar(50))
```
0xivanov, line 103: Don't we already set the max txn fee in ClientHelper

examples/src/main/java/com/hedera/hashgraph/sdk/examples/ZeroTokenOperationsExample.java
```diff
@@ -66,9 +66,25 @@ public static void main(String[] args) throws Exception {
...
+            .setKey(KeyList.of(alicePublicKey, contractHelper.contractId).setThreshold(1))
+            .freezeWith(client)
+            .sign(alicePrivateKey)
+            .execute(client)
+            .getReceipt(client);
+
         // Configure steps in ContractHelper
         contractHelper
-            .setPayableAmountForStep(0, Hbar.from(20))
+            .setPayableAmountForStep(0, Hbar.from(40)) // is 40 Hbar needed?
```
0xivanov, line 87: 40 seems a lot but if it's for all the steps it might be ok

examples/src/main/java/com/hedera/hashgraph/sdk/examples/ConsensusPubSubChunkedExample.java
thenswan: I agree, I have following observations:
- when I start local-node it doesn't work
- when I re-run example 10x times, it worked each time

cc @agadzhalov @isavov wdyt?

isavov (dismissed):

examples/src/main/java/com/hedera/hashgraph/sdk/examples/ConsensusPubSubChunkedExample.java
0xivanov: I think it's best to skip this test, as it might not even be designed to work on localhost.

0xivanov (commented):
Are the contracts getting compiled runtime? In Go we have hardcoded the bytecode, wondering if in Java there is a reason to update the `.sol` files.

thenswan:
> Are the contracts getting compiled runtime?

No, also hardcoded.


> wondering if in Java there is a reason to update the .sol files.

I think its to show updates in bytecode. And for people to understand whats going on in the contract itself :) 

@0xivanov

gradle/plugins/src/main/kotlin/com.hedera.gradle.examples.java.gradle.kts
```diff
@@ -77,9 +77,6 @@ abstract class RunAllExample : DefaultTask() {
             .asSequence()
             .map { it.name.replace(".java", "") }
             .filter { it != "ValidateChecksumExample" } // disabled this example, because it needs user input (but it WORKS)
-            .filter { it != "ZeroTokenOperationsExample" } // doesn't work
-            .filter { it != "SolidityPrecompileExample" } // doesn't work with hedera-local-node
-            .filter { it != "ConsensusPubSubChunkedExample" } // doesn't work on CI
```
0xivanov, line 82: should revert the pub sub i think

thenswan:
**The problem**
`ConsensusPubSubChunkedExample` in Java SDK is not working sometimes when running it against the local-node.

**How to reproduce**
Pre-requisites:
1. Clone Java SDK repo --> https://github.com/hashgraph/hedera-sdk-java
2. Configure dotenv --> https://github.com/hashgraph/hedera-sdk-java/blob/1c6325659557874479f3387b2f0f2028bd1870a1/examples/README.md#configuration

Steps to reproduce:
1. If Docker is running, then reopen the Docker (I mean exit the docker and run it again).
2. Start the local-node.
3. Run this command from the root of a project: `./gradlew :examples:runConsensusPubSubChunked`
Expected result:
ConsensusPubSubChunkedExample should fail the first time and then it should work any other time.

gradle/plugins/src/main/kotlin/com.hedera.gradle.examples.java.gradle.kts
0xivanov: @thenswan Do you want to skip this example for now and merge this PR. After it's fixed in local-node and has consistent behaviour we can enable it.

0xivanov (approved):
LGTM

PR #1970, opened 2024-08-27
feat: Simplified ipv4 address

### LINKED ISSUE #1969: Investigate non-standart ipv4 addresses
### Problem

IPv4 addresses would typically be represented as four octets (bytes), no two parts (network and host). While it's possible to divide an IPv4 address into network and host parts, the representation here is non-standard.

### Solution

-

### Alternatives

_No response_

0xivanov:
**Description**:
IPv4 addresses would typically be represented as four octets (bytes), no two parts (network and host). While it's possible to divide an IPv4 address into network and host parts, the representation here is non-standard.

**Related issue(s)**:

Fixes #1969

**Notes for reviewer**:

**Checklist**


agadzhalov (approved):
LGTM

Issue #1844, opened 2024-06-10
Too many threads when multiple Clients are used in a single process


OlegMazurov:
### Problem

A single Client can only provide limited throughput mostly due to this setting in network nodes:
```
netty.prod.maxConcurrentCalls = 10
```
Load tests require creating many Clients (10-150) in order to achieve maximum network throughput.
Unfortunately, each client creates its own thread pool with the number of threads equals to the number of CPUs on the machine. Running such test with 150 Clients on a 48-CPU machine creates 150*48 = 7200 threads, for which there is little justification.
I think sharing a single thread pool between multiple Clients should be acceptable.

### Solution

- share a static thread pool between multiple `Clients`
- provide a way to configure the size of the pool (`Runtime.getRuntime().availableProcessors()` by default)

0xivanov:
Hi @OlegMazurov, thanks for submitting this issue.
Do you think exposing an API for `Client` , where you can set the `ExecutorService` as a parameter would be effective for you?
This way you would have the ability to configure singleton thread pool for multiple clients and also configure the size of the pol

0xivanov:
> A single Client can only provide limited throughput mostly due to this setting in network nodes

Does this mean that having more than 10 threads in the pool is meaningless?

OlegMazurov:
> Do you think exposing an API for Client , where you can set the ExecutorService as a parameter would be effective for you?

Yes, it would work. However, I think the default behavior should also change: a singleton thread pool for multiple clients. If somebody wants a separate thread pool per client, they can use the API.

> > A single Client can only provide limited throughput mostly due to this setting in network nodes
> 
> Does this mean that having more than 10 threads in the pool is meaningless?

No, I use asynchronous calls so responses may come in bulk and require more parallelism to process. Moreover, the `maxConcurrentCalls` parameter is hidden (not exposed by the network) and can be set arbitrarily. The client should not make any assumptions about it.

0xivanov:
I would not change the default behaviour because no one has complained about this at this point. There are more complexities that would come with this change, like closing the client/network/executor, since now they are closing simultaneously. Maybe some time in the future.

You can see the newly provided apis [here](https://github.com/hashgraph/hedera-sdk-java/pull/1987)

0xivanov:
fixed in https://github.com/hashgraph/hedera-sdk-java/pull/1987

PR #1987, opened 2024-09-05
feat: Create `Client` with external `ExecutorService`

### LINKED ISSUE #1986: Status `Busy` is treated as `SERVER_ERROR`
### Description

Status `Busy` is treated as `SERVER_ERROR` but should be as `RETRY`

### Steps to reproduce

-

### Additional context

_No response_

### Hedera network

_No response_

### Version

lateset

### Operating system

None

0xivanov:
**Description**:

 - Adds new methods for creating `Client` with predefined `ExecutorService`.
 - Adds new method `closeChannels` to allow closing the channels without closing the executor.
 - Fixes handling status `BUSY`.

**Related issue(s)**:

Fixes #1986 #1844

**Notes for reviewer**:

**Checklist**


sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
```diff
@@ -345,6 +345,7 @@ private void delay(long delay) {
         }
         try {
             if (delay > 0) {
+                logger.debug("Sleeping for: " + delay + " | Thread name: " + Thread.currentThread().getName());
```
OlegMazurov, line 348: Use
```suggestion
                if (logger.isEnabledForLevel(LogLevel.DEBUG)) {
                    logger.debug("Sleeping for: " + delay + " | Thread name: " + Thread.currentThread().getName());
                }
```
to avoid unnecessary argument evaluation when `DEBUG` is not enabled.

sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
```diff
@@ -144,6 +144,27 @@ static ExecutorService createExecutor() {
...
+     * same Hedera network. Failure to do so will result in undefined behavior.
+     * The client will load balance all requests to Hedera using a simple round-robin scheme to
+     * chose nodes to send transactions to. For one transaction, at most 1/3 of the nodes will be tried.
+     *
+     * @param networkMap the map of node IDs to node addresses that make up the network.
+     * @param executor runs the grpc requests asynchronously. Note that calling `close()` method on one of the
+     *                 clients will close the executor for all the other clients sharing this executor
+     * @return {@link com.hedera.hashgraph.sdk.Client}
+     */
+    public static Client forNetworkWithExecutor(Map<String, AccountId> networkMap, ExecutorService executor) {
```
OlegMazurov, line 160: Changing name seems redundant. Why not overload the existing method?
```suggestion
    public static Client forNetwork(Map<String, AccountId> networkMap, ExecutorService executor) {
```
Ditto for other methods below.

0xivanov: Done

sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
0xivanov: `isEnabledForLevel` gets called in `logger.debug` so I don't know if I need to do it here. I saw it being used in a couple more places so I'll change it anyway.

agadzhalov (approved):
LGTM, ty @0xivanov

Issue #1794, opened 2024-04-30
Improvements of and documentation of the Gradle build


jjohannes:
This issue is a follow-up for the work done here: https://github.com/hashgraph/hedera-sdk-java/pull/1732
Each task should be converted to an issue and addressed separately. Tasks are aligned in a priority order.

```[tasklist]
### Tasks
- [ ] https://github.com/hashgraph/hedera-sdk-java/issues/1885
- [ ] https://github.com/hashgraph/hedera-sdk-java/issues/1898
- [ ] https://github.com/hashgraph/hedera-sdk-java/issues/1894
- [x] Introduce Guava fix, which should be in plugin in the future
- [x] Get rid of `mergeJar("com.google.code.findbugs:jsr305")`
- [x] Adjust the code style rules and tools to what its done in _hedera-services_ (e.g. remove sonar, introduce spotless, decide if we want to keep error prone etc.)
- [x] Revisit gradle setup in `tck` project
- [ ] https://github.com/hashgraph/hedera-sdk-java/issues/1896
- [x] Add more docs for the Gradle build (`.kts` files, code comments)
- [x] Resolve duplication in constraints publishing (right now there is some duplication)
- [x] Add PR Check for the Android Example (that it builds)
- [x] Add more docs for the Gradle build (`.md` files, in particular for adding/modifying dependencies)
- [x] Review what SDK dependencies should be transitive: https://github.com/hashgraph/hedera-sdk-java/pull/1732#discussion_r1603375896 --> moved to https://github.com/hashgraph/hedera-sdk-java/issues/2180
- [ ] Revisit `Taskfile.yml` (maybe make it as gradle tasks)
```

thenswan:
>I added the constraints publishing. Right now there is some duplication, but I am not sure at the moment how to best centralize this version. But as we use dependabot, we can maybe also rely on it updating all of them. I would like to think about that again as follow up so that we can get this PR finally in.

thenswan:
Guava fix, which should be in plugin in the future
https://github.com/gradlex-org/extra-java-module-info
Issue: https://github.com/gradlex-org/extra-java-module-info/issues/129

thenswan:
Reviewing what dependencies should be transitive: https://github.com/hashgraph/hedera-sdk-java/pull/1732#discussion_r1603375896

thenswan:
Add building the Android example to the GH Actions pipeline.

thenswan:
[...]

PR #2016, opened 2024-09-30
build: remove inconsistent code analysis / align with hedera-services

jjohannes:
**Description**:
Addresses the following items in #1794 

- Introduce Guava fix, which should be in plugin in the future
- Get rid of `mergeJar("com.google.code.findbugs:jsr305")`
- _(Partially)_ Adjust the code style rules and tools to what its done in _hedera-services_

This removes the setups for: sonar, error prone, spotbugs. Justification:
- Spotbugs does not run as part of the pipeline, but if you run it it gives a ton of errors
- Instead of sonar, we should use **codacy** as we do in `hedera-servcies` once we have the same Gradle setup
- Error prone did not really run, because it is an annotation processor that produces warnings but all warnings were turned off with this comment (??):

``` 
    // Disable warnings because I'm tired of them :P
    options.isWarnings = false
```

As we also do not use (and configure) these tools in _hedera-services_, my suggestion is to remove them for now. We can  re-introduce one of these tools later again, but then with proper configuration that checks what you want to check and that works with errors (and not warnings).

Instead, the `javac` linting capabilities are now configured similar to how it is done in _hedera-services_. As preparation for sharing this configuration in the next step (https://github.com/hashgraph/hedera-services/issues/14255).

**Related issue(s)**:

#1794 / https://github.com/hashgraph/hedera-services/issues/14255

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/gradle/plugins/src/main/kotlin/com.hedera.gradle.java-base.gradle.kts
+++ b/gradle/plugins/src/main/kotlin/com.hedera.gradle.java-base.gradle.kts
@@ -1,5 +1,3 @@
-import net.ltgt.gradle.errorprone.errorprone
-
 /*-
  *
  * Hedera Java SDK
@@ -23,7 +21,6 @@ import net.ltgt.gradle.errorprone.errorprone
 plugins {
     id("java")
     id("jacoco")
-    id("net.ltgt.errorprone")
     id("org.gradlex.java-module-dependencies")
     id("com.hedera.gradle.base")
     id("com.hedera.gradle.repositories")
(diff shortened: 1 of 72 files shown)


sdk/src/test/java/com/hedera/hashgraph/sdk/ContractFunctionParametersTest.java
```diff
@@ -563,9 +563,9 @@ void intSizesEncodeCorrectly() throws Exception {
                 if (bitWidth == 8) {
                     return (byte) (1 << (bitWidth - 1));
                 } else if (bitWidth <= 32) {
-                    return (int) (1 << (bitWidth - 1));
+                    return (1 << (bitWidth - 1));
                 } else if (bitWidth <= 64) {
-                    return (long) (1L << (bitWidth - 1));
+                    return (1L << (bitWidth - 1));
```
jjohannes, line 568: Adjusted to address "unnecessary cast" warning from `javac`.

gradle/plugins/src/main/kotlin/com.hedera.gradle.java-base.gradle.kts
```diff
@@ -45,56 +42,26 @@ jvmDependencyConflicts {
...
+        "try", // auto-closeable resource ignore is never referenced... (AutoClosableLock)
+        "missing-explicit-ctor", // class ... declares no explicit constructors
+        "removal",
+        "deprecation",
+        "overrides", // overrides equals, but neither it ... overrides hashCode method
+        "unchecked",
+        "rawtypes",
+        "exports",
+        "dep-ann"
+    )
```
jjohannes, line 59: This is what we also have in `hedera-services` and which will be provided by the shared configuration later.

0xivanov (commented):
You have missed one `@Var` in `KeyTest.java`

jjohannes:
Thanks @0xivanov - fixed.

Do you think this can be merged, or is there anything you would like to keep?

0xivanov (dismissed):
LGTM

0xivanov (dismissed):
LGTM

rbarkerSL (dismissed):

jjohannes:
I fixed the conflicts. Now you need to press the button one more time @0xivanov @rbarkerSL.

0xivanov (approved):
LGTM

rbarkerSL (approved):
LGTM

Issue #1975, opened 2024-08-30
Make `testEnv` `AutoCloseable`


0xivanov:
## 🆕🐥 First Timers Only

This issue is reserved for people who have never contributed to [Hedera](https://hedera.com) or any open source project in general.
We know that creating a pull request (PR) is a major barrier for new contributors.
The goal of this issue and all other issues labeled by [**'Good First Issue'**](https://github.com/issues?q=is%3Aopen+is%3Aissue+org%3Ahashgraph+archived%3Afalse+label%3A%22Good+First+Issue%22+) is to help you make your first contribution to Hedera.

## 👾 Description of the issue
A gRPC channel (`ManagedChannelImpl`) is being garbage collected without properly shutting down, causing the following error:

```
SEVERE: *~*~*~ Previous channel ManagedChannelImpl{logId=13, target=127.0.0.1:50211} was garbage collected without being shut down! ~*~*~*
Make sure to call shutdown()/shutdownNow()
java.lang.RuntimeException: ManagedChannel allocation site
```

This occurs because the testing environment is not consistently being closed, leading to open gRPC channels that are not properly terminated. Failure to shut down the channels can result in resource leaks and instability in the tests.

### Proposed Solution:
- To prevent this issue, make the `IntegrationTestEnv` class implement the `AutoCloseable` interface, so it can be automatically closed when the test finishes. This ensures that all gRPC channels are properly shut down, avoiding the need for manual calls to `shutdown()` or `shutdownNow()`.

- Additionally, the initialization of the testing environment should be wrapped in a **try-with-resources** block. This will ensure that the environment is closed automatically after the test, preventing resource leaks and the improper garbage collection of gRPC channels.

## 📋 Step by step guide to do a contribution

If you have never contributed to an open source project at GitHub, the following step-by-step guide will introduce you to the workflow. More information and concrete samples for shell commands for each step can be found in our [CONTRIBUTING.md](https://github.com/hashgraph/.github/blob/main/CONTRIBUTING.md) file.
A more detailed general documentation of the GitHub PR workflow can be found [here](https://github.com/firstcontributions/first-contributions/blob/master/README.md).

- [x] **Claim this issue:** Comment below that you are interested in working on the issue
- [x] **Wait for assignment:** A community member with the given rights will add you as an assignee of the issue
- [x] **Fork the repository:** You can do that in GitHub (by simply clicking the 'fork' button).
- [x] **Check out the forked repository**
[...]

PR #2023, opened 2024-10-04
Make testEnv implement Autoclosable Interface, replace occurrences

steffenboe:
**Description**:
Makes `IntegrationTestEnv` implement Autoclosable and replaces occurrences in tests with `try-with-resources`-block.

**Related issue(s)**:

Fixes #
#1975 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/sdk/src/testIntegration/java/com/hedera/hashgraph/sdk/test/integration/AccountAllowanceIntegrationTest.java
+++ b/sdk/src/testIntegration/java/com/hedera/hashgraph/sdk/test/integration/AccountAllowanceIntegrationTest.java
@@ -37,60 +37,60 @@ public class AccountAllowanceIntegrationTest {
     @Test
     @DisplayName("Can spend hbar allowance")
     void canSpendHbarAllowance() throws Throwable {
-        var testEnv = new IntegrationTestEnv(1);
+        try (var testEnv = new IntegrationTestEnv(1)) {
 
-        var aliceKey = PrivateKey.generateED25519();
-        var aliceId = new AccountCreateTransaction()
-            .setKey(aliceKey)
-            .setInitialBalance(new Hbar(10))
-            .execute(testEnv.client)
-            .getReceipt(testEnv.client)
-            .accountId;
+            var aliceKey = PrivateKey.generateED25519();
+            var aliceId = new AccountCreateTransaction()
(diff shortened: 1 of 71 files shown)


steffenboe:
need to take a look at the indentation changes yet, so i switched this PR to WIP

steffenboe:
ready for review!

0xivanov (commented):
Hey, nice work!
DCO check is failing because the commits are not signed off. Please check out https://github.com/hashgraph/.github/blob/main/CONTRIBUTING.md and see how to set up your GPG keys. 
When you've setup the keys you can rebase your commits while signing them.

0xivanov (commented):
Looks good, there are some usages missing thought, like in `TokenGrantKycIntegrationTest` and `TransactionResponseTest`

steffenboe:
you're correct, and these tests aren't the only ones where a replacement is not possible. this is because the close()-method gets called with parameters there. this is not conforming to the Autoclosable interface, and so these occurrences can't be replaced with a `try-with-resources` block.

0xivanov:
@steffenboe You are right, I had forgotten about that! I was actually thinking a while back to rework the `close` method to not have overloads and remove the wiping logic `wipeAccountHbars(newAccountId, newAccountKey);` . This additional logic is making the github action slow and we don't need it since the operator we are using in the `local-node` has 10k HBARs which should be enough.

Would you like to rework the `close` method to look something like this? :
```java
    public void close() throws Exception {
        if (!operatorId.equals(originalClient.getOperatorAccountId())) {
            var hbarsBalance = new AccountBalanceQuery()
                .setAccountId(operatorId)
                .execute(originalClient)
                .hbars;
            new TransferTransaction()
                .addHbarTransfer(operatorId, hbarsBalance.negated())
                .addHbarTransfer(Objects.requireNonNull(originalClient.getOperatorAccountId()), hbarsBalance)
                .freezeWith(originalClient)
                .signWithOperator(client)
                .execute(originalClient);
            client.close();
        }

        originalClient.close();
    }
```

And then you could finish the rest of the tests to use `try-with-resources`.

steffenboe:
Sounds good! Will work on that ;)

0xivanov (approved):
LGTM, TY @steffenboe

PR #2031, opened 2024-10-08
build: update test-only dependency - Jackson

jjohannes:
**Description**:

Jackson is a transitive dependency of 'json-snapshot'. In order to use Jackson as Java Module, we update it to the latest version instead of patching it.

Unfortunately, 'json-snapshot' has not been updated in years and the way we used it is not compatible with the latest Jackson version.

By adjusting the test setup as described here, we make it work: https://github.com/json-snapshot/json-snapshot.github.io/issues/27#issuecomment-583307554

**Related issue(s)**:

This is (also) done to allow us to use the same Jackson versions across multiple hedera repositories in the context of https://github.com/hashgraph/hedera-services/issues/14255.

**Notes for reviewer**:

**Checklist**


0xivanov (approved):
LGTM

rbarkerSL (approved):
LGTM

Issue #1669, opened 2023-12-09
Add support for spotless formatting of the CI pipelines and Gradle scripts


nathanklick:
## Description

As part of the release automation changes, we need to add spotless support for formatting and enforcing proper formatting of the CI pipeline configuration files and the Gradle build scripts.

jjohannes:
This will be solved by #2033

rbarkerSL:
Blocked by #2033

jjohannes:
Done by #2033 and active now. If anything is not working as expected, please report here: 
https://github.com/hiero-ledger/hiero-gradle-conventions/issues

...as it should work the same in all Hiero/Hedera Gradle setups.

PR #2033, opened 2024-10-09
build: use common Hiero Gradle config

jjohannes:
**Description**:

Before this can be integrated:

- [x] Integrate #2016
- [x] Integrate #2031 
- [x] Publish _hiero-gradle-conventions_ plugins to Gradle Plugin Portal and use from there

**Related issue(s)**:

https://github.com/hashgraph/hedera-services/issues/14255
#1794 
#1669 
#2054


developers.properties
```diff
@@ -0,0 +1 @@
+nikita.lebedev@limechain.tech=Nikita Lebedev
```
0xivanov, line 1: Nikita is no longer maintaining the SDK.

mishomihov00 (dismissed):
Review applies to:
`.github/workflows/build.yml`
`.github/workflows/release-artifacts.yml`

developers.properties
jjohannes: Until now, this was still configured (moved here from another place)

What should we put here?

In other places it currently looks like this, but also there I am not sure if it is right:
https://github.com/hashgraph/hedera-services/blob/14255-reusable-gradle-config/platform-sdk/developers.properties

The information ends up here: https://repo1.maven.org/maven2/com/hedera/hashgraph/sdk/2.45.0/sdk-2.45.0.pom
(`<developers>` section)


Should we put something like?

```
<some-team>@hiero.org=Some Team
```

 Maybe there is some general pattern you want to follow in the future? @hendrikebbers maybe?

jjohannes: After discussing with Hendrik, we concluded to remove this, if you have no objections.

Instead, we add a generic `<developer>The Hiero Team</developer>`.

Otherwise this it an additional place to maintain who develops what, which is quickly outdated (as is is here right now).

0xivanov (approved):
LGTM! Thanks @jjohannes

0xivanov:
Hi @jeromy-cannon , we need a review for this one.

0xivanov:
@hashgraph/devops-ci @hashgraph/release-engineering-managers  Would need a review on this one

jeromy-cannon (approved):

PR #2056, opened 2024-11-01
chore: remove `IllegalStateException` when specific node id is not present in address book

### LINKED ISSUE #2044: IllegalStateException: Some node account IDs did not map to valid nodes in the client's network
Hi,

Our wallet uses the Hedera Java SDK (com.hedera.hashgraph:sdk:2.29.1).
I'm reaching out because we encountered the following error on Hedera Java SDK when executing a transaction that is from the external dApp.

> IllegalStateException: Some node account IDs did not map to valid nodes in the client's network.

The code in the SDK where the exception occurs is as follows: https://github.com/hashgraph/hedera-sdk-java/blob/dce06be5477948f788d76a2a307f9225b89a82c4/sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java#L620

Upon investigating the cause, we found that the list of nodes in the dApp's transaction differs from the list of nodes in the Hedera Java SDK Client that our wallet is using, which leads to the error.

> List of nodes in the dApp's transaction: 
> [0.0.22, 0.0.16, 0.0.17, 0.0.28, 0.0.4, 0.0.5, 0.0.27, 0.0.14, 0.0.23, 0.0.11, 0.0.6, 0.0.19, 0.0.7, 0.0.10, 0.0.12, 0.0.31, 0.0.13, 0.0.30, 0.0.26, 0.0.8, 0.0.20, 0.0.18, 0.0.3, 0.0.25, 0.0.24, 0.0.21, 0.0.15, 0.0.29, 0.0.9]
> 
> List of nodes in the Hedera Java SDK Client:
> [0.0.5, 0.0.28, 0.0.3, 0.0.27, 0.0.19, 0.0.24, 0.0.15, 0.0.17, 0.0.32, 0.0.34, 0.0.18, 0.0.16, 0.0.22, 0.0.21, 0.0.4, 0.0.8, 0.0.14, 0.0.12, 0.0.23, 0.0.30, 0.0.29, 0.0.10, 0.0.20, 0.0.11, 0.0.9, 0.0.13, 0.0.25, 0.0.7, 0.0.33, 0.0.6, 0.0.31]

The node 0.0.26 is not present in the Client. 
I have heard that this node has been officially excluded from the node list. So the Client seems fine but some dApp may not recognize it.

I suggest enhancing the SDK code so that the transaction execution logic can continue. 
Is there a reason to throw an error and halt the transaction execution just because some nodes are missing on the Client? Even though the transaction can be executed with other nodes.. 
Rather than causing an exception, it would be better to at least log the error and allow the transaction execution logic to continue.
This would provide flexibility in case specific nodes are excluded from the list in the future.

0xivanov:
**Description**:

This PR ignores a node when the node is not present in the address book, instead of throwing an exception.
This helps with wallets not having to update the SDK every time the address book gets changed.

**Related issue(s)**:

Fixes #2044

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
+++ b/sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
@@ -616,8 +616,7 @@ abstract class Executable<SdkRequestT, ProtoRequestT extends MessageLite, Respon
             @Nullable
             var nodeProxies = client.network.getNodeProxies(accountId);
             if (nodeProxies == null || nodeProxies.size() == 0) {
-                throw new IllegalStateException(
-                    "Some node account IDs did not map to valid nodes in the client's network");
+                continue;
             }
 
             var node = nodeProxies.get(random.nextInt(nodeProxies.size()));
--- a/sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
+++ b/sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
@@ -601,7 +601,7 @@ abstract class Executable<SdkRequestT, ProtoRequestT extends MessageLite, Respon
         // failure the system can retry with different proxy on each attempt
         if (nodeAccountIds.size() == 1) {
             var nodeProxies = client.network.getNodeProxies(nodeAccountIds.get(0));
-            if (nodeProxies == null || nodeProxies.size() == 0) {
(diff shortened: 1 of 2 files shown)


agadzhalov (approved):
LGTM

PR #2073, opened 2024-11-11
feat: Implement `MirrorNodeContractQuery` 

### LINKED ISSUE #2076: Implement `MirrorNodeContractQuery`
### Problem

-

### Solution

-

### Alternatives

_No response_

0xivanov:
**Description**:

A new `MirrorNodeContractQuery` class has been introduced to support EVM gas estimation and simulation. This query class provides two distinct types of queries:

1. **`MirrorNodeContractCallQuery`** – Executes a contract call and returns the result, simulating the execution of the contract.
2. **`MirrorNodeContractEstimateGasQuery`** – Estimates the gas usage for a contract call, allowing developers to understand the cost of executing a transaction.

The inputs for these queries match those of `ContractExecuteTransaction`, making it easy for developers to calculate the cost of a follow-up transaction.

The `MirrorNodeContractQuery` issues an HTTP request to the `api/v1/contracts/call` endpoint of the Mirror Node Web3 module, which supports both gas estimation and transient simulation of state-changing transactions. This enables accurate cost predictions and helps developers optimize their contract interactions before executing them on the network.

**Related issue(s)**:

Fixes #2076

**Notes for reviewer**:

**Checklist**


sdk/src/main/java/com/hedera/hashgraph/sdk/MirrorNodeContractQuery.java
```diff
@@ -0,0 +1,208 @@
...
+
+/**
+ * MirrorNodeContractQuery returns a result from EVM execution such as cost-free execution of read-only smart contract
+ * queries, gas estimation, and transient simulation of read-write operations.
+ */
+public class MirrorNodeContractQuery {
+    private ContractId contractId = null;
+    private String contractEvmAddress = null;
+    private byte[] callData;
+    private long blockNumber;
```
Nana-EC, line 43: Missing `gasLimit`, `gasPrice` and `value` params.
See https://testnet.mirrornode.hedera.com/api/v1/docs/#/contracts/contractCall

sdk/src/main/java/com/hedera/hashgraph/sdk/MirrorNodeContractQuery.java
```diff
@@ -0,0 +1,208 @@
...
+                throw new CompletionException(ex);
+            })
+            .thenApply(MirrorNodeContractQuery::parseHexEstimateToLong);
+    }
+
+    static String createJsonPayload(byte[] data, String contractAddress, String blockNumber, boolean estimate) {
+        String hexData = Hex.toHexString(data);
+        return String.format("""
+            {
+              "data": "%s",
```
Nana-EC, line 192: Also missing `from`
Note `from` is also optional in the case we're simulating a contract create flow in which the data contains contract bytecode

Nana-EC (commented):
Nice start.
Missing some fields.
Also you might be able to leverage some Mirror Node web3 module tests as references of example calls.
@kselveliev should be able to help

sdk/src/main/java/com/hedera/hashgraph/sdk/MirrorNodeContractQuery.java
0xivanov: Added

sdk/src/main/java/com/hedera/hashgraph/sdk/MirrorNodeContractQuery.java
0xivanov: Added.

sdk/src/main/java/com/hedera/hashgraph/sdk/MirrorNodeContractQuery.java
```diff
@@ -0,0 +1,314 @@
...
+public class MirrorNodeContractQuery {
+    // The contract we are sending the transaction to
+    private ContractId contractId = null;
+    private String contractEvmAddress = null;
+    // The account we are sending the transaction from
+    private AccountId sender = null;
+    private String senderEvmAddress = null;
+    // The transaction callData
+    private byte[] callData;
+    // The amount we are sending to payable functions
```
kselveliev, line 50: We can send value on a contract deployment call not only payable functions, so the code comment can be adjusted here

sdk/src/testIntegration/java/com/hedera/hashgraph/sdk/test/integration/MirrorNodeContractQueryIntegrationTest.java
```diff
@@ -0,0 +1,278 @@
...
+            var fileId = Objects.requireNonNull(response.getReceipt(testEnv.client).fileId);
+
+            response = new ContractCreateTransaction()
+                .setAdminKey(testEnv.operatorKey)
+                .setGas(200000)
+                .setBytecodeFileId(fileId)
+                .execute(testEnv.client);
+
+            var contractId = Objects.requireNonNull(response.getReceipt(testEnv.client).contractId);
+
```
kselveliev, line 66: Write a comment why we do this

sdk/src/main/java/com/hedera/hashgraph/sdk/EntityIdHelper.java
```diff
@@ -40,7 +42,7 @@
 /**
  * Utility class used internally by the sdk.
  */
-class EntityIdHelper {
+public class EntityIdHelper {
```
0xivanov, line 45: note:
Made this class public in order to access it from the tests, changed most public methods to package private

rbarkerSL (dismissed):
Approval applies to `gradle/plugins/src/main/kotlin/com.hedera.gradle.java.gradle.kts`

kselveliev (approved):
LGTM

sdk/src/main/java/com/hedera/hashgraph/sdk/MirrorNodeContractQuery.java
```diff
@@ -0,0 +1,346 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
+import java.util.concurrent.ExecutionException;
+import org.bouncycastle.util.encoders.Hex;
+
+/**
+ * MirrorNodeContractQuery returns a result from EVM execution such as cost-free execution of read-only smart contract
+ * queries, gas estimation, and transient simulation of read-write operations.
+ */
+public class MirrorNodeContractQuery {
```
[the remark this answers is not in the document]
0xivanov: Updated the PR and the notion page. Introducing 2 new APIs: `MirrorNodeContractCallQuery` and `MirrorNodeContractEstimateGasQuery`

examples/src/main/java/com/hedera/hashgraph/sdk/examples/MirrorNodeContractQueriesExample.java
```diff
@@ -0,0 +1,176 @@
...
+            .setContractId(contractId)
+            .setGas(gas)
+            .setFunction("getMessage")
+            .setQueryPayment(new Hbar(1));
+
+        var result = callQuery
+            .execute(client);
+
+        /*
+         * Step 5:
```
0xivanov, line 153:
```suggestion
         * Step 6:
```

0xivanov:
Hey @Nana-EC @SimiHunjan can you please take a look at the updated version of the queries?
You can focus on the exposed APIs and not so much the business logic.

examples/src/main/java/com/hedera/hashgraph/sdk/examples/MirrorNodeContractQueriesExample.java
```diff
@@ -0,0 +1,176 @@
...
+        /*
+         * Step 6:
+         * Simulate the transaction for free, using the mirror node
+         */
+        var simulationResult = new MirrorNodeContractCallQuery()
+            .setContractId(contractId)
+            .setFunction("getMessage")
+            .execute(client);
+
+        System.out.println("Simulation result: " + decodeABIHex(simulationResult.substring(2)));
```
Nana-EC, line 161: nit; my guess is this is removing the `0x`. could move this into the `decodeABIHex` method.

examples/src/main/java/com/hedera/hashgraph/sdk/examples/MirrorNodeContractQueriesExample.java
```diff
@@ -0,0 +1,176 @@
...
+        System.out.println("Contract call result: " + result.getString(0));
+    }
+    private static String decodeABIHex(String hex) {
```
Nana-EC, line 164: This method could benefit from some function comments to highlight the algorithm approach to a reader.

examples/src/main/java/com/hedera/hashgraph/sdk/examples/MirrorNodeContractQueriesExample.java
```diff
@@ -0,0 +1,176 @@
...
+            .setQueryPayment(new Hbar(1));
+
+        var result = callQuery
+            .execute(client);
+
+        /*
+         * Step 6:
+         * Simulate the transaction for free, using the mirror node
+         */
+        var simulationResult = new MirrorNodeContractCallQuery()
```
Nana-EC, line 156: Would be good to expos all potential params. For instance i should be able to specify `gasLimit` etc here

Nana-EC (commented):
Looking good, some comments

examples/src/main/java/com/hedera/hashgraph/sdk/examples/MirrorNodeContractQueriesExample.java
0xivanov: Done

examples/src/main/java/com/hedera/hashgraph/sdk/examples/MirrorNodeContractQueriesExample.java
0xivanov: Done

examples/src/main/java/com/hedera/hashgraph/sdk/examples/MirrorNodeContractQueriesExample.java
0xivanov: Added more params to the example

agadzhalov (dismissed):

san-est (approved):
applies to files owned by us: `gradle/plugins/src/main/kotlin/com.hedera.gradle.java.gradle.kts`

examples/src/main/java/com/hedera/hashgraph/sdk/examples/MirrorNodeContractQueriesExample.java
```diff
@@ -0,0 +1,193 @@
...
+            .setContractMemo("Simple contract with string field")
+            .execute(client);
+
+        var contractId = Objects.requireNonNull(response.getReceipt(client).contractId);
+
+        /*
+         * Step 3:
+         * Wait for mirror node to import data
+         */
+        Thread.sleep(4000);
```
Nana-EC, line 126: nit: not a fan of sleep, we should be polling something instead 😅 
Maybe poll for transaction records or best would be query `api/v1/transactions/<transactionId>`
Not blocking but something to note

examples/src/main/java/com/hedera/hashgraph/sdk/examples/MirrorNodeContractQueriesExample.java
```diff
@@ -0,0 +1,193 @@
...
+        FileId newFileId = fileCreateTxReceipt.fileId;
+
+        Objects.requireNonNull(newFileId);
+        System.out.println("Created new file with ID: " + newFileId);
+
+        var response = new ContractCreateTransaction()
+            .setAdminKey(client.getOperatorPublicKey())
+            .setGas(200000)
+            .setConstructorParameters(new ContractFunctionParameters().addString("Hello from Hedera."))
+            .setBytecodeFileId(newFileId)
```
Nana-EC, line 116: Q: Do we need a fileCreate or can we pass the bytecode here. Bytecode doesn't look large

examples/src/main/java/com/hedera/hashgraph/sdk/examples/MirrorNodeContractQueriesExample.java
```diff
@@ -0,0 +1,193 @@
...
+            .setGasLimit(30_000)
+            .setBlockNumber(10000)
+            .setGasPrice(1234)
+            .setFunction("getMessage")
+            .execute(client);
+
+        System.out.println("Simulation result: " + decodeABIHexString(simulationResult));
+        System.out.println("Contract call result: " + result.getString(0));
+    }
+    private static String decodeABIHexString(String hex) {
```
Nana-EC, line 171: You should add a comment letting the users know why they need to decode this result.

examples/src/main/java/com/hedera/hashgraph/sdk/examples/MirrorNodeContractQueriesExample.java
```diff
@@ -0,0 +1,193 @@
...
+        // Trim 0x at the beginning
+        if (hex.startsWith("0x")) {
+            hex = hex.substring(2);
+        }
+
+        // Extract the length of the data by parsing the substring from position 64 to 128 as a hexadecimal integer
+        // This section represents the length of the dynamic data, specifically the number of bytes in the string or array
+        int length = Integer.parseInt(hex.substring(64, 128), 16);
```
Nana-EC, line 179: Good logic but it may not be clear to all devs why the indexes at 64, 128 are important.
You should add a function comment that basically gives users an overview of your algorithm so they can understand why your example takes this approach

Nana-EC (commented):
Nice work, left some details on the examples section

examples/src/main/java/com/hedera/hashgraph/sdk/examples/MirrorNodeContractQueriesExample.java
0xivanov: It's a good idea, but I'd like to keep the examples with as little logic as possible and we do not have such query in the SDK itself.

examples/src/main/java/com/hedera/hashgraph/sdk/examples/MirrorNodeContractQueriesExample.java
0xivanov: I added that the mirror node returns the data in ABI Hex format.

examples/src/main/java/com/hedera/hashgraph/sdk/examples/MirrorNodeContractQueriesExample.java
0xivanov: Added javadoc

examples/src/main/java/com/hedera/hashgraph/sdk/examples/MirrorNodeContractQueriesExample.java
0xivanov: Removed the file create

examples/src/main/java/com/hedera/hashgraph/sdk/examples/MirrorNodeContractQueriesExample.java
Nana-EC: Sounds good

examples/src/main/java/com/hedera/hashgraph/sdk/examples/MirrorNodeContractQueriesExample.java
Nana-EC: Yes, that ABI decoding internally would be great.
I think @ochikov and I had chat around this last year.
Maybe something to revisit next year 😆

Nana-EC (approved):
LG, thanks

agadzhalov (approved):

san-est (approved):
re-approving:
applies to files owned by us: gradle/plugins/src/main/kotlin/com.hedera.gradle.java.gradle.kts

Issue #2091, opened 2024-11-19
NodeCreate and NodeUpdate have nullable fields that cannot be set to null


jbair06:
### Problem

There are multiple fields in both NodeCreate and NodeUpdate that are either nullable, and shouldn't be, or are not nullable and maybe should be. In addition, the set methods prevent any field, even nullable fields, from being set to null.

As far as which fields should NOT be nullable, I would think that they all should be. The reason is that there is the case where a transaction may be partially completed, saved, then recreated from bytes. This process may chose to not set, or unset, a field. Not sure if there can be some validation in the freeze method to ensure require fields are set. 

Two examples in NodeCreateTransaction:
* accountId is required and perhaps should not be nullable 
* description is optional and should be nullable (not not be set to an empty string by default).

### Solution

NodeCreate and NodeUpdate should be revised to properly handle the nullability of fields in the transaction. This include adjusting the set methods as needed.

PR #2092, opened 2024-11-20
chore: fix serialize/deserialize for `NodeUpdate` and `NodeCreate`

### LINKED ISSUE #2090: NodeUpdateTransaction gossip and grpc certificate bytes should be BytesValue
### Description

In NodeUpdateTransaction, the gossipCaCertificate and the grpcCertificateHash are of type byte[]. The protobuf specifies these as of type BytesValue. 

The actual bug I am seeing is that when I do not set these values in a NodeUpdateTransaction, then serialize and deserialize the transaction (using toBytes and fromBytes), these values change from a null value to an empty byte[]. This is causing a number of issues both locally, as well as when submitting to the network.

### Steps to reproduce

Change the object types to BytesValue

### Additional context

_No response_

### Hedera network

mainnet

### Version

2.43.0

### Operating system

None

0xivanov:
**Description**:

This PR fixes:
1. `NodeCreate` and NodeUpdate have nullable fields that cannot be set to null.
2. `NodeUpdateTransaction` gossip and grpc certificate bytes deserialize to empty `ByteValue`/`ByteString`

**Related issue(s)**:

Fixes #2090 #2091

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/sdk/src/main/java/com/hedera/hashgraph/sdk/NodeUpdateTransaction.java
+++ b/sdk/src/main/java/com/hedera/hashgraph/sdk/NodeUpdateTransaction.java
@@ -464,13 +464,13 @@ public class NodeUpdateTransaction extends Transaction<NodeUpdateTransaction> {
             serviceEndpoints.add(Endpoint.fromProtobuf(serviceEndpoint));
         }
 
-        var protobufGossipCert = body.getGossipCaCertificate();
-        gossipCaCertificate = protobufGossipCert.equals(BytesValue.getDefaultInstance()) ?
-            null : protobufGossipCert.getValue().toByteArray();
+        if (body.hasGossipCaCertificate()) {
+            gossipCaCertificate = body.getGossipCaCertificate().getValue().toByteArray();
+        }
 
-        var protobufGrpcCert = body.getGrpcCertificateHash();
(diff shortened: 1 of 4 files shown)


sdk/src/main/java/com/hedera/hashgraph/sdk/NodeUpdateTransaction.java
jbair06, line 1: For the update of the GRPC Certificate hash, if an empty array is passed in on purpose in order to clear the value (which I don't think would ever happen), I think that value needs to be retained. Does that happen with this change? Or will an empty byte array always be turned into a null?

0xivanov: Now you cannot "clear" the certificates. The protbuf is treating null and empty array the same, when i call the getter.

agadzhalov (dismissed):
LGTM

sdk/src/main/java/com/hedera/hashgraph/sdk/NodeUpdateTransaction.java
0xivanov: Fixed.

agadzhalov (approved):
LGTM

PR #2096, opened 2024-11-22
fix: `NodeUpdateTransaction` sets description as empty string

### LINKED ISSUE #2091: NodeCreate and NodeUpdate have nullable fields that cannot be set to null

0xivanov:
**Description**:

**Related issue(s)**:

Fixes #2091

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/sdk/src/main/java/com/hedera/hashgraph/sdk/NodeUpdateTransaction.java
+++ b/sdk/src/main/java/com/hedera/hashgraph/sdk/NodeUpdateTransaction.java
@@ -454,7 +454,9 @@ public class NodeUpdateTransaction extends Transaction<NodeUpdateTransaction> {
             accountId = AccountId.fromProtobuf(body.getAccountId());
         }
 
-        description = body.getDescription().getValue();
+        if (body.hasDescription()) {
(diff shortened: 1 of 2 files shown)


agadzhalov (approved):
LGTM

jbair06 (commented):
Thanks for getting this done so quickly! Looks great

PR #2117, opened 2024-12-02
test: Add tests for serialization/deserialization with no setters

### LINKED ISSUE #2101: Review all transactions and queries to verify data is not getting overwritten in the serialization and deserialization process
### Problem

This [bug](https://github.com/hashgraph/hedera-sdk-java/issues/2098) raises some concerns in that there could be potentially be other places where data is overwritten in an unintended way when creating transactions/queries and following the serializing and deserializing process.

I know we couldn't fully test these transactions as Solo is still working on supporting these features so it could be that this case is specific to the node transactions. To be sure, we should look at doing the following:

- Audit other transaction and query types across the SDK and verify there aren't any other cases the intended data input by the user is not being overwritten
- Review our unit tests and add any missing test cases

### Solution

See above

### Alternatives

_No response_

0xivanov:
**Description**:

Add tests for serialization/deserialization with no setters

**Related issue(s)**:

Fixes #2101

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/sdk/src/test/java/com/hedera/hashgraph/sdk/AccountAllowanceApproveTransactionTest.java
+++ b/sdk/src/test/java/com/hedera/hashgraph/sdk/AccountAllowanceApproveTransactionTest.java
@@ -104,6 +104,14 @@ public class AccountAllowanceApproveTransactionTest {
     }
 
     @Test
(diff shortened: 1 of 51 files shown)


agadzhalov (approved):
LGTM

Issue #2017, opened 2024-10-01
Rename any usage of “Hedera”, “Swirlds” or “Hashgraph” to “Hiero”


0xivanov:
### Problem

- [x] https://github.com/hashgraph/hedera-sdk-java/issues/2054
- [x] https://github.com/hashgraph/hedera-sdk-java/pull/2141 - rename packages
- [x] https://github.com/hashgraph/hedera-sdk-java/pull/2150

### Solution

-

0xivanov:
The renaming is done for `/sdk` module:
https://github.com/hashgraph/hedera-sdk-java/pull/2045

We would need someone from devops to rename the gradle files in order to build the other modules.

jjohannes:
I just noticed that the packages after the rename are `org.hiero.sdk`. My understanding by now is that it should rather be
`org.hiero.<some-product-name>.sdk`

This mainly concerns the _group_ defined in _settings.gradle.kts_:

```
    module("sdk") { group = "org.hiero.<some-product-name>" }
    module("sdk-full") { group = "org.hiero.<some-product-name>" }
```

If that is changed. For consistency I would also
- Change all Java packages to `org.hiero.<some-product-name>.sdk`
- Change the name of the `sdk` module to `<some-product-name>-sdk` (ideally by renaming the `sdk` folder).

For more context, see this Slack message: https://swirldslabs.slack.com/archives/C058265B2BS/p1736924474202549


NOTE: I only put this comment here so that these things are considered before the final release of `2.47.0`. Not to force any change. But after the first release to Maven Central with the new name, it can't really be changed anymore.

0xivanov:
Hi @jjohannes ,FYI -  I am unable to see the slack conversation. 
What would be the product name in this case?

jjohannes:
The message is about the general pattern to name things. Which I wrote down for the _hedera-services_ modules.

-----

A good pattern that is commonly used (e.g. by apache libraries) would be to follow this naming scheme consistently:
- GA `domain.product:product-modulename`
- Java Module Name `domain.product.modulename`
- Folder in repository `product-modulename`

For example:
- `domain`  = `org.hiero.node`
- `product` = `consensus`
- `module-name` = `state-api`

Then:
- GA `org.hiero.node.consensus:consensus-state-api`  (G defined in settings.gradle.kts, A derived from folder name)
- Java Module Name `org.hiero.node.consensus.state.api` (defined in module-info.java)
[...]

PR #2141, opened 2024-12-13
chore: rename packages to hiero

0xivanov:
**Description**:

- [x] Rename packages to org.hiero
- [x] Switch licenses to SPDX-License
- [x] spotlessApply

**Related issue(s)**: https://github.com/hashgraph/hedera-sdk-java/issues/2017

Fixes #

**Notes for reviewer**:

**Checklist**


0xivanov:
Hey @jjohannes , do you think it's possible to rename the packages to without breaking the `example` module?

jjohannes:
Renaming a package is technically changing the API. To be consistent, we need to rename everything at once in my opinion:

- The packages in `sdk`
  - All usages of classes from the packages in tests and examples
- The Java Module Names [here](https://github.com/hashgraph/hedera-sdk-java/blob/0a72de56828cc2913b155d97225ccad1534086eb/sdk/src/main/java/module-info.java#L21) and [here](https://github.com/hashgraph/hedera-sdk-java/blob/0a72de56828cc2913b155d97225ccad1534086eb/sdk-full/src/main/java/module-info.java#L21)
- The **group** for the Maven Coordinates [here](https://github.com/hashgraph/hedera-sdk-java/blob/0a72de56828cc2913b155d97225ccad1534086eb/settings.gradle.kts#L7-L9) (and [here](https://github.com/hashgraph/hedera-sdk-java/blob/0a72de56828cc2913b155d97225ccad1534086eb/.github/workflows/release-artifacts.yml#L132) in publishing CI config).

Users that switch to the new Hiero version of the SDK will have to adjust there imports to the new package(s). But the classes themselves will work as before. Only the imports need to be adjusted.

0xivanov:
@jjohannes I don't know why the examples are not building, i haven't touched them. Do you have any ideas?

sdk-full/src/main/java/module-info.java
```diff
@@ -18,7 +18,7 @@
  *
  */
 
-module com.hedera.hashgraph.sdk.full {
+module com.hiero.sdk.full {
```
jjohannes, line 21: I think the new namespace should be `org.hiero`.

jjohannes:
> I don't know why the examples are not building

With the rename, the examples now automatically use the last published version of the `sdk` (2.46.0).

That version is without the recent Guava/CheckerQual version update. But the build setup used for the example expects a newer version due to our strict Java Module setup. The newer version has a `module-info` and this is now expected.

You can update the Guava version in the example directly – [dependency constraints block](https://github.com/hashgraph/hedera-sdk-java/blob/0a72de56828cc2913b155d97225ccad1534086eb/examples/build.gradle.kts#L20-L25) – to make it work with the previous version:
(Not this is only an issue in our strict checking setup, for users of the SDK without these strict checks, the example works)

```
dependencies.constraints {
    implementation("com.google.guava:guava:33.3.1-android")
    ...
}
```

mishomihov00 (approved):
Review applies to:
`.github/workflows/build.yml`
`.github/workflows/release-artifacts.yml`

tck/build.gradle.kts
```diff
@@ -63,6 +62,7 @@ dependencyAnalysis {
         onAny {
             severity("fail")
             exclude("com.google.protobuf:protobuf-javalite")
+            onUnusedDependencies { exclude(":sdk") }
```
jjohannes, line 65: Why is this needed?

.github/workflows/build.yml
```diff
@@ -101,7 +101,7 @@ jobs:
           OPERATOR_ID: '0.0.1022'
           HEDERA_NETWORK: 'localhost'
         run: |
-          ./gradlew -POPERATOR_ID=$OPERATOR_ID -POPERATOR_KEY=$OPERATOR_KEY -PHEDERA_NETWORK=$HEDERA_NETWORK :aggregation:testCodeCoverageReport --scan
+          ./gradlew -POPERATOR_ID=$OPERATOR_ID -POPERATOR_KEY=$OPERATOR_KEY -PHEDERA_NETWORK=$HEDERA_NETWORK :aggregation:testCodeCoverageReport --scan --rerun-tasks
```
jjohannes, line 104: The `--rerun-tasks` should not be added here. It will lead to recompiling the code and not reuse it from the step before. It will also not use anything from the build cache making the remote cache more or less useless. What is the reason for adding it?

tck/build.gradle.kts
0xivanov: some check was failing before, let's see if it's fixed now

.github/workflows/build.yml
0xivanov: Alright

0xivanov: I just wanted to make sure the tests are actually running

tck/build.gradle.kts
0xivanov: the healthcheck fails, not sure why.

jjohannes: It should not fail, but it does... Somehow the rename must confuse it. If you want you can keep it in and I will investigate later.

0xivanov: alright

jjohannes (dismissed):

jjohannes:
The package for the protobuf-generated classes should probably also be changed:

See: https://github.com/hashgraph/hedera-sdk-java/blob/main/sdk/src/main/proto/account.proto#L28-L29
And all other `proto` files

The content of `sdk/build/classes/java/main` after building, shows if you still have any `com.hedera` classes.

jjohannes (approved):

rbarkerSL (approved):
Approval applies to:
- .github/workflows/release-artifacts.yml
- example-android/app/build.gradle.kts
- example-android/build.gradle.kts
- example-android/settings.gradle.kts
- examples/build.gradle.kts
- settings.gradle.kts
- tck/build.gradle.kts

agadzhalov (approved):
LGTM, nice work, tyvm @0xivanov @jjohannes

Issue #2075, opened 2024-11-12
Epic: Migrate to Java 21


0xivanov:
### Action items

- [x] Bump Java version to 21
- [ ] Deprecate async/reactive code
- [x] Leverage virtual threads (introduced in Project Loom) instead of traditional ExecutorService for gRPC communication.
- [x] Manual testing

SimiHunjan:
@naydenovn should we start to review this one?

naydenovn:
Hi, @SimiHunjan. We will discuss it together with @0xivanov and after all priorities are considered we could take it into the sprint.

SimiHunjan:
thanks @naydenovn !

PR #2145, opened 2024-12-16
ci: pin exact Java version

jjohannes:
**Description**:

This is important for deterministic and reproducible builds. Without this, the version automatically changes when there is a new release.


.github/workflows/build.yml
```diff
@@ -43,7 +43,7 @@ jobs:
         uses: actions/setup-java@8df1039502a15bceb9433410b1a100fbe190c53b # v4.5.0
         with:
           distribution: temurin
-          java-version: 17
+          java-version: '17.0.13'
```
rbarkerSL, line 46: Is there a reason we are on 17.0.13 instead of 21.0.4?

rbarkerSL (commented):
Curious about this java version, should we not be building against the latest?

.github/workflows/build.yml
jjohannes: For moving from 17 to 21, there is a separate issue: 
https://github.com/hashgraph/hedera-sdk-java/issues/2075

One concern may be to not force users of the SDK to update their projects (yet).

0xivanov: @jjohannes @rbarkerSL Updating to java 21 might break some users and we haven't gotten to it yet, so the runners should use the same version as the target for the project - 17

rbarkerSL (approved):

Issue #2161, opened 2025-01-09
TCK: Implement token management transactions


0xivanov:
### Problem

-

### Solution

-

PR #2166, opened 2025-01-14
feat(TCK): Implement token management transactions

naydenovn:
Implements the following TCK transactions:
 - [x] `TokenFeeScheduleUpdate` endpoint
 - [x] `TokenAssociateTransaction` endpoint
 - [x] `TokenDissociateTransaction` endpoint
 - [x] `TokenPauseTransaction` endpoint
 - [x] `TokenUnpauseTransaction` endpoint
 - [x] `TokenFreezeTransaction` endpoint
 - [x] `TokenUnfreezeTransaction` endpoint
 - [x] `TokenGrantKycTransaction` endpoint
 - [x] `TokenRevokeKycTransaction ` endpoint
 - [x] `TokenMintTransaction` endpoint
 - [x] `TokenBurnTransaction` endpoint

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-sdk-java/issues/2161

**Checklist**


examples/src/main/java/com/hedera/hashgraph/sdk/examples/AccountCreationWithFundsTransfer.java
```diff
@@ -0,0 +1,60 @@
+package com.hedera.hashgraph.sdk.examples;
```
0xivanov, line 1: Remove this file

tck/src/main/java/org/hiero/tck/methods/sdk/param/token/AssociateDisassociateTokenParams.java
```diff
@@ -0,0 +1,39 @@
...
+            var parsedCommonTransactionParams = JSONRPCParamParser.parseCommonTransactionParams(jrpcParams);
+
+            return new AssociateDisassociateTokenParams(parsedAccountId, parsedTokenIds, parsedCommonTransactionParams);
+        } catch (ClassCastException e) {
+            throw new JSONRPCParseException("Invalid parameter type", e);
+        } catch (JSONRPCParseException e) {
+            throw e;
+        } catch (Exception e) {
+            throw new JSONRPCParseException("Failed to parse AssociateDisassociateToken parameters", e);
+        }
```
0xivanov, line 37: We can remove the error handling and just throw generic exception on method level. The handling would be done in `AbstractJSONRPC2Service: getArguments`.
We do not care what kind of exception we caught - if it happened in the `parse` method of the params we just propagate to the client that the param types are not right.

0xivanov: Same applies to all param files.

tck/src/main/java/org/hiero/tck/exception/JSONRPCParseException.java
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: Apache-2.0
+package org.hiero.tck.exception;
+
+public class JSONRPCParseException extends RuntimeException {
```
0xivanov, line 4: Can be removed.

tck/src/main/java/org/hiero/tck/util/JSONRPCParamParser.java
```diff
@@ -0,0 +1,66 @@
...
+                try {
+                    return new CustomFee().parse(jsonObj);
+                } catch (Exception e) {
+                    throw new JSONRPCParseException("Failed to parse CustomFee", e);
+                }
+            });
+        } catch (Exception e) {
+            throw new JSONRPCParseException("Failed to parse CustomFees array", e);
+        }
+    }
```
0xivanov, line 36: I think we can just throw the exception, because the error message will get overridden anyways and we don't have specific handling for `JSONRPCParseException`
```suggestion
    public static Optional<List<CustomFee>> parseCustomFees(Map<String, Object> jrpcParams) throws Exception {
            return parseJsonArray(jrpcParams, "customFees", jsonObj -> new CustomFee().parse(jsonObj));
    }

```

0xivanov (approved):
LGTM

PR #2167, opened 2025-01-14
build: update hiero-gradle-conventions to 0.3.0

jjohannes:
**Description**:

- Add `feature.publish-dependency-constraints` plugin (https://github.com/hiero-ledger/hiero-gradle-conventions/issues/41)
- Extends Yaml formatting by spotlessApply (https://github.com/hiero-ledger/hiero-gradle-conventions/issues/30)
- Automatically publishes build scans on CI so that we can remove `--scan` parameters (https://github.com/hiero-ledger/hiero-gradle-conventions/issues/32)


.github/workflows/build.yml
```diff
@@ -87,27 +87,27 @@ jobs:
...
 
       - name: Start Local Node
         run: npx @hashgraph/hedera-local start -d --network local --network-tag=0.57.0
 
       - name: Run Unit and Integration Tests
         env:
-          OPERATOR_KEY: '0xa608e2130a0a3cb34f86e757303c862bee353d9ab77ba4387ec084f881d420d4'
-          OPERATOR_ID: '0.0.1022'
-          HEDERA_NETWORK: 'localhost'
+          OPERATOR_KEY: "0xa608e2130a0a3cb34f86e757303c862bee353d9ab77ba4387ec084f881d420d4"
```
mishomihov00, line 100: This key must be a secret. @rbarkerSL Please confirm.

jjohannes: Could we do this change in a separate PR as this is only a reformatting of the existing configuration?

0xivanov: This key is for the local node env, so it's not an issue to be public.

0xivanov (approved):
LGTM

PR #2179, opened 2025-01-16
build: replace Taskfile with Gradle tasks

jjohannes:
**Description**:

The only things in the Taskfile that were still important are the _update*_ tasks. All other tasks just forwarded to Gradle.

With this change, the "update" tasks are implemented in Gradle and Taskfile (and with that a separate tool) is no longer needed.

While testing the `updateProto` task, I discovered that we still missed some adjustments in the "hedera -> hiero" change. These are done in this PR as well.

Also, remove `developers.properties` - see: https://github.com/hiero-ledger/hiero-gradle-conventions/issues/37

**Related issue(s)**:

#1794 - addresses open issue: _Revisit `Taskfile.yml` (maybe make it as gradle tasks)_


andrewb1269hg (approved):

0xivanov (approved):
LGTM

PR #2198, opened 2025-01-28
chore: rename 'sdk' to 'sdk-java'

jjohannes:
Attempt to do the Artifact / Module Name rename as suggested in https://github.com/hiero-ledger/hiero-sdk-java/issues/2017#issuecomment-2596251930


gradle/aggregation/build.gradle.kts
```diff
@@ -24,15 +24,15 @@ tasks.testCodeCoverageReport {
                 componentFilter { id -> id is ProjectComponentIdentifier }
                 attributes.attribute(
                     Category.CATEGORY_ATTRIBUTE,
-                    objects.named(Category.VERIFICATION)
+                    objects.named(Category.VERIFICATION),
```
andrewb1269hg, line 27: Is it expected to have a comma at the end of this line? Looks like the comma is new. Same for lines 31 and 35 of this file.

sdk-java-full/build.gradle.kts
```diff
@@ -17,17 +17,17 @@ dependencies {
 
 javaModuleDependencies.moduleNameToGA.put(
     "com.google.protobuf",
-    "com.google.protobuf:protobuf-java"
+    "com.google.protobuf:protobuf-java",
```
andrewb1269hg, line 20: Is it expected to have a comma at the end of this line?

sdk-java/build.gradle.kts
```diff
@@ -10,7 +10,7 @@ description = "Hedera™ Hashgraph SDK for Java"
 
 javaModuleDependencies.moduleNameToGA.put(
     "com.google.protobuf",
-    "com.google.protobuf:protobuf-javalite"
+    "com.google.protobuf:protobuf-javalite",
```
andrewb1269hg, line 13: Same comment about comma here.

andrewb1269hg (changes requested):

gradle/aggregation/build.gradle.kts
jjohannes: This is done by the auto-formatting, where we use [ktmft](https://github.com/facebook/ktfmt) in the default configuration for Kotlin files. Looks like something changed in the defaults (it now consistently adds the trailing comma). [Maybe it is this change.](https://github.com/facebook/ktfmt/issues/216)

With one of the updates of `hiero-gradle-conventions`, this update was deployed to this repo. And this is the first time this file was touched since then. Hence this reformatting.

If it is fine with you, I would stick with the defaults and not start customizing the ktfmt config. It only concerns Gradle files as we do not use Kotlin in other places.

0xivanov (approved):
LGTM

gradle/aggregation/build.gradle.kts
0xivanov: It's intentional so we should be good for now.

mishomihov00 (approved):

PR #2200, opened 2025-01-29
wip: Refactor `TokenRejectFlow`

naydenovn:
Currently, the TokenRejectFlow class and similar transaction flows are overly reliant on class hierarchies that treat flows as if they are transactions themselves. This design complicates the implementation and maintenance of transaction flows.

To improve this, we should refactor transaction flows so that transactions are treated as members of the flows rather than flows inheriting from transactions.

Fixes: https://github.com/hiero-ledger/hiero-sdk-java/issues/2189


sdk/src/main/java/org/hiero/sdk/TokenRejectFlow.java
```diff
@@ -144,6 +150,44 @@ public TokenRejectFlow setNodeAccountIds(List<AccountId> nodeAccountIds) {
...
+    /**
+     * Assign the tokenRejectTransaction.
+     * @param tokenRejectTransaction the tokenRejectTransaction.
+     * @return {@code this}
+     */
+    public TokenRejectFlow setTokenRejectTransaction(TokenRejectTransaction tokenRejectTransaction) {
+        Objects.requireNonNull(tokenRejectTransaction);
+        this.tokenRejectTransaction = tokenRejectTransaction;
+        return this;
+    }
```
0xivanov, line 170: Should not have setters for the transactions. They should be initialized internally.

sdk/src/main/java/org/hiero/sdk/TokenRejectFlow.java
```diff
@@ -49,6 +49,12 @@ public class TokenRejectFlow {
     @Nullable
     private UnaryOperator<byte[]> transactionSigner = null;
 
+    @Nullable
+    private TokenRejectTransaction tokenRejectTransaction = null;
```
0xivanov, line 53:
```suggestion
    private TokenRejectTransaction tokenRejectTransaction = new TokenRejectTransaction();
```

sdk/src/main/java/org/hiero/sdk/TokenRejectFlow.java
0xivanov: We should only expose them via getters.

examples/src/main/java/org/hiero/sdk/examples/TokenRejectExample.java
```diff
@@ -202,10 +202,20 @@ public static void main(String[] args) throws Exception {
...
+        TokenRejectTransaction rejectTx = new TokenRejectTransaction();
+        TokenDissociateTransaction dissTx = new TokenDissociateTransaction();
+
+        /*
+         * Step 9:
          * Execute the token reject flow -- reject NFTs.
          */
         System.out.println("Receiver rejects example NFTs...");
         new TokenRejectFlow()
+                .setTokenRejectTransaction(rejectTx)
```
0xivanov, line 217: should create a new flow, then get it's reject and dissociate txns and set owner and the token ids directly to them.

PR #2206, opened 2025-01-31
chore: Update hiero conventions

0xivanov:


0xivanov:
@jjohannes Do you think merging this would fix https://github.com/hiero-ledger/hiero-sdk-java/actions/runs/13071506171/job/36474073957?pr=2205 , as we did before?

jjohannes:
@0xivanov not sure. I commented on the other PR what you can do to make sure we do not hit the bug.

sdk-java/build.gradle.kts
```diff
@@ -31,7 +31,7 @@ testModuleInfo {
     requires("org.junit.jupiter.params")
     requires("org.mockito")
 
-    requiresStatic("java.annotation")
+    requires("java.annotation")
```
jjohannes, line 34:
```suggestion
```

This entry can and should now be removed completely the "test scope" inherits this from the "main scope" and this is now respected by the analysis. It should not suggest to add something else. I will have to check what's wrong there.

jjohannes (changes requested):

PR #2213, opened 2025-02-04
Fix decimal field in java tck

naydenovn:
Solves issue in java tck regarding field type changes in js tck tests


0xivanov (approved):
LGTM

PR #2224, opened 2025-02-11
chore: Snyk Integration

rbarker-dev:
Add snyk_monitor.yml and add Snyk checks in CI during build.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -89,7 +89,7 @@ jobs:
             ) &&
             !cancelled()
           }}
-        run: ${CG_EXEC} snyk test --all-sub-projects --severity-threshold=high --policy-path=.snyk --json-file-output=snyk-test.json --org=hiero-client-sdks
+        run: ${CG_EXEC} snyk test --all-projects --severity-threshold=high --policy-path=.snyk --json-file-output=snyk-test.json --org=hiero-client-sdks
 
       - name: Snyk Code
         id: snyk-code
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -40,6 +40,11 @@ jobs:
       - name: Checkout Code
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
+      - name: Setup NodeJS
+        uses: actions/setup-node@1d0ff469b7ec7b3cb9d8673fde0c81c44821de2a # v4.2.0
+        with:
+          node-version: 20
+
       - name: Setup Java
         uses: actions/setup-java@3a4f6e1af504cf6a31855fa899c6aa5355ba6c12 # v4.7.0
         with:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -40,7 +40,7 @@ jobs:
       - name: Checkout Code
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
-          fetch-depth: '0'
+          fetch-depth: "0"
 
       - name: Setup NodeJS
         uses: actions/setup-node@1d0ff469b7ec7b3cb9d8673fde0c81c44821de2a # v4.2.0
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -39,6 +39,8 @@ jobs:
 
       - name: Checkout Code
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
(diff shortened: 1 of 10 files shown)


sdk/src/main/java/com/hedera/hashgraph/sdk/PrivateKeyECDSA.java
```diff
@@ -158,6 +158,7 @@ public PrivateKey derive(int index) {
 
         var ki = keyData.add(new BigInteger(1, il)).mod(ECDSA_SECP256K1_CURVE.getN());
 
+        // The hardcoded array data is required for compatibility with 24-word recovery phrases
```
nathanklick, line 161: This one seems out of place and not relevant.

```suggestion
```

jsync-swirlds (approved):

nathanklick (approved):

settings.gradle.kts
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: Apache-2.0
-plugins { id("org.hiero.gradle.build") version "0.3.1" }
+plugins { id("org.hiero.gradle.build") version "0.3.3" }
```
jjohannes, line 2:
```suggestion
plugins { id("org.hiero.gradle.build") version "0.3.4" }
```

(but you can also do another update later)

sdk/build.gradle.kts
```diff
@@ -30,8 +30,7 @@ testModuleInfo {
     requires("org.junit.jupiter.api")
     requires("org.junit.jupiter.params")
     requires("org.mockito")
-
-    requiresStatic("java.annotation")
+    requires("java.annotation")
```
jjohannes, line 33:
```suggestion
```

Should be removed instead of changed.

jjohannes (changes requested):

sdk/build.gradle.kts
jjohannes: This is caused by update of dependency analysis plugin. Same as: https://github.com/hashgraph/hedera-services/pull/17876

.github/workflows/build.yml
```diff
@@ -38,6 +39,13 @@ jobs:
 
       - name: Checkout Code
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          fetch-depth: "0"
+
+      - name: Setup NodeJS
+        uses: actions/setup-node@1d0ff469b7ec7b3cb9d8673fde0c81c44821de2a # v4.2.0
+        with:
+          node-version: 20
```
PavelSBorisov, line 48: I think the node version may need to be 18 instead of 20 for the following reasons:
1. The [package.json of the snyk tool marks the engine as node ^18](https://github.com/snyk/cli/blob/main/package.json#L24)
2. Uniformity across jobs in the workflows

rbarker-dev: I've updated node version and java version for uniformity.

@0xivanov node-version needs to be updated to version 20-LTS or version 22 soon as node 18 goes end of life in April of this year. That should happen on another PR.

.github/workflows/snyk-monitor.yml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@94baf225fe0a508e581a564467443d0e2379123b # v4.3.0
+        with:
+          gradle-version: wrapper
+
+      - name: Compile
+        run: ./gradlew assemble
+
+      - name: Disable Gradle Configuration Cache
+        run: sed -i 's/^org.gradle.configuration-cache=.*$/org.gradle.configuration-cache=false/' gradle.properties
```
PavelSBorisov, line 42: shouldn't disabling the config cache be ordered before the the compile step?

rbarker-dev: Possibly......... ;)

jjohannes (approved):

PavelSBorisov (approved):

.github/workflows/snyk-monitor.yml
jjohannes: Mh no. The only reason to disable it is because Snyk does something that breaks this feature.
So doing it right before Snyk runs would be "more correct" imo. And easier to find later when it is fixed in Snyk....

There also seems to be a way to configure this without rewriting the properties directly in the Snyk call:
https://github.com/snyk/snyk-gradle-plugin/issues/239#issuecomment-1367246659

It does not really matter though as this is only a Gradle startup optimization which does not have much effect on CI.

0xivanov (approved):
LGTM

Issue #2018, opened 2024-10-01
HIP-991: Permissionless revenue-generating Topic IDs for Topic Operators


0xivanov:
HIP
https://hips.hedera.com/hip/hip-991

- [ ] Implement `TopicAllowanceApproveTransaction`
- [ ] Implement `TokenTopicAllowance` and `HbarTopicAllowance`
- [ ] Update `TopicCreateTransaction`, `TopicUpdateTransaction`, and `TopicInfo` to support new fields:
1.  `feeSchedule/feeExempt/freeMessage` keys
2. `get/setCustomFees`

PR #2233, opened 2025-02-20
feat(HIP-991): Permissionless revenue topics

naydenovn:
This HIP introduces new transactions and updates to existing transactions in the consensus node software. SDKs will need to support these new transactions and updates to existing ones so that developers can take advantage of the new features.

- [x] Implementation
- [x] Unit tests
- [x] E2E tests
- [x] Example

Fixes: #2018


sdk/src/main/java/com/hedera/hashgraph/sdk/Transaction.java
```diff
@@ -230,6 +233,9 @@ public abstract class Transaction<T extends Transaction<T>>
                 DurationConverter.fromProtobuf(sourceTransactionBody.getTransactionValidDuration()));
         setMaxTransactionFee(Hbar.fromTinybars(sourceTransactionBody.getTransactionFee()));
         setTransactionMemo(sourceTransactionBody.getMemo());
+        setCustomFeeLimits(customFeeLimits.stream()
+                .map(x -> CustomFeeLimit.fromProtobuf(x.toProtobuf()))
+                .toList());
```
0xivanov, line 238: This logic should set the SDK transaction's custom fees from protbuf transaction's custom fees. THe protobuf transaction in this case is the `sourceTransactionBody`.
Your implementation is setting the SDK transaction's custom fees from itself, from what i'm seeing - which would do nothing.

sdk/src/main/java/com/hedera/hashgraph/sdk/Transaction.java
```diff
@@ -744,6 +750,47 @@ public final T setTransactionMemo(String memo) {
         return (T) this;
     }
 
+    /**
+     * Extract the custom fee limits of the transaction
+     * @return the custom fee limits of the transaction
+     */
+    public List<CustomFeeLimit> getCustomFeeLimits() {
```
0xivanov, line 757: These getters/setters should be removed, because we don't want to expose them for every transaction that extends this class. You've made `customFeeLimits` package protected, so you can access it directly in the SDK -more specifically in the `TopicSubmitTransaction`'s setters and getters.

sdk/src/main/java/com/hedera/hashgraph/sdk/CustomFeeLimit.java
```diff
@@ -0,0 +1,87 @@
...
+    /**
+     * Constructor
+     *
+     * @param payerId       the payer's account id
+     * @param customFees    maximum fees user is willing to pay
+     */
+    public CustomFeeLimit(AccountId payerId, List<CustomFixedFee> customFees) {
+        this.payerId = payerId;
+        this.customFees = customFees;
+    }
```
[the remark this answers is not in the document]
0xivanov: If you want, you can make this constructor package private, in order to not change the current usages of it in the SDK.

sdk/src/main/java/com/hedera/hashgraph/sdk/CustomFeeLimit.java
```diff
@@ -0,0 +1,87 @@
...
+ * A maximum custom fee that the user is willing to pay.
+ * <p>
+ * This message is used to specify the maximum custom fee that given user is
+ * willing to pay.
+ */
+public class CustomFeeLimit {
+
+    public AccountId payerId;
+
+    public List<CustomFixedFee> customFees;
```
0xivanov, line 19: private

0xivanov (approved):

PavelSBorisov (approved):

PR #2285, opened 2025-04-07
2229 hip 551 batch transactions

naydenovn:
**Description**:

**Related issue(s)**:

Fixes #
https://github.com/hiero-ledger/hiero-sdk-java/issues/2229

**Notes for reviewer**:

**Checklist**


sdk/src/main/java/com/hedera/hashgraph/sdk/BatchTransaction.java
```diff
@@ -0,0 +1,157 @@
...
+        }
+    }
+
+    /**
+     * Initialize from the transaction body.
+     */
+    void initFromTransactionBody() {
+        var body = sourceTransactionBody;
+        System.out.println("hii");
+    }
```
0xivanov, line 128: It might be impossible to create batch txn from a protobuf body, since in the protobuf body we have a list of signed transaction bytes list, which I don't think could be converted to list of transactions.
We might need to throw `UnsupportedOperationException` here.
WDYT @rwalworth @gsstoykov @SimiHunjan ?

sdk/src/main/java/com/hedera/hashgraph/sdk/BatchTransaction.java
```diff
@@ -0,0 +1,155 @@
...
+            transaction.validateChecksums(client);
+        }
+    }
+
+    /**
+     * Initialize from the transaction body.
+     */
+    void initFromTransactionBody() {
+        var body = sourceTransactionBody;
+        System.out.println("hii");
```
gsstoykov, line 125: nit: Marking just so you don't forget :D

sdk/src/main/java/com/hedera/hashgraph/sdk/Transaction.java
```diff
@@ -750,6 +756,28 @@ public final T setTransactionMemo(String memo) {
         // noinspection unchecked
         return (T) this;
     }
+
+    /**
+     * This method is used to mark a transaction as part of a batch transaction or so-called inner transaction.
+     * Transaction is signed by the executor of the batch transaction
```
0xivanov, line 762:
```js
// Batchify method is used to mark a transaction as part of a batch transaction or make it so-called inner transaction.
// The Transaction will be frozen and signed by the operator of the client.
```

sdk/src/main/java/com/hedera/hashgraph/sdk/Transaction.java
```diff
@@ -1143,7 +1175,12 @@ public T freezeWith(@Nullable Client client) {
             }
 
             try {
-                nodeAccountIds.setList(client.network.getNodeAccountIdsForExecute());
+                if (batchKey == null) {
+                    nodeAccountIds.setList(client.network.getNodeAccountIdsForExecute());
+                } else {
+                    nodeAccountIds.setList(
+                            Collections.singletonList(AccountId.fromString(ATOMIC_BATCH_NODE_ACCOUNT_ID)));
```
0xivanov, line 1182: Might need to revisit this, since `0.0.0` will not work on multi-realm/shard env.

ivaylonikolov7: I think it's supposed to be 0.0.0 as part of the HIP:

> Must have nodeAccountID set to 0.0.0. This acts as a marker to indicate that an inner transaction is part of a batch, and it allows resubmitting a batch transaction without the need to re-sign the inner transactions.

sdk/src/testIntegration/java/com/hedera/hashgraph/sdk/test/integration/BatchTransactionIntegrationTest.java
```diff
@@ -0,0 +1,221 @@
...
+
+            var key = PrivateKey.generateECDSA();
+
+            var tx1 = new AccountCreateTransaction()
+                    .setKeyWithoutAlias(key)
+                    .setInitialBalance(new Hbar(1))
+                    .batchify(testEnv.client, testEnv.operatorKey);
+
+            var tx2 = new AccountCreateTransaction()
+                    .setKeyWithoutAlias(key)
```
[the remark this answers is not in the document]
0xivanov: `key` is newly created

sdk/src/main/java/com/hedera/hashgraph/sdk/BatchTransaction.java
```diff
@@ -0,0 +1,256 @@
...
+     * @throws IllegalStateException if this transaction is frozen
+     * @throws IllegalStateException if any inner transaction is not frozen or missing a batch key
+     * @throws IllegalArgumentException if any transaction is of a blacklisted type
+     */
+    public BatchTransaction setInnerTransactions(List<Transaction> transactions) {
+        Objects.requireNonNull(transactions);
+        requireNotFrozen();
+
+        // Validate all transactions before setting
+        transactions.forEach(this::validateInnerTransaction);
```
ivaylonikolov7, line 114: wdyt about checking if the transaction is signed/frozen in the validate transaction?

ivaylonikolov7 (commented):
Doesn't it make more sense to use a separate client(aliceClient, bobClient, caroleClient) in both functions. This way we won't have to generate transactionIds, signing transactions? Also I think this way we won't have to pass the client in the methods. We would have to exclude the account creation outside of the method. But I don't think it makes sense to have a function called `executeBatchWithBatchify` that also creates accounts. Kinda ruins the single responsibility principle.

sdk/src/main/java/com/hedera/hashgraph/sdk/Transaction.java
```diff
@@ -633,6 +640,10 @@ protected ScheduleCreateTransaction doSchedule(TransactionBody.Builder bodyBuild
         return scheduled;
     }
 
+    protected boolean isBatchedAndNotBatchTransaction() {
+        return batchKey != null && !(this instanceof BatchTransaction);
```
ivaylonikolov7, line 644: are we keeping this check - `!(this instanceof BatchTransaction)`  just for better readability of the code? 

I think we already override it in the batchtransaction to `return true`.

sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
```diff
@@ -364,7 +364,9 @@ public O execute(Client client) throws TimeoutException, PrecheckStatusException
      */
     public O execute(Client client, Duration timeout) throws TimeoutException, PrecheckStatusException {
         Throwable lastException = null;
-
+        if (isBatchedAndNotBatchTransaction()) {
+            throw new IllegalArgumentException("Cannot execute batchified transaction outside of BatchTransaction");
+        }
```
ivaylonikolov7, line 369: Is this something Java specific because I can't seem to find a scenario where you would enter this in the executable class considering you have a `return false` in lines 459-462?

sdk/src/main/java/com/hedera/hashgraph/sdk/BatchTransaction.java
0xivanov: You mean the inner transactions or the batch transaction?

sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
0xivanov: I'm overriding `isBatchedAndNotBatchTransaction` in the `Transaction.java` class and for all other childs of `Executable` (in this case queries) it will return false, since this whole logic does not apply to `Query`

0xivanov:
@ivaylonikolov7 It's an example and it's purpose is to showcase different flows. In this case using `batchify` or not using `batchify`, it's 2 examples in one and each one is self-sufficient - that's the reason why we don't reuse accounts.

0xivanov (commented):
LGTM

PavelSBorisov (approved):

0xivanov (approved):

PR #2322, opened 2025-05-05
chore: append address book on node creation from network entry

naydenovn:
**Description**:

**Related issue(s)**:

Fixes #
https://github.com/hiero-ledger/hiero-sdk-java/issues/2317

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/sdk/src/main/java/com/hedera/hashgraph/sdk/Network.java
+++ b/sdk/src/main/java/com/hedera/hashgraph/sdk/Network.java
@@ -234,6 +234,12 @@ class Network extends BaseNetwork<Network, AccountId, Node> {
 
     @Override
     protected Node createNodeFromNetworkEntry(Map.Entry<String, AccountId> entry) {
+        if (addressBook != null && !addressBook.isEmpty()) {
+            return new Node(entry.getValue(), entry.getKey(), executor)
+                    .setVerifyCertificates(verifyCertificates)
+                    .setAddressBookEntry(addressBook.get(entry.getValue()));
+        }
+
         return new Node(entry.getValue(), entry.getKey(), executor).setVerifyCertificates(verifyCertificates);
     }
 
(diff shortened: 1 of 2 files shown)


sdk/src/main/java/com/hedera/hashgraph/sdk/Network.java
```diff
@@ -234,6 +234,12 @@ synchronized Map<String, AccountId> getNetwork() {
 
     @Override
     protected Node createNodeFromNetworkEntry(Map.Entry<String, AccountId> entry) {
+        if (addressBook != null && !addressBook.isEmpty()) {
+            return new Node(entry.getValue(), entry.getKey(), executor)
+                    .setVerifyCertificates(verifyCertificates)
+                    .setAddressBookEntry(addressBook.get(entry.getValue()));
+        }
+
         return new Node(entry.getValue(), entry.getKey(), executor).setVerifyCertificates(verifyCertificates);
```
steven-sheehy, line 243: This doesn't handle another corner case where the address book has `N` entries and gets updated to have `N-X` entries e.g. it shrinks in size. If the previously set address book entry goes missing or the entire address book is cleared then it should also clear the corresponding `node.addressBookEntry` values.

```suggestion
        var addressBookEntry = addressBook != null ? addressBook.get(entry.getValue()) : null;
        return new Node(entry.getValue(), entry.getKey(), executor)
            .setAddressBookEntry(addressBookEntry)
            .setVerifyCertificates(verifyCertificates);
```

naydenovn: Note taken. I've made the necessary changes. Thanks for your input.

steven-sheehy (approved):

0xivanov (approved):
LGTM

PR #2343, opened 2025-05-22
build: bump Gradle to 8.14 / gradle-conventions to 0.4.1 / example deps

jjohannes:
**Description**:

- bump Gradle to 8.14
- bump hiero-gradle-conventions to 0.4.1
- bump all dependencies in examples to latest versions


.github/workflows/build.yml
```diff
@@ -89,7 +89,7 @@ jobs:
             ) &&
             !cancelled()
           }}
-        run: ${CG_EXEC} snyk test --all-projects --severity-threshold=high --policy-path=.snyk --json-file-output=snyk-test.json --org=hiero-client-sdks
+        run: ${CG_EXEC} snyk test --all-projects '--configuration-matching=^(?!_internal-).*' --severity-threshold=high --policy-path=.snyk --json-file-output=snyk-test.json --org=hiero-client-sdks
```
jjohannes, line 92: Exclude "_internal-*" that the Android plugin automatically adds to the Android example. Snyk currently fails although we do not use these: There are no tests in the example (but Snyk still does "catch all" by default).

PavelSBorisov (approved):

PR #2402, opened 2025-07-10
jumbo transaction

naydenovn:
**Description**:
Deprecation of EthereumFlow
EthereumTransaction should be used instead.

With the introduction of jumbo transactions, it should always be less cost and more efficient to use EthereumTransaction instead.

**Related issue(s)**:

Fixes https://github.com/hiero-ledger/hiero-sdk-java/issues/2250

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/sdk/src/test/java/com/hedera/hashgraph/sdk/EthereumFlowMockTest.java
+++ b/sdk/src/test/java/com/hedera/hashgraph/sdk/EthereumFlowMockTest.java
@@ -4,7 +4,6 @@ package com.hedera.hashgraph.sdk;
 import static org.assertj.core.api.Assertions.assertThat;
 
 import com.google.protobuf.ByteString;
-import com.hedera.hashgraph.sdk.proto.FileID;
 import com.hedera.hashgraph.sdk.proto.Response;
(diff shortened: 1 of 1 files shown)


sdk/src/main/java/com/hedera/hashgraph/sdk/EthereumFlow.java
```diff
@@ -10,16 +10,17 @@
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 import javax.annotation.Nullable;
+import org.bouncycastle.util.encoders.Hex;
 
 /**
  * Execute an Ethereum transaction on Hedera
  */
 public class EthereumFlow {
```
0xivanov, line 21: Deprecate the ethereum flow and add something like this
"// Deprecated: use EthereumTransaction instead.
// With the introduction of jumbo transactions, it should always be less cost and more efficient to use EthereumTransaction instead."

0xivanov: Also add this deprecating notice in the PR description

PR #2406, opened 2025-07-14
add validation for nodeId inside NodeUpdateTransaction and NodeDeleteTransaction

naydenovn:
**Description**:
When using NodeUpdateTransaction, if the nodeId is not explicitly set, it defaults to 0 in the Consensus Node handler, which can unintentionally apply the transaction to node 0.0.3 (account ID 3) in local setups or production environments. This can lead to erroneous or risky node updates.

**Related issue(s)**:

Fixes https://github.com/hiero-ledger/hiero-sdk-java/issues/2388

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/sdk/src/test/java/com/hedera/hashgraph/sdk/NodeDeleteTransactionTest.java
+++ b/sdk/src/test/java/com/hedera/hashgraph/sdk/NodeDeleteTransactionTest.java
@@ -169,7 +169,6 @@ public class NodeDeleteTransactionTest {
         var transaction = new NodeDeleteTransaction();
 
         var exception = assertThrows(IllegalStateException.class, () -> transaction.getNodeId());
-        assertThat(exception.getMessage())
-            .isEqualTo("NodeDeleteTransaction: 'nodeId' has not been set");
+        assertThat(exception.getMessage()).isEqualTo("NodeDeleteTransaction: 'nodeId' has not been set");
     }
 }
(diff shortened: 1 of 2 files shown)


sdk/src/main/java/com/hedera/hashgraph/sdk/NodeUpdateTransaction.java
```diff
@@ -526,4 +526,20 @@ void onFreeze(TransactionBody.Builder bodyBuilder) {
...
+    /**
+     * Freeze this transaction with the given client.
+     *
+     * @param client the client to freeze with
+     * @return this transaction
+     * @throws IllegalStateException if nodeId is not set
+     */
+    @Override
+    public NodeUpdateTransaction freezeWith(@Nullable Client client) {
+        if (nodeId == 0) {
```
0xivanov: what if we need to set the nodeId to actually be 0?

sdk/src/main/java/com/hedera/hashgraph/sdk/NodeDeleteTransaction.java
```diff
@@ -30,7 +31,7 @@
  */
 public class NodeDeleteTransaction extends Transaction<NodeDeleteTransaction> {
 
-    private long nodeId = 0;
+    private Long nodeId;
```
0xivanov, line 34: Need to add null checks for the getters, since it can result in null pointer. Same for node update

0xivanov (approved):
LGTM

PR #2456, opened 2025-09-05
fix: reties on topic unsubscribe

emiliyank:
**Description**:
Disable retries when customer unsubscribes from a topic.

**Related issue(s):**
Fixes https://github.com/hiero-ledger/hiero-sdk-java/issues/2453

**Checklist**

### DIFF
--- a/sdk/src/main/java/com/hedera/hashgraph/sdk/TopicMessageQuery.java
+++ b/sdk/src/main/java/com/hedera/hashgraph/sdk/TopicMessageQuery.java
@@ -320,11 +320,8 @@ public final class TopicMessageQuery {
 
             @Override
             public void onError(Throwable t) {
-                if (attempt >= maxAttempts || !retryHandler.test(t)) {
+                if (attempt >= maxAttempts || !retryHandler.test(t) || cancelledByClient.get()) {
                     errorHandler.accept(t, null);
-                    return;
-                }
-                if (cancelledByClient.get()) {
                     return;
                 }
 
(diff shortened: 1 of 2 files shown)


0xivanov (approved):

sdk/src/main/java/com/hedera/hashgraph/sdk/TopicMessageQuery.java
```diff
@@ -316,7 +320,7 @@ public void onNext(ConsensusTopicResponse consensusTopicResponse) {
 
             @Override
             public void onError(Throwable t) {
-                if (attempt >= maxAttempts || !retryHandler.test(t)) {
+                if (attempt >= maxAttempts || !retryHandler.test(t) || cancelledByClient.get()) {
```
steven-sheehy, line 323: I believe this doesn't fix the problem. We should not call retry handler or error handler if cancelled by client. This should be moved to a separate if statement that just returns.

0xivanov: If `cancelledByClient` we will always call the error handler and return. What is the problem with calling the error handler?

steven-sheehy: The problem is clearly mentioned in the original issue and is not addressed:

> The TopicMessageQuery logs a warning Error subscribing to topic 0.0.1043 during attempt #0. Waiting 500 ms before next attempt: CANCELLED: unsubscribe when the user explicitly unsubscribes from the topic. It also invokes the retry handler (which in our case logs another warning message) even though the call should definitely not be retried after the user requested cancellation.

The entire point of the issue was that we get unwanted error logs and your change still results in unwanted error logs.

0xivanov: I was left with the impression that the issue was that it retried. Will open a follow-up PR.

0xivanov: Here is the follow up: #2463

PR #2463, opened 2025-09-15
fix: warnings on topic unsubscribe

emiliyank:
**Description**:
Do not call errorHandler when topic cancelled by client.

**Related issue(s)**:

Fixes #2456

**Checklist**


sdk/src/testIntegration/java/com/hedera/hashgraph/sdk/test/integration/TopicMessageIntegrationTest.java
```diff
@@ -123,4 +124,51 @@ void canReceiveALargeTopicMessage() throws Exception {
...
+
+            Thread.sleep(3000);
+
+            var receivedMessage = new AtomicBoolean(false);
+            var retryWarningLogged = new AtomicBoolean(false);
+
+            var retryHandler = new java.util.function.Predicate<Throwable>() {
+                @Override
+                public boolean test(Throwable throwable) {
+                    retryWarningLogged.set(true);
```
steven-sheehy, line 147: We should do a similar check for the error handler to ensure it's not invoked for user cancel.

sdk/src/testIntegration/java/com/hedera/hashgraph/sdk/test/integration/TopicMessageIntegrationTest.java
```diff
@@ -123,4 +124,51 @@ void canReceiveALargeTopicMessage() throws Exception {
...
+        try (var testEnv = new IntegrationTestEnv(1)) {
+
+            var response = new TopicCreateTransaction()
+                    .setAdminKey(testEnv.operatorKey)
+                    .setTopicMemo("[e2e::TopicCreateTransaction]")
+                    .execute(testEnv.client);
+
+            var topicId = Objects.requireNonNull(response.getReceipt(testEnv.client).topicId);
+
+            Thread.sleep(3000);
```
steven-sheehy: I'm surprised this test class isn't flaky with all these sleeps. Instead of sleeps, you should really be polling the mirror node REST API for the create to show up first. Just a suggestion, can ignore.

sdk/src/testIntegration/java/com/hedera/hashgraph/sdk/test/integration/TopicMessageIntegrationTest.java
emiliyank: added such check

steven-sheehy (approved):

sdk/src/testIntegration/java/com/hedera/hashgraph/sdk/test/integration/TopicMessageIntegrationTest.java
emiliyank: Thanks! Seems that the  Thread.sleep(3000); is not needed at all in this case.

PR #2480, opened 2025-09-29
mirror node security port number in url

### LINKED ISSUE #2466: Make mirror node security protocol rely on port number
### Problem

Several SDK methods, such as AccountId.populateAccountEvmAddress and AccountId.populateAccountNum, currently reconstruct request URLs by slicing strings and prepending https://. This logic strips out ports and ignores the configured scheme, which breaks connectivity when clients use mirror nodes with custom ports or protocols.

### Solution

The SDK should respect the full URL (scheme + port + host) as defined in client.mirrorNetwork.
URL construction should be centralized in a network/endpoint abstraction, not repeated in individual methods.
Other SDK features should benefit automatically from consistent URL handling.

### Alternatives

_No response_

emiliyank:
**Description**:
Add a separate method to generate Rest Base URL.

Fixes #2466 

**Checklist**


sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
```diff
@@ -442,6 +442,20 @@ public synchronized List<String> getMirrorNetwork() {
         return mirrorNetwork.getNetwork();
     }
 
+    /**
+     * Build the REST base URL for the next healthy mirror node.
+     * Returns a string like `https://host[:port]/api/v1` or for local networks `http://localhost:5551/api/v1`.
+     */
+    String getMirrorRestBaseUrl(boolean isContractCall) {
+        try {
+            boolean isLocal = getLedgerId() == null;
```
0xivanov: we should not rely on the ledger id to tell us if we are on local network

sdk/src/main/java/com/hedera/hashgraph/sdk/EntityIdHelper.java
```diff
@@ -277,7 +276,7 @@ static CompletableFuture<Long> getAccountNumFromMirrorNodeAsync(Client client, S
      * @param client
      * @param num
      */
-    public static CompletableFuture<EvmAddress> getEvmAddressFromMirrorNodeAsync(Client client, long num) {
+    public static CompletableFuture<EvmAddress>     getEvmAddressFromMirrorNodeAsync(Client client, long num) {
```
0xivanov: this line looks not formatted

sdk/src/main/java/com/hedera/hashgraph/sdk/MirrorNode.java
```diff
@@ -36,4 +36,44 @@ protected String getAuthority() {
     BaseNodeAddress getKey() {
         return address;
     }
+
+    /**
+     * Build the REST base URL for this mirror node.
+     *
+     * @param isLocalNetwork when true, force HTTP and local default ports (5551/8545)
```
0xivanov: `isLocalNetwork` can be removed

sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
```diff
@@ -442,6 +442,20 @@ public synchronized List<String> getMirrorNetwork() {
         return mirrorNetwork.getNetwork();
     }
 
+    /**
+     * Build the REST base URL for the next healthy mirror node.
+     * Returns a string like `https://host[:port]/api/v1` or for local networks `http://localhost:5551/api/v1`.
+     */
+    String getMirrorRestBaseUrl(boolean isContractCall) {
```
0xivanov: Having this param specifically for the localhost + contract call scenario is too much, i think.
Maybe we can make the `MirrorContractCallQuery` to set the port to 8545 explicitly in this case and remove this param from here

sdk/src/main/java/com/hedera/hashgraph/sdk/Client.java
```diff
@@ -442,6 +442,20 @@ public synchronized List<String> getMirrorNetwork() {
         return mirrorNetwork.getNetwork();
     }
 
+    /**
+     * Build the REST base URL for the next healthy mirror node.
+     * Returns a string like `https://host[:port]/api/v1`.
+     * If the selected mirror node is a local host (localhost/127.0.0.1) returns `http://localhost:{5551|8545}/api/v1`.
+     */
+    String getMirrorRestBaseUrl() {
```
0xivanov: can be made public

sdk/src/main/java/com/hedera/hashgraph/sdk/MirrorNode.java
```diff
@@ -36,4 +36,42 @@ protected String getAuthority() {
...
+        } else if (port == 443) {
+            scheme = "https";
+        } else {
+            scheme = "https";
+        }
+
+        StringBuilder base = new StringBuilder();
+        base.append(scheme).append("://").append(host);
+        // Omit default ports
+        if (!((scheme.equals("http") && port == 80) || (scheme.equals("https") && port == 443))) {
```
0xivanov: what is the purpose of this if?

emiliyank: To skip the default ports - otherwise we will have for example `https://mirror.example.com:443/api/v1`

Issue #2468, opened 2025-09-18
Hiero hooks and an application to allowances


emiliyank:
### Problem
This HIP introduces **hooks**, programmable Hiero extension points that let users customize the behavior of their entities. The initial implementation focuses on **EVM hooks** and **account allowance hooks** as the first extension point.

Implement the needed SDK changes according to this design doc: https://github.com/hiero-ledger/sdk-collaboration-hub/pull/47

PR #2505, opened 2025-10-13
Hip 1195 add hiero hook transfers

emiliyank:
**Description**:
Add hooks to TransferTransaction

Fixes #2468 

**Checklist**

### DIFF
--- a/sdk/src/testIntegration/java/com/hedera/hashgraph/sdk/test/integration/TransferTransactionHooksIntegrationTest.java
+++ b/sdk/src/testIntegration/java/com/hedera/hashgraph/sdk/test/integration/TransferTransactionHooksIntegrationTest.java
@@ -164,91 +164,92 @@ class TransferTransactionHooksIntegrationTest {
         }
     }
 
-    //    @Test
-    //    @DisplayName(
-    //            "Given an account has an allowance hook for token transfers, when a TransferTransaction includes token
-    // transfers from that account, then the hook validates the token allowance and approves valid transfers")
-    //    void fungibleTokenTransferWithAllowanceHookSucceeds() throws Exception {
-    //        try (var testEnv = new IntegrationTestEnv(1)) {
-    //            var hookContractId = createContractId(testEnv);
-    //
-    //            var hookDetails = new HookCreationDetails(
-    //                    HookExtensionPoint.ACCOUNT_ALLOWANCE_HOOK, 1L, new LambdaEvmHook(hookContractId));
-    //
-    //            var receiverKey = PrivateKey.generateED25519();
-    //            var receiverId = new AccountCreateTransaction()
-    //                    .setKeyWithoutAlias(receiverKey)
-    //                    .setInitialBalance(new Hbar(2))
-    //                    .execute(testEnv.client)
-    //                    .getReceipt(testEnv.client)
-    //                    .accountId;
-    //
-    //            // Attach allowance hook to operator (sender)
-    //            new AccountUpdateTransaction()
-    //                    .setAccountId(receiverId)
-    //                    .addHookToCreate(hookDetails)
-    //                    .freezeWith(testEnv.client)
-    //                    .sign(receiverKey)
-    //                    .execute(testEnv.client)
-    //                    .getReceipt(testEnv.client);
-    //
-    //            // Create fungible token with operator as treasury
-    //            var tokenId = new TokenCreateTransaction()
-    //                    .setTokenName("FT-HOOK")
-    //                    .setTokenSymbol("FTH")
-    //                    .setTokenType(TokenType.FUNGIBLE_COMMON)
-    //                    .setDecimals(2)
-    //                    .setInitialSupply(10_000) // 100.00 units
-    //                    .setTreasuryAccountId(testEnv.operatorId)
-    //                    .setAdminKey(testEnv.operatorKey)
-    //                    .setSupplyKey(testEnv.operatorKey)
-    //                    .setKycKey(testEnv.operatorKey)
-    //                    .freezeWith(testEnv.client)
-    //                    .signWithOperator(testEnv.client)
-    //                    .execute(testEnv.client)
-    //                    .getReceipt(testEnv.client)
-    //                    .tokenId;
-    //
-    //            // Associate + KYC receiver
-    //            new TokenAssociateTransaction()
-    //                    .setAccountId(receiverId)
-    //                    .setTokenIds(List.of(tokenId))
-    //                    .freezeWith(testEnv.client)
-    //                    .sign(receiverKey)
-    //                    .execute(testEnv.client)
-    //                    .getReceipt(testEnv.client);
-    //
-    //            new TokenGrantKycTransaction()
-    //                    .setAccountId(receiverId)
-    //                    .setTokenId(tokenId)
-    //                    .execute(testEnv.client)
-    //                    .getReceipt(testEnv.client);
-    //
-    //            // Ensure the allowance hook is attached to the debited account (operator)
-    //            var hookDetails2 = new HookCreationDetails(
-    //                    HookExtensionPoint.ACCOUNT_ALLOWANCE_HOOK, 2L, new LambdaEvmHook(hookContractId));
-    //            new AccountUpdateTransaction()
-    //                    .setAccountId(testEnv.operatorId)
-    //                    .addHookToCreate(hookDetails2)
-    //                    .freezeWith(testEnv.client)
-    //                    .signWithOperator(testEnv.client)
-    //                    .execute(testEnv.client)
-    //                    .getReceipt(testEnv.client);
-    //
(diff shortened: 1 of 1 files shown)


sdk/src/main/java/com/hedera/hashgraph/sdk/AbstractTokenTransferTransaction.java
```diff
@@ -171,6 +171,33 @@ public T addApprovedTokenTransferWithDecimals(TokenId tokenId, AccountId account
...
+     * Add a non-approved fungible token transfer with an allowance hook.
+     *
+     * @param tokenId the token id
+     * @param accountId the account id (sender if amount negative, receiver if positive)
+     * @param value the amount
+     * @param hookCall the hook call to execute
+     * @param hookType PRE_TX_ALLOWANCE_HOOK or PRE_POST_TX_ALLOWANCE_HOOK
+     * @return {@code this}
+     */
+    public T addTokenTransferWithHook(
```
0xivanov: I think the logic in this method can replace `doAddTokenTransfer` and both `addTokenTransferWithHook` and `addTokenTransfer` call it. Maybe a bigger refactor can be made taking into account decimals and approvals as well. So we would have a function that can add transfers with hooks, approvals, decimals, etc.. And the top level callers can pass as parameters what they would like to set.

0xivanov: Also, no changes should be made to the abstract class since it would modify Airdrop transaction - which is not what we want.
Put everything in TransferTransaction.java

sdk/src/main/java/com/hedera/hashgraph/sdk/EvmHookCall.java
```diff
@@ -0,0 +1,91 @@
...
+                .build();
+    }
+
+    /**
+     * Create an EvmHookCall from a protobuf message.
+     *
+     * @param proto the protobuf EvmHookCall
+     * @return a new EvmHookCall instance
+     */
+    public static EvmHookCall fromProtobuf(com.hedera.hashgraph.sdk.proto.EvmHookCall proto) {
```
0xivanov:
```suggestion
    static EvmHookCall fromProtobuf(com.hedera.hashgraph.sdk.proto.EvmHookCall proto) {
```

sdk/src/main/java/com/hedera/hashgraph/sdk/HookCall.java
```diff
@@ -0,0 +1,142 @@
...
+        return builder.build();
+    }
+
+    /**
+     * Create a HookCall from a protobuf message.
+     *
+     * @param proto the protobuf HookCall
+     * @return a new HookCall instance
+     */
+    public static HookCall fromProtobuf(com.hedera.hashgraph.sdk.proto.HookCall proto) {
```
0xivanov:
```suggestion
   static HookCall fromProtobuf(com.hedera.hashgraph.sdk.proto.HookCall proto) {
```

sdk/src/main/java/com/hedera/hashgraph/sdk/HookEntityId.java
```diff
@@ -0,0 +1,125 @@
...
+        return builder.build();
+    }
+
+    /**
+     * Create a HookEntityId from a protobuf message.
+     *
+     * @param proto the protobuf HookEntityId
+     * @return a new HookEntityId instance
+     */
+    public static HookEntityId fromProtobuf(com.hedera.hashgraph.sdk.proto.HookEntityId proto) {
```
0xivanov:
```suggestion
    static HookEntityId fromProtobuf(com.hedera.hashgraph.sdk.proto.HookEntityId proto) {
```

sdk/src/main/java/com/hedera/hashgraph/sdk/HookId.java
```diff
@@ -0,0 +1,91 @@
...
+                .build();
+    }
+
+    /**
+     * Create a HookId from a protobuf message.
+     *
+     * @param proto the protobuf HookId
+     * @return a new HookId instance
+     */
+    public static HookId fromProtobuf(com.hedera.hashgraph.sdk.proto.HookId proto) {
```
0xivanov:
```suggestion
    static HookId fromProtobuf(com.hedera.hashgraph.sdk.proto.HookId proto) {
```

sdk/src/main/java/com/hedera/hashgraph/sdk/TransferTransaction.java
```diff
@@ -149,6 +177,83 @@ public TransferTransaction addApprovedHbarTransfer(AccountId accountId, Hbar val
...
+            NftHookType senderHookType,
+            NftHookType receiverHookType) {
+        Objects.requireNonNull(nftId, "nftId cannot be null");
+        Objects.requireNonNull(senderAccountId, "senderAccountId cannot be null");
+        Objects.requireNonNull(receiverAccountId, "receiverAccountId cannot be null");
+
+        HookCall senderCall = null;
+        HookCall receiverCall = null;
+
+        if (senderHookType == NftHookType.PRE_HOOK_SENDER || senderHookType == NftHookType.PRE_POST_HOOK_SENDER) {
```
0xivanov: do we actually need these 2 `if`s. I think we `switch` though the types in `toProtobuf`, don't we?

sdk/src/main/java/com/hedera/hashgraph/sdk/AbstractTokenTransferTransaction.java
```diff
@@ -171,6 +171,33 @@ public T addApprovedTokenTransferWithDecimals(TokenId tokenId, AccountId account
...
+            TokenId tokenId, AccountId accountId, long value, HookCall hookCall, HookType hookType) {
+        requireNotFrozen();
```
0xivanov: can require non null for the inputs

sdk/src/main/java/com/hedera/hashgraph/sdk/TransferTransaction.java
```diff
@@ -224,10 +329,31 @@ void initFromTransactionBody() {
         var body = sourceTransactionBody.getCryptoTransfer();
 
         for (var transfer : body.getTransfers().getAccountAmountsList()) {
```
0xivanov, line 378: in `initFromTransactionBody` there is no check for hook data for nfts and fungible token transfers - only for `HbarTransfer`.

0xivanov: for example - the newly added constructor for `TokenNftTransfer` is not used here

sdk/src/main/java/com/hedera/hashgraph/sdk/TransferTransaction.java
emiliyank: `TokenTransfer.toProtobuf()` only writes a hook if the matching call AND type for that side are both non-null, then it switches on that side’s type to pick the correct proto field. The two `if`s ensure we attach the single hookCall to the correct sides (sender or receiver) based on the provided types. Is it possible that only one of the sender or receiver hooks should be set?

sdk/src/main/java/com/hedera/hashgraph/sdk/HookType.java
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: Apache-2.0
+package com.hedera.hashgraph.sdk;
+
+/**
+ * Enum representing the types of hooks that can be attached to HBAR transfers.
+ * <p>
+ * These types correspond to the protobuf definitions for hook calls in the
+ * AccountAmount message, specifically the hook_call oneof field.
+ */
+public enum HookType {
```
0xivanov:
```suggestion
public enum FungibleHookType {
```

sdk/src/main/java/com/hedera/hashgraph/sdk/TransferTransaction.java
0xivanov: I think the method should have 2 hook types -> one for sender and one for receiver -> add them both to the NftTransfer object and then do the switch logic there

sdk/src/main/java/com/hedera/hashgraph/sdk/HookCall.java
```diff
@@ -0,0 +1,142 @@
...
+/**
+ * Specifies a call to a hook from within a transaction.
+ * <p>
+ * Often the hook's entity is implied by the nature of the call site. For example, when using an account allowance hook
+ * inside a crypto transfer, the hook's entity is necessarily the account whose authorization is required.
+ * <p>
+ * For future extension points where the hook owner is not forced by the context, we include the option to fully
+ * specify the hook id for the call.
+ */
+public class HookCall {
```
0xivanov: need to add unit test for these new classes

PR #2557, opened 2025-12-12
feat:initial logic and tests for changing node account id

### LINKED ISSUE #2464: Changing Node Account Ids
- Implement the test plan that validates the functionality for updating the `account_id` field in `NodeUpdateTransactionBody` as part of the Dynamic Address Book (DAB) enhancement. The feature allows node operators to change the account associated with their node.
- Reference: https://github.com/hiero-ledger/sdk-collaboration-hub/pull/53

emiliyank:
**Description**:
feat:initial logic and tests for changing node account id

**Related issue(s)**:

Fixes #2464 

**Checklist**

### DIFF
--- a/sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
+++ b/sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
@@ -342,6 +342,26 @@ abstract class Executable<SdkRequestT, ProtoRequestT extends MessageLite, Respon
     }
 
     /**
+     * Updates the client's network from the address book if a mirror network is configured.
+     * Logs any errors at TRACE level without throwing exceptions.
+     *
+     * @param client The client whose network should be updated
+     */
+    private void updateNetworkFromAddressBook(Client client) {
+        try {
+            if (client.getMirrorNetwork() != null && !client.getMirrorNetwork().isEmpty()) {
+                client.updateNetworkFromAddressBook();
+            }
+        } catch (Exception updateError) {
+            if (logger.isEnabledForLevel(LogLevel.TRACE)) {
+                logger.trace(
+                        "failed to update client address book after INVALID_NODE_ACCOUNT_ID: {}",
+                        updateError.getMessage());
+            }
+        }
+    }
+
+    /**
      * Execute this transaction or query
      *
      * @param client The client with which this will be executed.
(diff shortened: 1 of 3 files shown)


sdk/build.gradle.kts
```diff
@@ -90,6 +90,9 @@ tasks.register<Exec>("updateProto") {
 
 tasks.withType<Test>().configureEach {
     if (project.hasProperty("skipNodeUpdateTest")) {
-        filter { excludeTestsMatching("*NodeUpdateTransactionIntegrationTest") }
+        // filter { excludeTestsMatching("*NodeUpdateTransactionIntegrationTest") }
```
0xivanov: shall we remove the comment?

.github/workflows/build.yml
```diff
@@ -176,10 +176,10 @@ jobs:
 
       - name: Prepare Hiero Solo
         id: solo
-        uses: hiero-ledger/hiero-solo-action@dd0048139ef1e40fd6067f01bf94eb42a67294f4 # v0.15
+        uses: hiero-ledger/hiero-solo-action@fbca3e7a99ce9aa8a250563a81187abe115e0dad # v0.16
         with:
           installMirrorNode: true
-          hieroVersion: v0.69.0-alpha.1
+          hieroVersion: "v0.69.0-alpha.1"
```
0xivanov: do we need the quotation marks?

sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
```diff
@@ -446,6 +461,7 @@ public O execute(Client client, Duration timeout) throws TimeoutException, Prech
                         currentTimeout = Duration.between(Instant.now(), timeoutTime);
                         delay(Math.min(currentTimeout.toMillis(), grpcRequest.getDelay()));
                     }
+                    advanceRequest(); // Advance to next node before retrying
```
0xivanov, line 484: This line will change the node even when the node that we work with is healthy. For example if it gives status BUSY there is no need to change it because these throttles are network scoped and all other nodes would give this response.

sdk/src/main/java/com/hedera/hashgraph/sdk/Executable.java
```diff
@@ -679,10 +695,9 @@ Node getNodeForExecute(int attempt) {
     private ProtoRequestT getRequestForExecute() {
         var request = makeRequest();
 
-        // advance the internal index
-        // non-free queries and transactions map to more than 1 actual transaction and this will cause
-        // the next invocation of makeRequest to return the _next_ transaction
-        advanceRequest();
+        // NOTE: advanceRequest() is now called explicitly in the retry logic
```
0xivanov: can remove these comments in my opinion, because it explains why we've removed something, but in the future one cannot see what actually has been deleted

PR #2593, opened 2026-02-04
feat(TCK): implement ContractInfoQuery

### LINKED ISSUE #2592: feat(TCK): Implement ContractInfoQuery
### Problem

Implement ContractInfoQuery

### Solution

Implement ContractInfoQuery

### Alternatives

_No response_

emiliyank:
**Description**: 
feat(TCK): implement ContractInfoQuery

**Related issue(s)**:
Fixes #2592

### DIFF
--- a/tck/src/main/java/com/hedera/hashgraph/tck/methods/sdk/response/ContractResponse.java
+++ b/tck/src/main/java/com/hedera/hashgraph/tck/methods/sdk/response/ContractResponse.java
@@ -4,41 +4,34 @@ package com.hedera.hashgraph.tck.methods.sdk.response;
 import com.hedera.hashgraph.sdk.Status;
 import java.util.LinkedHashMap;
 import java.util.Map;
-import lombok.Data;
 import net.minidev.json.JSONAware;
 import net.minidev.json.JSONObject;
 
-@Data
-public class ContractResponse {
-    private final String contractId;
-    private final Status status;
-
-    @Data
-    public static class ContractInfoQueryResponse implements JSONAware {
-        private final String contractId;
-        private final String accountId;
-        private final String contractAccountId;
-        private final String adminKey;
-        private final String expirationTime;
-        private final String autoRenewPeriod;
-        private final String autoRenewAccountId;
-        private final String storage;
-        private final String contractMemo;
-        private final String balance;
-        private final Boolean isDeleted;
-        private final String maxAutomaticTokenAssociations;
-        private final String ledgerId;
-        private final StakingInfoResponse stakingInfo;
-
-        @Data
-        public static class StakingInfoResponse implements JSONAware {
-            private final Boolean declineStakingReward;
-            private final String stakePeriodStart;
-            private final String pendingReward;
-            private final String stakedToMe;
-            private final String stakedAccountId;
-            private final String stakedNodeId;
-
+public record ContractResponse(String contractId, Status status) {
+    public record ContractInfoQueryResponse(
+            String contractId,
+            String accountId,
+            String contractAccountId,
+            String adminKey,
+            String expirationTime,
+            String autoRenewPeriod,
+            String autoRenewAccountId,
+            String storage,
+            String contractMemo,
+            String balance,
+            Boolean isDeleted,
+            String maxAutomaticTokenAssociations,
+            String ledgerId,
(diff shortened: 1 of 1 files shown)


tck/src/main/java/com/hedera/hashgraph/tck/methods/sdk/param/contract/InfoQueryContractParams.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * InfoQueryContractParams for contract info query method
+ */
+@SuppressWarnings("OptionalUsedAsFieldOrParameterType")
+@Getter
+@AllArgsConstructor
+@NoArgsConstructor
+public class InfoQueryContractParams extends JSONRPC2Param {
+    private Optional<String> contractId;
```
hendrikebbers, line 20: We should not use Optional as field. See https://stackoverflow.com/questions/26327957/should-java-8-getters-return-optional-type/26328555#26328555

tck/src/main/java/com/hedera/hashgraph/tck/methods/sdk/param/contract/InfoQueryContractParams.java
```diff
@@ -0,0 +1,37 @@
...
+    private Optional<String> queryPayment;
+    private Optional<String> maxQueryPayment;
+    private String sessionId;
+
+    @Override
+    public InfoQueryContractParams parse(Map<String, Object> jrpcParams) throws Exception {
```
hendrikebbers, line 26: Missing null check of jrpcParams

tck/src/main/java/com/hedera/hashgraph/tck/methods/sdk/response/ContractResponse.java
```diff
@@ -2,10 +2,93 @@
...
+import net.minidev.json.JSONAware;
+import net.minidev.json.JSONObject;
 
 @Data
 public class ContractResponse {
     private final String contractId;
     private final Status status;
+
+    @Data
+    public static class ContractInfoQueryResponse implements JSONAware {
```
hendrikebbers: could be a record

tck/src/main/java/com/hedera/hashgraph/tck/methods/sdk/response/ContractResponse.java
```diff
@@ -2,10 +2,93 @@
...
+        private final String storage;
+        private final String contractMemo;
+        private final String balance;
+        private final Boolean isDeleted;
+        private final String maxAutomaticTokenAssociations;
+        private final String ledgerId;
+        private final StakingInfoResponse stakingInfo;
+
+        @Data
+        public static class StakingInfoResponse implements JSONAware {
```
hendrikebbers: could be a record

tck/src/main/java/com/hedera/hashgraph/tck/methods/sdk/response/ContractResponse.java
```diff
@@ -2,10 +2,93 @@
...
+                }
+                if (stakedNodeId != null) {
+                    json.put("stakedNodeId", stakedNodeId);
+                }
+                return json.toJSONString();
+            }
+        }
+
+        @Override
+        public String toJSONString() {
```
hendrikebbers, line 61: We need to add unit tests for the custom JSON parsing

tck/src/main/java/com/hedera/hashgraph/tck/methods/sdk/param/contract/InfoQueryContractParams.java
0xivanov: Since some tck params can be missing, I would prefer to use optional instead of null checks everywhere

tck/src/main/java/com/hedera/hashgraph/tck/methods/sdk/param/contract/InfoQueryContractParams.java
0xivanov: In this case, the null check isn’t necessary because this method is an override from the JSON-RPC library contract. 
Also, this code lives in our testing framework rather than production code. We’re intentionally keeping it lightweight and avoiding additional guard clauses that would introduce unnecessary overhead

PR #2614, opened 2026-02-12
fix: block field and related logic for MirrorNodeContractQuery

### LINKED ISSUE #2612: Unable to query historical blocks via MirrorNodeContractQuery
### Description

The `MirrorNodeContractQuery` is unable to query historical blocks. The reason is because it's using an incorrect `blockNumber` field instead of the correct `block` field. The mirror node is ignoring the unknown field and defaulting to latest block.

Note that care should be taken to not regress this by just changing it to `block`. Doing that will then always send block zero if not set by user since the field is a primitive field that is always sent. Instead, make it a `Long` that is null so that if not set it defaults to empty which is latest. Document this behavior in the javadocs.

https://github.com/hiero-ledger/hiero-sdk-java/blob/c213bf8f55009822dc9c6ceac1c31d2930eb956c/sdk/src/main/java/com/hedera/hashgraph/sdk/MirrorNodeContractQuery.java#L313

Please also check other SDK languages for the same problem.

### Steps to reproduce

Run MirrorNodeContractQuery with setBlockNumber(someLong) and get latest results instead of historical.

### Additional context

_No response_

### Hedera network

other

### Version

2.66.0

### Operating system

None

mustafauzunn:
**Description**:

The `MirrorNodeContractQuery` cannot query historical blocks. The reason is that it's using an incorrect `blockNumber` field instead of the correct `block` field. The mirror node is ignoring the unknown field and defaulting to the latest block.

**Related issue(s)**:

Fixes #2612 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/sdk/src/main/java/com/hedera/hashgraph/sdk/MirrorNodeContractQuery.java
+++ b/sdk/src/main/java/com/hedera/hashgraph/sdk/MirrorNodeContractQuery.java
@@ -202,7 +202,7 @@ public abstract class MirrorNodeContractQuery<T extends MirrorNodeContractQuery<
         return self();
     }
 
-    public Long getBlock() {
+    public Long getBlockNumber() {
         return this.block;
     }
 
@@ -214,7 +214,7 @@ public abstract class MirrorNodeContractQuery<T extends MirrorNodeContractQuery<
      * @param block the block number at which to simulate the contract call
      * @return {@code this}
      */
-    public T setBlock(Long block) {
+    public T setBlockNumber(Long block) {
         this.block = block;
         return self();
     }
--- a/examples/src/main/java/com/hedera/hashgraph/sdk/examples/MirrorNodeContractQueriesExample.java
+++ b/examples/src/main/java/com/hedera/hashgraph/sdk/examples/MirrorNodeContractQueriesExample.java
@@ -117,7 +117,7 @@ public class MirrorNodeContractQueriesExample {
                 .setContractId(contractId)
                 .setSender(client.getOperatorAccountId())
                 .setGasLimit(30_000)
-                .setBlock(10000L)
+                .setBlockNumber(10000L)
                 .setGasPrice(1234)
                 .setFunction("getMessage")
                 .execute(client);
--- a/sdk/src/test/java/com/hedera/hashgraph/sdk/MirrorNodeContractQueryTest.java
+++ b/sdk/src/test/java/com/hedera/hashgraph/sdk/MirrorNodeContractQueryTest.java
@@ -91,11 +91,11 @@ class MirrorNodeContractQueryTest {
     @Test
     void testSetAndGetBlockNumber() {
         long blockNumber = 123456;
-        mirrorNodeContractEstimateGasQuery.setBlock(blockNumber);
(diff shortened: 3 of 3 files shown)


sdk/src/main/java/com/hedera/hashgraph/sdk/MirrorNodeContractQuery.java
```diff
@@ -201,20 +202,20 @@ public T setGasPrice(long gasPrice) {
         return self();
     }
 
-    public long getBlockNumber() {
-        return this.blockNumber;
+    public long getBlock() {
+        return this.block;
```
hendrikebbers, line 206: We have an auto boxing problem here since the field is defined as `Long` (allows `null` and initialized as `null`) but the method returns primitive `long`.

sdk/src/main/java/com/hedera/hashgraph/sdk/MirrorNodeContractQuery.java
```diff
@@ -240,7 +241,7 @@ protected long estimate(Client client) throws ExecutionException, InterruptedExc
      */
     protected String call(Client client) throws ExecutionException, InterruptedException {
         fillEvmAddresses();
-        var blockNum = this.blockNumber == 0 ? "latest" : String.valueOf(this.blockNumber);
+        var blockNum = this.block != null ? String.valueOf(this.block) : null;
```
hendrikebbers: Let's avoid to use `var`. especially in those conversations  it is misleading about the type.

sdk/src/main/java/com/hedera/hashgraph/sdk/MirrorNodeContractQuery.java
```diff
@@ -265,7 +266,7 @@ private CompletableFuture<Long> getEstimateGasFromMirrorNodeAsync(Client client)
                 .thenApply(MirrorNodeContractQuery::parseHexEstimateToLong);
     }
 
-    private CompletableFuture<String> executeMirrorNodeRequest(Client client, String blockNumber, boolean estimate) {
+    private CompletableFuture<String> executeMirrorNodeRequest(Client client, String block, boolean estimate) {
```
hendrikebbers: Why do we use String for it if it is only used with numeric input?

sdk/src/main/java/com/hedera/hashgraph/sdk/MirrorNodeContractQuery.java
mustafauzunn: Logic refactored

sdk/src/main/java/com/hedera/hashgraph/sdk/MirrorNodeContractQuery.java
mustafauzunn: Logic refactored

sdk/src/main/java/com/hedera/hashgraph/sdk/MirrorNodeContractQuery.java
mustafauzunn: Logic refactored

0xivanov (approved):
LGTM

PR #2638, opened 2026-02-25
build: fix dependencies issues caused by json-snapshot / update Gradle

jjohannes:
This PR fixes the currently failing **Snyk Scan** step.

Main issue with failing updates are fishy dependency definitions in [json-snapshot ](https://github.com/json-snapshot/json-snapshot.github.io) that we did not discover before.

**Build Updated**:

- org.hiero.gradle.build --> 0.7.4
- Gradle --> 9.3.1
- Android example: com.android.application --> 9.0.1 (Kotlin is now included)

**Dependency Updates + Fixes**
- Fixup metadata of `io.github.json-snapshot:json-snapshot` (problem discovered in #2632)
- Update Spring Boot and JUnit to latest versions (supersedes #2632)

In a follow we'll move the patch code into the plugins – https://github.com/hiero-ledger/hiero-gradle-conventions/pull/422 – and can then remove it from here with the next update.

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -1,4 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
+import org.gradlex.jvm.dependency.conflict.resolution.JvmDependencyConflictsExtension
+
 plugins { id("org.hiero.gradle.build") version "0.7.4" }
 
 rootProject.name = "hedera-sdk-java"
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -186,7 +186,7 @@ jobs:
         run: ./gradlew assemble
 
       - name: Code Quality Checks
-        run: ./gradlew qualityCheck :examples:qualityCheck --continue --rerun-tasks
+        run: ./gradlew qualityCheck :examples:qualityCheck --continue
```
jjohannes, line 189: This was a workaround we can remove now to speed things up a bit

PR #2654, opened 2026-03-13
feat: HIP-1137

### LINKED ISSUE #2644: Implement HIP-1137 — Block Node discoverability via on-chain registry
## Summary

The Hiero Java SDK does not currently support the registered node registry introduced by [HIP-1137](https://github.com/hiero-ledger/hiero-improvement-proposals/blob/main/HIP/hip-1137.md). This issue tracks the implementation of full SDK support for creating, updating, deleting, and discovering registered nodes (Block Nodes, mirror nodes, and RPC relays) via on-chain data.

The implementation should follow the [SDK design document](https://github.com/hiero-ledger/sdk-collaboration-hub/blob/hip-1137/proposals/hips/hip-1137.md) and align with the patterns established by the existing consensus node transactions (`NodeCreateTransaction`, `NodeUpdateTransaction`, `NodeDeleteTransaction`).

## Problem

HIP-1137 adds three new transactions to the `AddressBookService` and extends several existing types. Without SDK support, developers must resort to raw protobuf construction to interact with the registered node registry.

Key gaps:
- No `RegisteredNodeCreateTransaction`, `RegisteredNodeUpdateTransaction`, or `RegisteredNodeDeleteTransaction` classes
- No `RegisteredServiceEndpoint` hierarchy (`BlockNodeServiceEndpoint`, `MirrorNodeServiceEndpoint`, `RpcRelayServiceEndpoint`)
- No `BlockNodeApi` enum
- `TransactionReceipt` does not expose the `registeredNodeId` field
- `NodeCreateTransaction` and `NodeUpdateTransaction` do not support the new `associatedRegisteredNodes` field
- No `RegisteredNode`, `RegisteredNodeAddressBook`, or `RegisteredNodeAddressBookQuery` types

### New transaction types

- **`RegisteredNodeCreateTransaction`** — creates a registered node with an `adminKey`, optional `description`, optional `nodeAccountId`, and a list of service endpoints (1–50). On success the receipt contains the network-assigned `registeredNodeId`.
- **`RegisteredNodeUpdateTransaction`** — updates an existing registered node by `registeredNodeId`. Supports changing `adminKey` (requires both old and new key signatures), `description`, `nodeAccountId`, and replacing the service endpoint list.
- **`RegisteredNodeDeleteTransaction`** — removes a registered node by `registeredNodeId`. Must be signed by the node's `adminKey` or authorized by network governance.

All three transactions must be schedulable via `ScheduleCreateTransaction`.

### New data types

- **`BlockNodeApi`** enum — `OTHER`, `STATUS`, `PUBLISH`, `SUBSCRIBE_STREAM`, `STATE_PROOF`.
- **`RegisteredServiceEndpoint`** — abstract base with `ipAddress` (bytes) or `domainName` (string), `port`, and `requiresTls`. Three concrete subtypes:
  - **`BlockNodeServiceEndpoint`** — adds `endpointApi: BlockNodeApi`
  - **`MirrorNodeServiceEndpoint`** — empty subtype (future-proofing)
  - **`RpcRelayServiceEndpoint`** — empty subtype (future-proofing)
- **`RegisteredNode`** — immutable representation of a registered node as stored in network state.
- **`RegisteredNodeAddressBook`** — collection of `RegisteredNode` objects.

### New query type

- **`RegisteredNodeAddressBookQuery`** — queries the mirror node for registered nodes and returns a `RegisteredNodeAddressBook`. Implementation should be deferred until the mirror node API is available, but the class skeleton should be defined.

### Updates to existing types

- **`TransactionReceipt`** — add nullable `registeredNodeId` field.
- **`NodeCreateTransaction`** — add `associatedRegisteredNodes` list and `addAssociatedRegisteredNode(long registeredNodeId)`.
- **`NodeUpdateTransaction`** — add nullable `associatedRegisteredNodes` list, `addAssociatedRegisteredNode(long registeredNodeId)`, and `clearAssociatedRegisteredNodes()`. The protobuf uses a wrapper message for three-state semantics (not set / empty list / non-empty list).

### Wiring

- Register the new transaction types in the `Transaction` dispatch logic (`Transaction.java`) for `fromBytes()` deserialization.
- Map to protobuf `DataCase` values and schedulable body fields.

### Patterns to follow

The existing `NodeCreateTransaction` / `NodeUpdateTransaction` / `NodeDeleteTransaction` in `sdk/src/main/java/com/hedera/hashgraph/sdk/` serve as the primary reference. Each new registered node transaction should:

- Extend the base `Transaction<T>` class
- Implement `onFreeze(TransactionBody.Builder)` and `fromProtobuf(TransactionBody)` methods
- Support scheduling via `ScheduleCreateTransaction`

### Endpoint hierarchy

The design document specifies inheritance-based endpoints. In Java this maps to:

- An abstract `RegisteredServiceEndpoint` class holding shared fields (`ipAddress`, `domainName`, `port`, `requiresTls`)
- `BlockNodeServiceEndpoint` extending the base with `endpointApi`
- `MirrorNodeServiceEndpoint` and `RpcRelayServiceEndpoint` as currently empty subtypes

Each subtype needs `fromProtobuf()` / `toProtobuf()` round-trip support, following the pattern in `Endpoint.java`.

### Key files to create or modify

**New files** (under `sdk/src/main/java/com/hedera/hashgraph/sdk/`):
- `RegisteredNodeCreateTransaction.java`
- `RegisteredNodeUpdateTransaction.java`
- `RegisteredNodeDeleteTransaction.java`
- `RegisteredServiceEndpoint.java`
- `BlockNodeServiceEndpoint.java`
- `MirrorNodeServiceEndpoint.java`
- `RpcRelayServiceEndpoint.java`
- `BlockNodeApi.java`
- `RegisteredNode.java`
- `RegisteredNodeAddressBook.java`
- `RegisteredNodeAddressBookQuery.java`

**Existing files to update:**
- `TransactionReceipt.java` — add `registeredNodeId`
- `NodeCreateTransaction.java` — add `associatedRegisteredNodes`
- `NodeUpdateTransaction.java` — add `associatedRegisteredNodes` with three-state wrapper semantics
- `Transaction.java` — register new types in the deserialization dispatch
- Proto definitions in `sdk/src/main/proto/`

### Protobuf dependencies

The implementation depends on new protobuf definitions from HIP-1137:
- `registered_node_create.proto`
- `registered_node_update.proto`
- `registered_node_delete.proto`
- `registered_service_endpoint.proto` (or equivalent)
- Updated `transaction_body.proto` (fields 78–80)
- Updated `schedulable_transaction_body.proto` (fields 49–51)
- Updated `transaction_receipt.proto` (field 16)

### Testing strategy

1. **Unit tests** — verify serialization round-trips for all new types, field validation, and getter/setter correctness.
2. **Integration tests** — execute the full registered node lifecycle against a test network:
   - Create a registered node with various endpoint types and verify the receipt contains a `registeredNodeId`
   - Update the node's description, endpoints, and admin key
   - Associate a registered node with a consensus node
   - Delete the registered node
   - Verify failure cases (missing admin key, empty endpoints, non-existent node ID, already-deleted node)
3. **TCK alignment** — corresponding test cases should be defined in the [TCK repository](https://github.com/hiero-ledger/hiero-sdk-tck) per the design document's 18-point test plan.

## Acceptance Criteria

- [ ] Implement `RegisteredNodeCreateTransaction`, `RegisteredNodeUpdateTransaction`, and `RegisteredNodeDeleteTransaction` following existing transaction patterns
- [ ] Implement the `RegisteredServiceEndpoint` hierarchy (`BlockNodeServiceEndpoint`, `MirrorNodeServiceEndpoint`, `RpcRelayServiceEndpoint`) and the `BlockNodeApi` enum
- [ ] Implement `RegisteredNode` and `RegisteredNodeAddressBook` data types
- [ ] Define the `RegisteredNodeAddressBookQuery` class skeleton
- [ ] Update `TransactionReceipt` to expose `registeredNodeId`
- [ ] Update `NodeCreateTransaction` and `NodeUpdateTransaction` with `associatedRegisteredNodes` support
- [ ] Register new transaction types in `Transaction` dispatch
- [ ] Ensure all three new transactions are schedulable
- [ ] Include unit tests for serialization, field validation, and getter/setter correctness
- [ ] Include integration tests covering the registered node lifecycle
- [ ] Maintain backwards compatibility with existing APIs
- [ ] Pass all CI checks

## References

- **HIP-1137 specification:** https://github.com/hiero-ledger/hiero-improvement-proposals/blob/main/HIP/hip-1137.md
- **SDK design document:** https://github.com/hiero-ledger/sdk-collaboration-hub/blob/hip-1137/proposals/hips/hip-1137.md
- **HIP discussion:** https://github.com/hiero-ledger/hiero-improvement-proposals/pull/1137
- **Existing consensus node transaction patterns:**
  - `sdk/src/main/java/com/hedera/hashgraph/sdk/NodeCreateTransaction.java`
  - `sdk/src/main/java/com/hedera/hashgraph/sdk/NodeUpdateTransaction.java`
  - `sdk/src/main/java/com/hedera/hashgraph/sdk/NodeDeleteTransaction.java`
- **Existing endpoint pattern:**
  - `sdk/src/main/java/com/hedera/hashgraph/sdk/Endpoint.java`

manishdait:
**Description**:
This PR introduce HIP 1137

**Related issue(s)**:

Fixes #2644

**Notes for reviewer**:

**Checklist**


sdk/src/testIntegration/java/com/hedera/hashgraph/sdk/test/integration/RegisteredNodeCreateTransactionIntegrationTest.java
```diff
@@ -0,0 +1,69 @@
...
+                    .setDomainName("test.block.com")
+                    .setPort(443)
+                    .setEndpointApi(BlockNodeApi.STATUS));
+
+            var response = new RegisteredNodeCreateTransaction()
+                    .setAdminKey(key)
+                    .setDescription("test description")
+                    .setServiceEndpoints(serviceEndpoints)
+                    .freezeWith(testEnv.client)
+                    .execute(testEnv.client);
```
mustafauzunn, line 111: We need to sign this transaction.
Please look other tests and integration tests for missing signing

manishdait: Done

manishdait:
Hi @mustafauzunn, Is there any update on the Mirrornode Rest endpoint for registerdNode

mustafauzunn:
Hi @manishdait , you can follow this EPIC https://github.com/hiero-ledger/hiero-mirror-node/issues/12932
The endpoint task is in progress

mustafauzunn:
Hi @manishdait 
we have some changes to the hip. Please check  https://github.com/hiero-ledger/sdk-collaboration-hub/pull/167

manishdait:
@mustafauzunn, Thanks for sharing

mustafauzunn:
Hi @manishdait 
It seems like mirror node changes are ready https://github.com/hiero-ledger/hiero-mirror-node/issues/12932

manishdait:
Hi @mustafauzunn, I updated the `hieroVersion` in the solo-action to `0.73.0` for testing, but the Hook-related e2e tests are failing. I try to run them on my local `solo` setup and on `testnet` but the Hook related issues still fails with `HOOK_NOT_ENABLED` status. At the moment, I’m not sure whether this is caused by the` hieroVersion` change or an issue with the SDK’s hook tests.

Here is the log for `AccountCreateTransactionHooksIntegrationTest` integration test, I run using the testnet
```
> Task :sdk:testIntegration
  com.hedera.hashgraph.sdk.test.integration.AccountCreateTransactionHooksIntegrationTest ✘ Given AccountCreateTransaction with basic EVM hook, when executed, then receipt status is SUCCESS (15.2s)
                                                                                                                                                                                                                             
    com.hedera.hashgraph.sdk.PrecheckStatusException: Hedera transaction `0.0.6105114@1777016676.272000651` failed pre-check with the status `HOOKS_NOT_ENABLED`
        at app/com.hedera.hashgraph.sdk.test.integration@2.70.0/com.hedera.hashgraph.sdk.test.integration.AccountCreateTransactionHooksIntegrationTest.accountCreateWithBasicLambdaHookSucceeds(AccountCreateTransactionHooksIntegrationTest.java:39)
                                                                                                                                                                                                                             

  0 passing (19.2s)
  1 pending                                                                                                                                                                                                                  
  1 failing                                                                                                                                                                                                                  


2 tests completed, 1 failed, 1 skipped

> Task :sdk:testIntegration FAILED

FAILURE: Build failed with an exception.

```

mustafauzunn:
Hi @manishdait you can disable hooks tests for now. Hooks are disabled on 0.73.0

sdk/src/main/java/com/hedera/hashgraph/sdk/NodeCreateTransaction.java
```diff
@@ -396,6 +398,53 @@ public NodeCreateTransaction setGrpcWebProxyEndpoint(@Nullable Endpoint grpcWebP
...
+    }
+
+    /**
+     * Add a registered nodes operated by the same entity as this node.
+     * @param associatedRegisteredNode the associated registered node.
+     * @return {@code this}
+     */
+    public NodeCreateTransaction addAssociatedRegisteredNode(long associatedRegisteredNode) {
+        requireNotFrozen();
+        if (associatedRegisteredNodes.size() >= 20) {
```
manishdait: Hi @mustafauzunn should we remove this check and let it handle by the consensus node to throw the `PrecheckError` with `MAX_REGISTERED_NODES_EXCEEDED`. also similar with other

.github/workflows/build.yml
```diff
@@ -176,7 +176,7 @@ jobs:
 
       - name: Prepare Hiero Solo
         id: solo
-        uses: hiero-ledger/hiero-solo-action@328bc84c3b00a990a151418144fd682a4eb76ea6 # v0.17
+        uses: hiero-ledger/hiero-solo-action@4d42a74e8e644a2753f3bb7a2afa429305375b14 # v0.16
```
rwalworth: **issue:** This change downgrades `hiero-solo-action` from v0.17 back to v0.16, but `main` has already been bumped to v0.19 in #2719. As-is this will conflict on rebase, and the downgrade itself looks unintentional (the second occurrence in this file was left at the newer pinned SHA). Suggest rebasing onto current `main` and keeping the v0.19 pin everywhere - only the `hieroVersion` / `mirrorNodeVersion` bump on lines ~258 is needed for HIP-1137.

sdk/src/main/java/com/hedera/hashgraph/sdk/NodeUpdateTransaction.java
```diff
@@ -455,6 +469,62 @@ public NodeUpdateTransaction setGrpcWebProxyEndpoint(@Nullable Endpoint grpcWebP
...
+
+    /**
+     * Clear all registered nodes associated with this node.
+     * @return {@code this}
+     */
+    public NodeUpdateTransaction clearAssociatedRegisteredNode() {
+        requireNotFrozen();
+        associatedRegisteredNodes = new ArrayList<>();
+        return this;
+    }
```
rwalworth, line 515: **issue:** The HIP-1137 design doc spells this method as plural - `clearAssociatedRegisteredNodes()`:

```
NodeUpdateTransaction clearAssociatedRegisteredNodes()
```

This should be fixed before merge to prevent breaking changes or deprecations. Same suggestion for the new `clearServiceEndpoint()` method below - `clearServiceEndpoints()` reads better and matches the existing `serviceEndpoints` field name.

sdk/src/main/java/com/hedera/hashgraph/sdk/NodeUpdateTransaction.java
```diff
@@ -455,6 +469,62 @@ public NodeUpdateTransaction setGrpcWebProxyEndpoint(@Nullable Endpoint grpcWebP
...
+     * consensus node.
+     * <p>
+     * This field is OPTIONAL and MAY be empty.<br/>
+     * This field MUST NOT contain more than twenty(20) entries.<br/>
+     * Every entry in this list MUST be a valid `registered_node_id` for a
+     * current registered node.
+     *
+     * @param associatedRegisteredNodes list of associated registered node.
+     * @return {@code this}
+     * @throws IllegalArgumentException if the list is empty or contains more than 8 endpoints
```
rwalworth: **issue:** This Javadoc looks copy-pasted from somewhere else:

The method doesn't actually validate either condition, "endpoints" is the wrong noun (should be "registered nodes"), and the limit per HIP-1137 is 20, not 8. Suggest removing this `@throws` line (or updating the body of the doc to reference the 20-entry limit and `MAX_REGISTERED_NODES_EXCEEDED`, but no `@throws` since it's enforced server-side).

sdk/src/main/java/com/hedera/hashgraph/sdk/NodeUpdateTransaction.java
```diff
@@ -316,6 +320,16 @@ public NodeUpdateTransaction addServiceEndpoint(Endpoint serviceEndpoint) {
...
 
+    /**
+     * Clear serviceEndpoint lists.
+     * @return {@code this}
+     */
+    public NodeUpdateTransaction clearServiceEndpoint() {
+        requireNotFrozen();
+        serviceEndpoints.clear();
+        return this;
+    }
```
rwalworth: **question (non-blocking):** Was the `clearServiceEndpoint()` method intentional for this PR? It's not part of the HIP-1137 design doc (which only adds `associatedRegisteredNodes` to `NodeUpdateTransaction`), and it changes the behavior of the consensus `NodeUpdateTransaction`, not the new `RegisteredNodeUpdateTransaction`.

If it's needed for one of the integration tests, that's fine - happy to keep it, but the name should be `clearServiceEndpoints()` and ideally a one-liner mentioning HIP-1137 in the Javadoc so future readers understand why a non-HIP-1137 method landed in this PR. Otherwise, consider removing it and shipping in a separate PR with its own issue.

sdk/src/main/java/com/hedera/hashgraph/sdk/RegisteredNodeCreateTransaction.java
```diff
@@ -0,0 +1,209 @@
...
+ * address book.
+ *<p>
+ * This transaction, once complete, SHALL add a new registered node to the
+ * network state.
+ * The new registered node SHALL be visible and discoverable upon
+ * completion of this transaction.
+ */
+public class RegisteredNodeCreateTransaction extends Transaction<RegisteredNodeCreateTransaction> {
+    private Key adminKey;
+    private String description = "";
```
rwalworth: **issue (non-blocking):** The design doc declares `description` as `@@nullable string`, but here it defaults to `""` and `setDescription` rejects null. The result is that every transaction sends `description = ""` on the wire, even when the caller never set one - there's no way to send "description not set".

Compare this with `RegisteredNodeUpdateTransaction` which correctly treats it as nullable. Suggest making this nullable and only writing the protobuf field when non-null:

```java
@Nullable private String description;
...
if (description != null) {
    builder.setDescription(description);
}
```

sdk/src/main/java/com/hedera/hashgraph/sdk/RegisteredNodeUpdateTransaction.java
```diff
@@ -0,0 +1,256 @@
...
+     */
+    public RegisteredNodeUpdateTransaction setServiceEndpoints(List<RegisteredServiceEndpoint> serviceEndpoints) {
+        this.requireNotFrozen();
+        Objects.requireNonNull(serviceEndpoints, "serviceEndpoints cannot be null");
+
+        for (RegisteredServiceEndpoint serviceEndpoint : serviceEndpoints) {
+            RegisteredServiceEndpoint.validateNoIpAndDomain(serviceEndpoint);
+        }
+
+        this.serviceEndpoints = serviceEndpoints;
```
rwalworth: **issue (non-blocking):** `setServiceEndpoints` here stores the caller's list reference directly, but the corresponding method on `RegisteredNodeCreateTransaction` defensively copies (`new ArrayList<>(serviceEndpoints)`). That asymmetry means a caller who reuses their list builder will get different behavior between create and update, and an external mutation can corrupt a frozen transaction.

```diff
- this.serviceEndpoints = serviceEndpoints;
+ this.serviceEndpoints = new ArrayList<>(serviceEndpoints);
```

sdk/src/main/java/com/hedera/hashgraph/sdk/RegisteredNodeUpdateTransaction.java
```diff
@@ -0,0 +1,256 @@
...
+     *
+     * @param registeredNodeId the registered node identifier.
+     * @return {@code this}
+     * @throws IllegalArgumentException if registeredNodeId is negative
+     */
+    public RegisteredNodeUpdateTransaction setRegisteredNodeId(long registeredNodeId) {
+        this.requireNotFrozen();
+        this.registeredNodeId = registeredNodeId;
+        return this;
+    }
```
rwalworth, line 91: **nitpick:** The Javadoc says `@throws IllegalArgumentException if registeredNodeId is negative` but the body never validates. Either drop the `@throws` line or add the validation (the same applies to the identical Javadoc on `RegisteredNodeDeleteTransaction.setRegisteredNodeId`).

sdk/src/main/java/com/hedera/hashgraph/sdk/RegisteredNodeUpdateTransaction.java
```diff
@@ -0,0 +1,256 @@
...
+        return adminKey;
+    }
+
+    /**
+     * Set administrative key controlled by the node operator.
+     * <p>
+     * This key MUST sign this transaction.<br/>
+     * This key MUST sign each transaction to update this node.<br/>
+     * This field MUST contain a valid `Key` value.<br/>
+     * This field is REQUIRED.
```
rwalworth: **nitpick:** The Javadoc on the update transaction's `setAdminKey` says "This field is REQUIRED.", but for an *update* it's optional (only set when rotating). The wording was probably copy-pasted from the create transaction.

sdk/src/main/java/com/hedera/hashgraph/sdk/RegisteredServiceEndpoint.java
rwalworth, line 1: **thought (non-blocking):** Splitting the abstract endpoint into `RegisteredServiceEndpoint` (getters + protobuf hooks) and `RegisteredServiceEndpointBase<T>` (self-typed setters) is an unusual hierarchy - the design doc only describes one abstract base. The same self-typed builder pattern works on a single class:

```java
public abstract class RegisteredServiceEndpoint<T extends RegisteredServiceEndpoint<T>> { ... }
```

The public API is fine, but it would simplify the surface area and remove a class that exists only to host setters. Worth a quick consider.

sdk/src/main/java/com/hedera/hashgraph/sdk/BlockNodeServiceEndpoint.java
```diff
@@ -0,0 +1,142 @@
...
+        }
+
+        return blockNodeEndpoint;
+    }
+
+    @Override
+    com.hedera.hashgraph.sdk.proto.RegisteredServiceEndpoint toProtobuf() {
+        if (ipAddress == null && domainName == null) {
+            throw new IllegalArgumentException(
+                    "RegisterServiceEndpoint must define either an ipAddress or  domainName");
```
rwalworth: **nitpick:** Small typo in the error message - `"ipAddress or  domainName"` has a double space and `RegisterServiceEndpoint` is missing the trailing `ed`. Same typo in `GeneralServiceEndpoint.toProtobuf()`.

```diff
- "RegisterServiceEndpoint must define either an ipAddress or  domainName"
+ "RegisteredServiceEndpoint must define either an ipAddress or a domainName"
```

sdk/src/main/java/com/hedera/hashgraph/sdk/BlockNodeServiceEndpoint.java
```diff
@@ -0,0 +1,142 @@
...
+        if (ipAddress != null) {
+            registeredServiceEndpoint.setIpAddress(ByteString.copyFrom(this.ipAddress));
+        }
+
+        if (domainName != null) {
+            registeredServiceEndpoint.setDomainName(this.domainName);
+        }
+
+        return registeredServiceEndpoint.build();
+    }
```
rwalworth, line 115: **issue (non-blocking):** `BlockNodeServiceEndpoint.toProtobuf()` and `GeneralServiceEndpoint.toProtobuf()` both throw `IllegalArgumentException` when neither `ipAddress` nor `domainName` is set, but `MirrorNodeServiceEndpoint.toProtobuf()` and `RpcRelayServiceEndpoint.toProtobuf()` silently produce a protobuf with no address (which the network will then reject with `INVALID_REGISTERED_ENDPOINT`).

For consistency, all four subtypes should either validate at the SDK boundary or none should - letting two of the four sneak past local validation is the worst of both worlds. Suggest moving the check into the abstract `RegisteredServiceEndpoint.toProtobuf()` (or into a shared helper called from each subtype) so all endpoint types behave identically.

sdk/src/main/java/com/hedera/hashgraph/sdk/RegisteredNode.java
```diff
@@ -0,0 +1,162 @@
...
+import javax.annotation.Nonnegative;
+
+/**
+ * Class representing single registered node in the network state.
+ * Each registered node in the network state SHALL represent a single
+ * non-consensus node that is registered on the network.
+ * Registered node identifiers SHALL only be unique within a single
+ * realm and shard combination.
+ */
+public class RegisteredNode {
```
rwalworth: **suggestion (non-blocking):** The design doc marks `RegisteredNode` as `@@finalType`, but the class isn't declared `final`. Same applies to `RegisteredNodeAddressBook`. Marking these `final` matches the design intent and prevents accidental subclassing.

sdk/src/main/java/com/hedera/hashgraph/sdk/RegisteredNode.java
```diff
@@ -0,0 +1,162 @@
...
+                ? adminKey.get("_type").getAsString()
+                : "";
+
+        String key = adminKey.get("key").getAsString();
+        return switch (type) {
+            case "ED25519" -> PublicKey.fromStringED25519(key);
+            case "ECDSA_SECP256K1" -> PublicKey.fromStringECDSA(key);
+            default -> PublicKey.fromString(key);
+        };
+    }
```
rwalworth, line 114: **suggestion:** This guard is dead code:

```java
String type = adminKey.get("_type").getAsString() != null
        ? adminKey.get("_type").getAsString()
        : "";
```

`JsonElement.getAsString()` never returns null - if `_type` is missing the prior `.get("_type")` returns null and you'd NPE on `.getAsString()`. Suggest something like:

```java
JsonElement typeElem = adminKey.get("_type");
String type = (typeElem != null && !typeElem.isJsonNull()) ? typeElem.getAsString() : "";
```

sdk/src/main/java/com/hedera/hashgraph/sdk/RegisteredNode.java
```diff
@@ -0,0 +1,162 @@
...
+     * Extract the registeredNode from a byte array.
+     *
+     * @param bytes the byte array
+     * @return {@code RegisteredNode} the extracted registeredNode
+     * @throws InvalidProtocolBufferException when there is an issue with the protobuf
+     */
+    public static RegisteredNode fromBytes(byte[] bytes) throws InvalidProtocolBufferException {
+        return fromProtobuf(com.hedera.hashgraph.sdk.proto.RegisteredNode.parseFrom(bytes).toBuilder()
+                .build());
+    }
```
rwalworth, line 125: **nitpick:** The `.toBuilder().build()` round-trip is unnecessary - `parseFrom(bytes)` already returns the message you want.

```diff
- return fromProtobuf(com.hedera.hashgraph.sdk.proto.RegisteredNode.parseFrom(bytes).toBuilder().build());
+ return fromProtobuf(com.hedera.hashgraph.sdk.proto.RegisteredNode.parseFrom(bytes));
```

sdk/src/main/java/com/hedera/hashgraph/sdk/RegisteredNodeAddressBook.java
```diff
@@ -0,0 +1,22 @@
...
+package com.hedera.hashgraph.sdk;
+
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Collection of RegisteredNode objects.
+ */
+public class RegisteredNodeAddressBook {
+    public final List<RegisteredNode> registeredNodes;
```
rwalworth, line 12: **suggestion:** The design doc declares `registeredNodes` as `@@immutable`, but the field here is a publicly exposed mutable `List`. Wrap in `Collections.unmodifiableList(...)` in the constructor (matching what `RegisteredNode` does for its own `serviceEndpoints` field):

```diff
- this.registeredNodes = registeredNodes;
+ this.registeredNodes = Collections.unmodifiableList(registeredNodes);
```

sdk/src/main/java/com/hedera/hashgraph/sdk/RegisteredNodeAddressBookQuery.java
```diff
@@ -0,0 +1,94 @@
...
+            } else {
+                baseUrl = baseUrl.replace(":5551", ":8084");
+            }
+        }
+
+        return performQueryToMirrorNodeAsync(baseUrl, apiEndpoint, null).exceptionally(ex -> {
+            client.getLogger().error("Error while performing post request to Mirror Node: " + ex.getMessage());
+            throw new CompletionException(ex);
+        });
+    }
```
rwalworth: **issue (non-blocking):** Hardcoding `:5551` → `:8084` for localhost (with a `hedera.mirror.registerednode.port` system-property override) is a fragile workaround for local solo testing. It bakes a development concern into production code and will silently mis-route any user whose local mirror happens to listen on `:5551`.

A few options worth considering:
1. Don't override at all - let the user configure the mirror REST URL on the `Client` and stop guessing.
2. If a localhost shim really is needed for solo, gate it behind `IntegrationTestEnv` rather than the Query class itself.
3. If kept, at minimum document the magic ports so a debugging user finds them.

Happy to defer this to a follow-up if you'd rather keep moving - the SDK contract is fine, this is a runtime detail.

sdk/src/main/java/com/hedera/hashgraph/sdk/RegisteredNodeAddressBookQuery.java
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * Executes the query with the user supplied client
+     *
+     * @param client The Client instance to perform the operation with
+     * @return The registeredAddressBook
+     * @throws ExecutionException
+     * @throws InterruptedException
+     */
+    public RegisteredNodeAddressBook execute(Client client) throws ExecutionException, InterruptedException {
```
rwalworth, line 137: **suggestion (non-blocking):** The design doc specifies `@@async @@throws RegisteredNodeAddressBook execute(client: Client)`. In Java that idiomatically maps to *both* a blocking `execute(Client)` and a `executeAsync(Client)` returning a `CompletableFuture<RegisteredNodeAddressBook>` - same pattern as `MirrorNodeContractQuery` and the existing `AddressBookQuery`.

Right now only the blocking variant is exposed (and `executeMirrorNodeRequest` returning `CompletableFuture<String>` is package-private). Worth surfacing a public `executeAsync` so async callers don't have to wrap in their own thread.

sdk/src/main/java/com/hedera/hashgraph/sdk/TransactionReceipt.java
```diff
@@ -118,6 +118,16 @@ public final class TransactionReceipt {
...
 
+    /**
+     * The identifier of a newly created RegisteredNode.
+     * <p>
+     * This value SHALL be set following a `createRegisteredNode`
+     * transaction.<br/>
+     * This value SHALL NOT be set following any other transaction.<br/>
+     * This value SHALL be unique within a given network.
+     */
+    public final long registeredNodeId;
```
rwalworth, line 129: **thought:** The design doc declares `registeredNodeId` as `@@nullable uint64`, but the field is a primitive `long` here, so callers can't distinguish "not set" from "0". This matches how the existing `nodeId` field is modeled, so I think it's fine for consistency with the codebase - just flagging that the example code uses `<= 0` as the "missing" sentinel, which works as long as the network never assigns 0.

If you want to align more strictly with the design doc you'd need a `Long` boxed type - but that ripples through the constructor signature and `toString`. Probably not worth it unless we change `nodeId` too.

sdk/src/main/proto/registered_service_endpoint.proto
rwalworth, line 1: **nitpick:** ~95% of the diff in this file is whitespace/indentation changes (2-space vs 4-space). The actual semantic change is `+ GeneralServiceEndpoint general_service = 8;` and the new `GeneralServiceEndpoint` message. If you can land the indentation change in a separate commit (or revert the indentation entirely so the diff only shows the semantic addition), it makes future review of this proto much easier.

sdk/src/main/proto/basic_types.proto
```diff
@@ -1840,6 +1840,7 @@ enum HederaFunctionality {
     AtomicBatch = 108;
 
     /**
+     * Update one or more storage slots in an lambda EVM hook.
```
rwalworth: **nitpick:** This change adds a doc-comment line about `LambdaSStore` that is unrelated to HIP-1137:

```diff
+     * Update one or more storage slots in an lambda EVM hook.
      * (DEPRECATED) Remove once no production throttle assets reference it.
~~~

Out of scope for this PR - please drop it (or split it into its own follow-up).

File: sdk/src/main/proto/node_update.proto [hunk 21]
```diff
@@ -168,6 +168,7 @@ message NodeUpdateTransactionBody {
      */
     proto.ServiceEndpoint grpc_proxy_endpoint = 10;
 
+
```
rwalworth: **nitpick:** The only diff in `sdk/src/main/proto/node_update.proto` is an added blank line. Probably not intentional - feel free to drop it to keep the diff clean.

examples/src/main/java/com/hedera/hashgraph/sdk/examples/RegisteredNodeLifeCycleExample.java
```diff
@@ -0,0 +1,186 @@
...
+         */
+
+        NodeUpdateTransaction disassociateTx = new NodeUpdateTransaction()
+                .setNodeId(0)
+                .setAssociatedRegisteredNodes(List.of()) // Empty list clear associated registeredNode
+                .freezeWith(client);
+
+        System.out.println("Disassociating registered node " + registeredNodeId + " with consensus node...");
+        disassociateTx.execute(client);
+        associateTxResponse.getReceipt(client);
```
rwalworth: **issue:** Copy-paste bug - after the disassociate execute, the example calls `getReceipt` on the *associate* response (a second time), not the disassociate response:

```java
disassociateTx.execute(client);                // ← response is discarded
associateTxResponse.getReceipt(client);        // ← wrong response
```

Likely should be:

```diff
- disassociateTx.execute(client);
- associateTxResponse.getReceipt(client);
+ TransactionResponse disassociateTxResponse = disassociateTx.execute(client);
+ disassociateTxResponse.getReceipt(client);
```

examples/src/main/java/com/hedera/hashgraph/sdk/examples/RegisteredNodeLifeCycleExample.java
```diff
@@ -0,0 +1,186 @@
...
+         * Step 1:
+         * Generate an admin key pair and configure a BlockNodeServiceEndpoint
+         * for use in the RegisterNodeTransaction.
+         */
+        PrivateKey adminKey = PrivateKey.generateED25519();
+        BlockNodeServiceEndpoint initialEndpoint = new BlockNodeServiceEndpoint()
+                .setIpAddress(new byte[] {127, 0, 0, 1})
+                .setPort(443)
+                .setRequiresTls(true)
+                .addEndpointApi(BlockNodeApi.SUBSCRIBE_STREAM);
```
rwalworth: **suggestion (non-blocking):** The design doc's lifecycle (step 2) explicitly demonstrates passing multiple APIs at once: `endpointApis` set to `[BlockNodeApi.SUBSCRIBE_STREAM, BlockNodeApi.STATUS]`. The example here only calls `addEndpointApi(BlockNodeApi.SUBSCRIBE_STREAM)`. Since this is the showcase example for HIP-1137, it's worth exercising the multi-API case so users see how it works.

examples/src/main/java/com/hedera/hashgraph/sdk/examples/RegisteredNodeLifeCycleExample.java
```diff
@@ -0,0 +1,186 @@
...
+
+        /*
+         * Step 5:
+         * Add the registeredNodeId as associatedRegisteredNodes to a Node.
+         */
+
+        NodeUpdateTransaction associateTx = new NodeUpdateTransaction()
+                .setNodeId(0)
+                .addAssociatedRegisteredNode(registeredNodeId)
+                .freezeWith(client);
```
rwalworth, line 152: **thought (non-blocking):** Per the design doc step 13, this step "Sign with the consensus node's admin key and execute". The example just relies on the operator key. That'll only work in environments where the operator happens to also control consensus node 0 (so it's fine for solo / testnet runs of the example), but the comment in the example might mislead a user into thinking any client account can update any node. Worth a one-liner clarifying the assumption.

sdk/src/testIntegration/java/com/hedera/hashgraph/sdk/test/integration/RegisteredNodeUpdateTransactionIntegrationTest.java
```diff
@@ -0,0 +1,287 @@
...
+
+            var registeredNodes = new RegisteredNodeAddressBookQuery()
+                    .setRegisteredNodeId(nodeId)
+                    .execute(testEnv.client)
+                    .registeredNodes;
+            assertThat(registeredNodes).hasSize(1);
+
+            var registeredNode = registeredNodes.getFirst();
+            assertThat(registeredNode.serviceEndpoints).hasSize(1);
+            assertThat(registeredNode.serviceEndpoints.getFirst() instanceof BlockNodeServiceEndpoint);
```
rwalworth: **issue (test):** This assertion is a no-op:

```java
assertThat(registeredNode.serviceEndpoints.getFirst() instanceof BlockNodeServiceEndpoint);
```

`assertThat(boolean)` returns a `BooleanAssert` that's never terminated with `.isTrue()`, so the test passes regardless of the `instanceof` result. You'll then crash one line later when the cast actually fails on a real regression. Same issue at line 228.

```diff
- assertThat(registeredNode.serviceEndpoints.getFirst() instanceof BlockNodeServiceEndpoint);
+ assertThat(registeredNode.serviceEndpoints.getFirst()).isInstanceOf(BlockNodeServiceEndpoint.class);
```

sdk/src/testIntegration/java/com/hedera/hashgraph/sdk/test/integration/RegisteredNodeCreateTransactionIntegrationTest.java
```diff
@@ -0,0 +1,205 @@
...
+
+            var receipt = response.getReceipt(testEnv.client);
+
+            assertThat(receipt.status).isEqualTo(Status.SUCCESS);
+            assertThat(receipt.registeredNodeId).isGreaterThan(0);
+        }
+    }
+
+    @Test
+    @DisplayName("Can create a registered node with mirrorNodeServiceEndpoint")
```
rwalworth, line 49: **nitpick:** Typo in the test method name - `WitMirrorNode` should be `WithMirrorNode`. Easy to miss but worth catching before it becomes a stable test name people grep for.

rwalworth (changes requested):
Really nice work tracking the HIP through all the protobuf and design changes @manishdait. The shape of the implementation lines up well with the design doc, the test coverage maps cleanly to the 19-point test plan, and everything is properly signed.

I left a handful of comments below for @mustafauzunn to check out.  Happy to chat through any of them.

**Heads-up on solo + mirror v0.153.0** - there's an open P0 against solo ([hiero-ledger/solo#4228](https://github.com/hiero-ledger/solo/issues/4228)) reporting that the mirror node REST endpoint `GET /api/v1/network/registered-nodes` returns an empty `registered_nodes` array on solo `>= 0.69.0` even after a successful `RegisteredNodeCreateTransaction` reaches consensus - this is the exact mirror version (`v0.153.0`) and the exact endpoint your `RegisteredNodeAddressBookQuery` hits. The most recent update notes the failure mode looks more like a warm-up race than a hard version regression.

Worth knowing how the JS SDK landed HIP-1137 against the same image combo (it merged on 2026-05-04 in `hiero-sdk-js` #3846):

- They use the **same** `hieroVersion: v0.73.0` / `mirrorNodeVersion: v0.153.0` - they did NOT downgrade.
- They did NOT skip HIP-1137 ingestion-dependent tests (only one unrelated test for "description > 100 UTF-8 bytes" is skipped, pending an empirical status pin).
- They built **retry + backoff into the query itself** (`src/network/RegisteredNodeAddressBookQuery.js`): configurable `setMaxAttempts` / `setMaxBackoff`, exponential backoff starting at 250ms, plus a shared `src/network/mirrorRestRetry.js` helper that classifies retryable errors (5xx, AbortError, ECONN/ENETUNREACH, "fetch failed", etc.) and re-pages until the deadline.
- They added `installBlockNode: true` to the affected build.yml jobs.
- For HIP-1261 (`FeeEstimateQuery`) they had a separate but related journey - first skipping the tests entirely (`722aefbc`), then re-enabling them (`#3979`), then ultimately landing a `beforeAll` polling probe with a 10-minute timeout that doesn't run any test until the mirror's `FeeEstimationService` actually answers (`#3996`).

The Java implementation here doesn't have an analogous retry layer - `RegisteredNodeAddressBookQuery` is a single async fetch, and the integration tests rely on a fixed `Thread.sleep(5000)` to wait out mirror ingestion. Given solo #4228 is open and the failure profile is a timing race, that combination is likely to flake on CI. Two non-blocking suggestions to harden it:

1. Port JS's retry/backoff + paging into `RegisteredNodeAddressBookQuery` (configurable `setMaxAttempts` / `setMaxBackoff`, retry on transient HTTP/network errors).
2. Replace the `Thread.sleep(5000)` waits in the integration tests with a poll-with-deadline that re-issues the query until it returns a non-empty result or the deadline expires - same shape as JS's `waitForFeeEstimationServiceReady`.

sdk/src/main/java/com/hedera/hashgraph/sdk/RegisteredServiceEndpoint.java
manishdait: I initially try to use the single class for this, the only reason, I stuck with the two class approach was to avoid  generics raw type warning. If we use `RegisteredServiceEndpoint<T>`, the `List` in `RegisteredNode` ends up throwing 'Raw use' warnings or forcing to use `List<RegisteredServiceEndpoint<?>>`.

But happy to change it if you’d prefer the single class version though, let me know what you think.

sdk/src/main/java/com/hedera/hashgraph/sdk/RegisteredNodeAddressBookQuery.java
manishdait: Done

sdk/src/main/java/com/hedera/hashgraph/sdk/RegisteredNodeAddressBookQuery.java
```diff
@@ -0,0 +1,308 @@
...
+
+        return baseUrl.replace("/api/v1", "");
+    }
+
+    private String buildInitialPath() {
+        StringBuilder path = new StringBuilder("/api/v1/network/registered-nodes");
+        limit = limit > 0 ? limit : DEFAULT_LIMIT;
+        path.append("?limit=").append(limit);
+
+        if (registeredNodeId > 0) {
```
mustafauzunn: The filter `if (registeredNodeId > 0)` means node ID 0 cannot be queried.

sdk/src/main/java/com/hedera/hashgraph/sdk/RegisteredNodeAddressBookQuery.java
```diff
@@ -0,0 +1,308 @@
...
+
+    /**
+     * Executes the query asynchronously with the user supplied client
+     *
+     * @param client The Client instance to perform the operation with.
+     * @return A future representing the retrieved {@link RegisteredNodeAddressBook}.
+     */
+    public CompletableFuture<RegisteredNodeAddressBook> executeAsync(Client client) {
+        Objects.requireNonNull(client, "client must not be null");
+        Objects.requireNonNull(client, "client must not be null");
```
mustafauzunn: Remove the duplicate line

sdk/src/main/java/com/hedera/hashgraph/sdk/RegisteredNodeAddressBookQuery.java
```diff
@@ -0,0 +1,308 @@
...
+    /**
+     * Executes the query asynchronously with the user supplied client with timeout
+     *
+     * @param client The Client instance to perform the operation with.
+     * @param timeout The maximum duration for each individual HTTP request.
+     * @return A future representing the retrieved {@link RegisteredNodeAddressBook}.
+     */
+    public CompletableFuture<RegisteredNodeAddressBook> executeAsync(Client client, Duration timeout) {
+        Objects.requireNonNull(client, "client must not be null");
+        Objects.requireNonNull(client, "client must not be null");
```
mustafauzunn: Remove the duplicate line

sdk/src/main/java/com/hedera/hashgraph/sdk/RegisteredNodeAddressBookQuery.java
manishdait: Done

mustafauzunn (approved):

PavelSBorisov (approved):

PR #2883, opened 2026-08-17
feat: MirrorNodeTokenBalanceQuery

### LINKED ISSUE #2885: Implement MirrorNodeTokenBalanceQuery

mustafauzunn:
**Description**:

Implement balance queries

**Related issue(s)**:

Fixes #2885 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/examples/build.gradle.kts
+++ b/examples/build.gradle.kts
@@ -85,6 +85,9 @@ abstract class RunAllExample : DefaultTask() {
                 .filter { it != "AccountHooksExample" }
                 .filter { it != "ContractHooksExample" }
                 .filter { it != "TransferTransactionHooksExample" }
+                .filter { it != "FeeEstimateQueryExample" }
+                .filter { it != "AccountCreateWithHtsExample" }
+                .filter { it != "AccountAliasExample" }
                 .toList()
 
         exampleClasses.forEach { className ->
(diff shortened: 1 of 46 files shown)


hendrikebbers (changes requested):
I have huge concerns with the given implementation. Let me explain why:

`private static final HttpClient HTTP_CLIENT = HttpClient.newHttpClient();` now appears in:

- `MirrorNodeAccountBalanceQuery`
- `MirrorNodeTokenBalanceQuery` (this PR)
- `FeeEstimateQuery`
- `RegisteredNodeAddressBookQuery`
- (plus `EntityIdHelper.performQueryToMirrorNodeAsync`, which builds a new `HttpClient` on every call.)

Consequences beyond the duplication itself:

- Each static `HttpClient` owns its own selector thread and connection pool.
- None of them is ever closed. `HttpClient` is `AutoCloseable` since Java 21, and none of these participate in `Client.close()`. A short-lived Client leaks transport threads.

Other problems: 

- The `HttpClient` doesn't use client.executor, so `CompletableFuture.supplyAsync(..., client.executor)` gives the appearance of executor control while the transport keeps its own threads.
- `HttpClient.newHttpClient() `with no builder means callers cannot supply: a corporate proxy, a custom truststore or mTLS material for a private mirror node, an Authorization header, connection-pool / HTTP-version settings, a User-Agent identifying the SDK version (mirror-node operators do want this for traffic attribution), or tracing/observability interceptors.
- `java.net.http` is not part of the Android class library at any API level. Every class listed above therefore throws `NoClassDefFoundError` on Android. I assume Android is a quite important target for the lib to create native Wallets.

My suggestion is to stop all this development and start with an architecture concept for how HTTP calls are done in the SDK. If that s not happening every PR like this will make the project worse. I already mentioned that in https://github.com/hiero-ledger/sdk-collaboration-hub/pull/281

PR #2899, opened 2026-08-19
chore: Decouple JSON-RPC result serialization from json-smart bean serialization

### LINKED ISSUE #2890: Decouple JSON-RPC result serialization from json-smart's getter-based bean writer
`JRPCController.handleJSONRPC2Request` returns `resp.toJSONString()`. Inside `jsonrpc2-base`, the
result object is handed to `net.minidev.json.JSONValue`, which serialises unknown types through
`net.minidev.asm.BeansWriterASM`. That writer discovers properties **exclusively by bean getter
name** (`getX()` / `isX()`).

This has two concrete consequences.

**It already forces hand-written workarounds.** `methods/sdk/response/GetAccountInfoResponse.java:29-36`:

```java
// Explicit getters to ensure proper JSON serialization
public boolean getIsDeleted() {
    return isDeleted;
}

public boolean getIsReceiverSignatureRequired() {
    return isReceiverSignatureRequired;
}
```

Lombok generates `isDeleted()` for a `boolean isDeleted` field, which the writer maps to the JSON
key `deleted` — so a manual `getIsDeleted()` had to be added to get the wire format right. The
wire contract of the TCK server currently depends on Lombok's getter-naming rules.

**It silently drops records.** Verified against the version this project actually resolves
(`net.minidev:json-smart:2.6.0`, via `accessors-smart:2.6.0` + `asm`):

```
record : {}
bean   : {"accountId":"0.0.1","deleted":true,"status":"OK"}
```

A record component accessor is `accountId()`, not `getAccountId()`, so every record serialises to an
empty JSON object. There is **no compile error and no runtime error** — the TCK server would answer
valid JSON-RPC responses with an empty result payload, and the failure would surface as confusing
assertion failures in the TCK driver rather than as a server error.

That is the reason no response type in this module can be a record today, and it is the single
prerequisite for the rest of this series.

## Proposed change

Convert the method result to a `Map`/`JSONObject` **before** it reaches `JSONRPC2Response`, using a
serializer that understands modern Java types, instead of relying on json-smart's reflective bean
writer.

`JSONValue` handles `Map` natively and does not go through `BeansWriterASM` at all, so this needs no
change to `jsonrpc2-base` and no custom `JsonWriterI` registration.

Suggested place: `AbstractJSONRPC2Service.process`, where the result is wrapped today:

```java
Object result = method.invoke(this, args);
return new JSONRPC2Response(result, req.getID());
```

becomes a conversion step through an `ObjectMapper` (`convertValue(result, Map.class)`) before
wrapping. Jackson supports records natively since 2.12.

Notes:

- `jackson-databind` is already on the TCK runtime classpath transitively via
  `runtimeOnly("spring.boot.starter.web")`, but it is **not** currently declared for compilation —
  `tck/build.gradle.kts` needs an explicit `requires("com.fasterxml.jackson.databind")` and the
  matching constraint in `hiero-dependency-versions/build.gradle.kts`. There are currently zero
  Jackson usages in `tck/src`.
- Alternative if introducing Jackson here is unwanted: register a custom `JsonWriterI` with
  json-smart that reflects over record components. Cheaper in dependencies, but it means maintaining
  reflection code by hand. The Jackson route is recommended.

## The part that needs care

**This is a wire-format change until proven otherwise.** The current output shape is an accident of
Lombok's getter naming; any new serializer will produce a shape derived from field/component names.
The two known divergences are the `getIsDeleted` / `getIsReceiverSignatureRequired` pair above, but
there may be more (e.g. `@Data` classes where a field name and its getter disagree, or where a
getter exists without a backing field).

The TCK is a conformance harness — a silently changed response key breaks the driver, not this
build. So the acceptance criteria below are deliberately weighted towards proving the output is
byte-identical.

## Acceptance criteria

- [ ] JSON-RPC result payloads are produced by a serializer that supports records; json-smart's
      `BeansWriterASM` is no longer on the path for result objects.
- [ ] A test captures the serialised JSON for **every** type in `methods/sdk/response` (21 classes)
      before and after the change and asserts the two are equal — no renamed, added or dropped keys.
- [ ] The manual `getIsDeleted()` / `getIsReceiverSignatureRequired()` workarounds in
      `GetAccountInfoResponse` are removed, with the test above proving the keys are unchanged.
- [ ] A record used as a method result serialises correctly (regression guard for the `{}` behaviour
      documented above).
- [ ] `null` / `Optional.empty()` handling in the output is unchanged (json-smart and Jackson differ
      in their defaults for null inclusion — pin this explicitly rather than inheriting a default).
- [ ] The TCK suite is run against a network and passes.

## References

- `tck/src/main/java/com/hedera/hashgraph/tck/methods/AbstractJSONRPC2Service.java` — result wrapping
- `tck/src/main/java/com/hedera/hashgraph/tck/controller/JRPCController.java` — `resp.toJSONString()`
- `tck/src/main/java/com/hedera/hashgraph/tck/methods/sdk/response/GetAccountInfoResponse.java:29-36`
- `net.minidev:json-smart:2.6.0` (resolved via `spring-boot-dependencies`)

manishdait:
**Description**:
This PR decouples JSON-RPC result serialization from json-smarts  bean serialization by converting method results to `Map<String, Object>` using jackson before creating the `JSONRPC2Response`.

**Related issue(s)**:

Fixes #2890

**Notes for reviewer**:
- Jackson field visibility configuration can be remove once the response dtos become records.
- Added @JsonInclude(NON-NULL) cause some tck response need the null fields to exclude
- Run the changes against hiero-sdk-tck and solo
Screenshot From 2026-08-20 00-06-02

**Checklist**


tck/src/main/java/com/hedera/hashgraph/tck/methods/sdk/response/token/TokenResponse.java
```diff
@@ -2,29 +2,10 @@
 package com.hedera.hashgraph.tck.methods.sdk.response.token;
 
 import com.hedera.hashgraph.sdk.Status;
+import lombok.AllArgsConstructor;
 
+@AllArgsConstructor
```
hendrikebbers, line 7: Is that needed? At the end one goal would be to remove Lombok usage at all. Adding new usage if not possible looks wrong to me.

tck/src/main/java/com/hedera/hashgraph/tck/methods/AbstractJSONRPC2Service.java
```diff
@@ -129,4 +138,14 @@ private Object[] getArguments(final Method method, final Map<String, Object> jrp
...
         return args;
     }
+
+    /**
+     * Converts a method result to a map for JSON-RPC serialization.
+     *
+     * @param result the method result to convert
+     * @return the converted result as a map
+     */
+    private Map<String, Object> convertValue(final Object result) {
```
hendrikebbers: since OBJECT_MAPPER is static this method can be static, too

tck/src/main/java/com/hedera/hashgraph/tck/methods/sdk/response/TransactionReceiptResponse.java
```diff
@@ -1,14 +1,14 @@
...
+import com.fasterxml.jackson.annotation.JsonInclude;
 import java.util.List;
 import lombok.Data;
-import net.minidev.json.JSONAware;
-import net.minidev.json.JSONObject;
 import org.jspecify.annotations.Nullable;
 
 @Data
-public class TransactionReceiptResponse implements JSONAware {
+@JsonInclude(JsonInclude.Include.NON_NULL)
```
hendrikebbers, line 10: While I 100% like those changes I do not see any changes in tests. Do we have any tests on that we check that the objects are serialized correctly?

tck/src/main/java/com/hedera/hashgraph/tck/methods/sdk/response/token/TokenResponse.java
manishdait: yay, added it because i removed the constructor for this class. but It will get removed once class get converted to record.

tck/src/main/java/com/hedera/hashgraph/tck/methods/AbstractJSONRPC2Service.java
manishdait: Done

tck/src/main/java/com/hedera/hashgraph/tck/methods/sdk/response/TransactionReceiptResponse.java
manishdait: done, added serialization unit test for all the dtos

hendrikebbers (dismissed):

mustafauzunn (commented):
@manishdait is this changes tested with latest main in tck repo

tck/src/main/java/com/hedera/hashgraph/tck/methods/sdk/response/TransactionReceiptResponse.java
```diff
@@ -51,51 +51,10 @@ public class TransactionReceiptResponse implements JSONAware {
...
-        putIfNotNull(json, "duplicates", duplicates);
-        putIfNotNull(json, "children", children);
-        putIfNotNull(json, "nodeId", nodeId);
-
-        return json.toJSONString();
-    }
-
     @Data
-    public static class ExchangeRate implements JSONAware {
+    public static class ExchangeRate {
```
mustafauzunn, line 56: TransactionReceiptResponse.ExchangeRate lost its null suppression. Its toJSONString() used putIfNotNull, but the nested class did not get @JsonInclude(NON_NULL)

manishdait: Done

mustafauzunn (approved):

PR #2920, opened 2026-08-28
chore: Converted TCK response classes to records

### LINKED ISSUE #2891: Convert the 21 response DTOs to records
Depends on https://github.com/hiero-ledger/hiero-sdk-java/issues/2890

All 21 classes in `tck/src/main/java/com/hedera/hashgraph/tck/methods/sdk/response/` are pure data
carriers built with Lombok's `@Data`. 16 of them have exclusively `final` fields, which means
`@Data` generates precisely what a record is: an all-args constructor, accessors, `equals`,
`hashCode` and `toString`. For example `response/AccountResponse.java` in full:

```java
@Data
public class AccountResponse {
    private final String accountId;
    private final Status status;
    private final String transactionId;
}
```

This is a record written the long way round, with an annotation processor as the mechanism.

The module already contains a record for exactly this purpose —
`methods/JSONRPC2Error.java:18`:

```java
public record ErrorData(Status status, String message) {}
```

so the direction is not in dispute; the response package simply predates it and could not follow
because of the serialization issue tracked in [TCK 1/4].

## Proposed change

Convert the response types in `methods/sdk/response/` (including the nested `schedule/` and `token/`
packages and the nested static classes inside `GetAccountInfoResponse`) from `@Data` classes to
records, and drop their Lombok imports.

The classes fall into two groups:

**Group A — all fields final (16 classes).** Mechanical conversion. Examples: `AccountResponse`,
`AccountBalanceResponse`, `FileInfoResponse`, `TokenInfoResponse` (28 fields), `ContractCallResponse`,
`NftInfoResponse`, `NodeResponse`, `ScheduleResponse`, `TransactionReceiptResponse`.

**Group B — non-final fields (5 classes):** `ContractResponse`, `GetAccountInfoResponse`,
`ScheduleInfoResponse`, `SetupResponse`, `GenerateKeyResponse`. These carry `@Data` plus
`@NoArgsConstructor`/`@AllArgsConstructor`, i.e. they are technically mutable. A check of the call
sites shows they are nonetheless always constructed through the all-args constructor and never
mutated afterwards (e.g. `AccountService.mapAccountInfoResponse`, `AccountService:267`), so the
mutability is unused and they can become records too. This should be re-verified per class during
implementation rather than assumed.

`GetAccountInfoResponse` additionally contains nested `@Data` types (`LiveHashResponse`,
`TokenRelationshipInfo`, `StakingInfoResponse`, allowance responses) — these become nested records.

## Notes

- Two files in this package import `jakarta.annotation.Nullable`
  (`TopicInfoResponse.java:4`, `schedule/ScheduleInfoResponse.java:4`) while five siblings import
  `javax.annotation.Nullable`. Worth unifying while touching these files, though the actual
  annotation-library decision is tracked separately.
- `@Nonnegative` appears on 2 fields here (`AccountBalanceResponse`). It carries no enforcement; keep
  or drop consistently with whatever is decided repo-wide.
- Records are implicitly final and cannot extend a class. Nothing in this package uses inheritance
  today, so this is not a constraint here — unlike the parameter classes, see [TCK 3/4].

## Acceptance criteria

- [ ] No `@Data` and no `lombok.*` import remains under `methods/sdk/response/`.
- [ ] The serialised JSON for every response type is byte-identical to before the change (reuse the
      test introduced in [TCK 1/4]).
- [ ] Mutability was verified as unused for each of the 5 Group-B classes, or the class was left as-is
      with a comment explaining why it must stay mutable.
- [ ] The TCK suite is run against a network and passes.

## References

- `tck/src/main/java/com/hedera/hashgraph/tck/methods/sdk/response/` — 21 classes
- `tck/src/main/java/com/hedera/hashgraph/tck/methods/JSONRPC2Error.java:18` — existing record
- `tck/src/main/java/com/hedera/hashgraph/tck/methods/sdk/AccountService.java:266` — construction site

manishdait:
**Description**:
This PR converts the tck response classes to the records

**Related issue(s)**:

Fixes #2891

**Notes for reviewer**:
- Updated parameters type for `ContractCallResponse` class to use string instead of raw Hbar and AccountId class.
- Tested against hiero-sdk-tck with solo version `0.87.1`
Screenshot From 2026-08-30 18-38-03

**Checklist**

### DIFF
--- a/tck/src/main/java/com/hedera/hashgraph/tck/methods/AbstractJSONRPC2Service.java
+++ b/tck/src/main/java/com/hedera/hashgraph/tck/methods/AbstractJSONRPC2Service.java
@@ -1,8 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 package com.hedera.hashgraph.tck.methods;
 
-import com.fasterxml.jackson.annotation.JsonAutoDetect;
-import com.fasterxml.jackson.annotation.PropertyAccessor;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.hedera.hashgraph.sdk.PrecheckStatusException;
@@ -37,10 +35,7 @@ public abstract class AbstractJSONRPC2Service implements RequestHandler {
     // although the tck driver would not call these methods in such way
     private final Map<String, Method> methodMap;
 
-    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper()
-            // TODO: Should remove the below visiblity once dtos are change to use record.
-            .setVisibility(PropertyAccessor.ALL, JsonAutoDetect.Visibility.NONE)
-            .setVisibility(PropertyAccessor.FIELD, JsonAutoDetect.Visibility.ANY);
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
 
     protected AbstractJSONRPC2Service() {
         methodMap = new HashMap<>();
--- a/tck/src/main/java/com/hedera/hashgraph/tck/methods/sdk/ContractService.java
+++ b/tck/src/main/java/com/hedera/hashgraph/tck/methods/sdk/ContractService.java
@@ -73,8 +73,8 @@ public class ContractService extends AbstractJSONRPC2Service {
                 result.gasUsed,
                 result.logs,
                 result.gas,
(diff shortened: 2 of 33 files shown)


mustafauzunn (commented):
Some fileds javadoc dropped during conversion

mustafauzunn (approved):

PR #2933, opened 2026-09-04
chore: Replace reflective no-arg instantiation of JSON-RPC parameters with a typed parser registry

### LINKED ISSUE #2892: Replace reflective no-arg instantiation of JSON-RPC parameters with a typed parser registry
Depends on https://github.com/hiero-ledger/hiero-sdk-java/issues/2890 and https://github.com/hiero-ledger/hiero-sdk-java/issues/2891

Parameter objects are created reflectively in `methods/AbstractJSONRPC2Service.java:118-126`:

```java
var paramInstance = paramTypes[i].newInstance();
if (paramInstance instanceof JSONRPC2Param jsonRpcParam) {
    args[i] = jsonRpcParam.parse(jrpcParams);
}
```

An empty throwaway instance is constructed solely so that `parse()` — an instance method that
behaves as a static factory — can be invoked on it. `parse()` then builds the *real* instance via
the all-args constructor. Three problems follow from this.

**1. The contract is documented in prose, not in the type system.** `methods/JSONRPC2Param.java:13-18`:

> *IMPORTANT: all inheriting classes should include the following Lombok annotations: `@Getter`,
> `@AllArgsConstructor`, and `@NoArgsConstructor`. These annotations are needed for the instance
> creation via reflection.*

A new parameter class that forgets `@NoArgsConstructor` compiles cleanly, passes review, and fails
only at runtime. And it fails badly: `getArguments` catches `Exception` broadly and converts it into
`InvalidJSONRPC2ParamsException` → a generic `INVALID_PARAMS` JSON-RPC error. A missing annotation is
therefore indistinguishable from a genuinely malformed client request. This is currently the only
thing forcing `@NoArgsConstructor` onto 57 classes.

**2. It uses a deprecated JDK API.** `Class.newInstance()` has been deprecated since Java 9 (it
propagates checked exceptions undeclared). The suppression `@SuppressWarnings("java:S1874")` on
`process` acknowledges this rather than fixing it.

**3. It blocks records.** `JSONRPC2Param` is an abstract *class*, and records cannot extend a class.
Combined with the mandatory no-arg constructor, the 57 parameter classes cannot become records
regardless of what happens to the serialization side.

## Proposed change

Replace "instantiate, then parse" with "parse via a factory". Two viable shapes:

**Option A — static factory + registry (recommended).** Turn `JSONRPC2Param` into a marker interface
(or drop it entirely) and register a parser per parameter type:

```java
Map<Class<?>, Function<Map<String, Object>, ? extends JSONRPC2Param>> PARSERS
```

populated once at startup. `getArguments` then looks up the parser instead of calling
`newInstance()`. Each parameter class exposes `static XParams parse(Map<String, Object>)`.

**Option B — reflective canonical-constructor invocation.** Keep discovery reflective but call the
canonical constructor instead of the no-arg one. Less code to write, but keeps a reflective contract
that the compiler still cannot check — it only moves the failure, it does not remove it.

Option A is recommended precisely because it makes the contract compile-time visible, which is the
main point of this issue.

Either way:

- `@NoArgsConstructor` becomes unnecessary on all 57 classes.
- Error handling can distinguish "no parser registered for this parameter type" (a server bug,
  → `INTERNAL_ERROR`) from "the client sent unusable params" (→ `INVALID_PARAMS`). Today both are
  the latter.
- The parameter classes become eligible to be records in a follow-up (see below).

## Scope note: records for parameters

Once the abstract base class and the no-arg constructor are gone, the 65 parameter classes *could*
become records. That conversion is deliberately **not** required by this issue, for two reasons:

- The parameter classes hold 289 `Optional<…>` fields, flagged 20× with
  `@SuppressWarnings("OptionalUsedAsFieldOrParameterType")`. `Optional` as a record component is its
  own design debate (the alternative being nullable components plus accessor-level `Optional`), and
  it should not be settled inside a refactoring issue.
- The `parse` bodies are long and mechanical (`AccountCreateParams` has 12 components); converting
  them is a large diff with little benefit until the `Optional` question is answered.

Recommendation: land the redesign first, then open a separate issue for the record conversion once
the `Optional` policy is decided. Note that [TCK 4/4] (removing Lombok) *does* require the
`@Getter`/`@AllArgsConstructor` usages to be dealt with — either by converting to records or by
generating/writing the accessors another way.

## Acceptance criteria

- [ ] `Class.newInstance()` is gone from `AbstractJSONRPC2Service`; `@SuppressWarnings("java:S1874")`
      is removed with it.
- [ ] A parameter type that is not correctly wired fails **at compile time**, or — if that is not
      achievable — at application startup, never per-request.
- [ ] `@NoArgsConstructor` is removed from all 57 `JSONRPC2Param` subclasses.
- [ ] The prose contract in `JSONRPC2Param`'s Javadoc is replaced by whatever the type system now
      enforces.
- [ ] A "no parser registered" condition maps to `INTERNAL_ERROR`, not `INVALID_PARAMS`.
- [ ] The broad `catch (Exception e)` in `getArguments` is narrowed so that genuine bugs are not
      reported to the client as malformed input.
- [ ] The TCK suite is run against a network and passes, including the negative cases (malformed
      params must still yield `INVALID_PARAMS`).

## References

- `tck/src/main/java/com/hedera/hashgraph/tck/methods/AbstractJSONRPC2Service.java:112-131`
- `tck/src/main/java/com/hedera/hashgraph/tck/methods/JSONRPC2Param.java`
- `tck/src/main/java/com/hedera/hashgraph/tck/methods/sdk/param/account/AccountCreateParams.java` —
  representative of the 57 affected classes

manishdait:
**Description**:
This PR replace the no-args instantiation of JSON-RPC params to use the typed registry.

**Changes Made:**
- Update the `JSONRPC2Param` abstract class to the marker interface as describe in the issue
- Remove the @NoArgsConstructor
- Change  `parse()` method in the request classes to static one
- Created a `JSONRPC2ParamRegistry`

**Related issue(s)**:

Fixes #2892

**Notes for reviewer**:
- Keep the registry manually maintained rather than dynamically populated. since the issue aims to eliminate reflection, happy to make it dynamic if needed :)
- Test the changes against hiero-sdk-tck using solo version `0.87.1`
Screenshot From 2026-09-04 18-48-12

**Checklist**

### DIFF
--- a/tck/src/main/java/com/hedera/hashgraph/tck/methods/JSONRPC2ParamRegistry.java
+++ b/tck/src/main/java/com/hedera/hashgraph/tck/methods/JSONRPC2ParamRegistry.java
@@ -3,6 +3,7 @@ package com.hedera.hashgraph.tck.methods;
 
 import static java.util.Map.entry;
 
+import com.hedera.hashgraph.tck.exception.InvalidJSONRPC2ParamsException;
 import com.hedera.hashgraph.tck.methods.sdk.param.BaseParams;
 import com.hedera.hashgraph.tck.methods.sdk.param.CustomFee;
 import com.hedera.hashgraph.tck.methods.sdk.param.GenerateKeyParams;
@@ -155,10 +156,12 @@ public final class JSONRPC2ParamRegistry {
      * @param params the JSON-RPC request parameters to parse
      * @return the parsed JSON-RPC parameter
      */
-    public static JSONRPC2Param parse(Class<?> parameterType, Map<String, Object> params) {
+    public static JSONRPC2Param parse(Class<?> parameterType, Map<String, Object> params)
+            throws InvalidJSONRPC2ParamsException {
         Function<Map<String, Object>, ? extends JSONRPC2Param> parser = REGISTRY.get(parameterType);
         if (parser == null) {
-            throw new IllegalArgumentException("No parser registered for parameter type: " + parameterType.getName());
+            throw new InvalidJSONRPC2ParamsException(
+                    "No parser registered for parameter type: " + parameterType.getName());
         }
 
         return parser.apply(params);
--- a/tck/src/main/java/com/hedera/hashgraph/tck/methods/JSONRPC2ParamRegistry.java
+++ b/tck/src/main/java/com/hedera/hashgraph/tck/methods/JSONRPC2ParamRegistry.java
@@ -1,0 +1,148 @@
+// SPDX-License-Identifier: Apache-2.0
+package com.hedera.hashgraph.tck.methods;
+
+import static java.util.Map.entry;
+
+import com.hedera.hashgraph.tck.methods.sdk.param.BaseParams;
+import com.hedera.hashgraph.tck.methods.sdk.param.CustomFee;
+import com.hedera.hashgraph.tck.methods.sdk.param.GenerateKeyParams;
+import com.hedera.hashgraph.tck.methods.sdk.param.SetupParams;
+import com.hedera.hashgraph.tck.methods.sdk.param.TransactionReceiptQueryParams;
+import com.hedera.hashgraph.tck.methods.sdk.param.account.AccountAllowanceParams;
+import com.hedera.hashgraph.tck.methods.sdk.param.account.AccountBalanceQueryParams;
+import com.hedera.hashgraph.tck.methods.sdk.param.account.AccountCreateParams;
+import com.hedera.hashgraph.tck.methods.sdk.param.account.AccountDeleteParams;
+import com.hedera.hashgraph.tck.methods.sdk.param.account.AccountUpdateParams;
+import com.hedera.hashgraph.tck.methods.sdk.param.account.GetAccountInfoParams;
+import com.hedera.hashgraph.tck.methods.sdk.param.contract.ContractByteCodeQueryParams;
+import com.hedera.hashgraph.tck.methods.sdk.param.contract.ContractCallQueryParams;
+import com.hedera.hashgraph.tck.methods.sdk.param.contract.CreateContractParams;
+import com.hedera.hashgraph.tck.methods.sdk.param.contract.DeleteContractParams;
+import com.hedera.hashgraph.tck.methods.sdk.param.contract.ExecuteContractParams;
+import com.hedera.hashgraph.tck.methods.sdk.param.contract.InfoQueryContractParams;
+import com.hedera.hashgraph.tck.methods.sdk.param.contract.UpdateContractParams;
+import com.hedera.hashgraph.tck.methods.sdk.param.ethereum.EthereumTransactionParams;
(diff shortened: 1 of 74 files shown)


hendrikebbers (dismissed):

tck/src/main/java/com/hedera/hashgraph/tck/methods/JSONRPC2ParamRegistry.java
```diff
@@ -0,0 +1,200 @@
...
+     */
+    private static <InputType, ReturnType extends JSONRPC2Param> Function<InputType, ReturnType> toUncheckedFunction(
+            @NonNull CheckedFunction<InputType, ReturnType> function) {
+        Objects.requireNonNull(function, "function must not be null");
+
+        return args -> {
+            try {
+                return function.apply(args);
+            } catch (Exception e) {
+                throw new IllegalArgumentException("Failed to parse JSON-RPC parameters");
```
hendrikebbers:
```suggestion
                throw new IllegalArgumentException("Failed to parse JSON-RPC parameters", e);
```

tck/src/main/java/com/hedera/hashgraph/tck/util/JSONRPCParamParser.java
```diff
@@ -31,7 +31,7 @@ public static Optional<List<AllowanceParams>> parseAllowances(Map<String, Object
     }
 
     public static Optional<List<CustomFee>> parseCustomFees(Map<String, Object> jrpcParams) throws Exception {
-        return parseJsonArray(jrpcParams, "customFees", jsonObj -> new CustomFee().parse(jsonObj));
+        return parseJsonArray(jrpcParams, "customFees", jsonObj -> CustomFee.parse(jsonObj));
```
hendrikebbers: I now we have others in the class, too. But can we can rid of the magic string inline and make it a constant, pls

manishdait: @hendrikebbers, if I’m not mistaken, do you mean making the "customFees" field name a constant? :)

tck/src/main/java/com/hedera/hashgraph/tck/methods/JSONRPC2ParamRegistry.java
manishdait: Done

mustafauzunn (commented):
Some of the param classes implement the empty interface and some don't.
Do we need the empty interface?

manishdait:
Yes, that is intentional. The other parameter classes are used as internal field types, so they not implement the interface
The marker interface can be remove, since we are manually registering the methods to the `JSONRPC2ParamRegistry`.
I kept it because the issue specifically mentions using the abstract class as a marker interface :)

mustafauzunn (approved):

PR #2951, opened 2026-09-15
fix(tck): return deletedAt/executedAt in ScheduleInfoQuery response

### LINKED ISSUE #2950: TCK server: ScheduleInfoQuery response uses `deleted`/`executed` instead of spec's `deletedAt`/`executedAt`
## Description

The TCK server's `getScheduleInfo` response serialises the deletion and execution timestamps as `deleted` and `executed` (record components of `ScheduleInfoResponse`). The TCK spec's output table defines them as `deletedAt` and `executedAt`:

https://hiero-ledger.github.io/hiero-sdk-tck/test-specifications/schedule-service/ScheduleInfoQuery.html

The Go TCK server already returns the spec names. This was masked because the TCK tests reading these fields only asserted `undefined` or were guarded by an `if`. hiero-ledger/hiero-sdk-tck#712 (fixing hiero-ledger/hiero-sdk-tck#711) makes test #4 assert a non-empty `deletedAt`, which fails against the Java server:

```
AssertionError: expected { creatorAccountId: '0.0.2', …(9) } to have property 'deletedAt'
```

## Solution

Rename the two record components to `executedAt` / `deletedAt` and update the serialisation test.

ddeskov-limechain:
## Description

The TCK server's `getScheduleInfo` response serialised the deletion and execution timestamps as `deleted` / `executed`, while the TCK spec's output table defines `deletedAt` / `executedAt` (the Go TCK server already uses the spec names). Renames the two record components of `ScheduleInfoResponse` and updates its serialisation test.

Needed by hiero-ledger/hiero-sdk-tck#712, which makes TCK test #4 assert a non-empty `deletedAt`.

## Related issue(s)

Fixes #2950

## Notes for reviewer

Positional record rename only; `ScheduleService.getScheduleInfo` constructs the record positionally so it needs no change.

## Checklist

- [x] Documented (n/a, spec already defines the names)
- [x] Tested (`ScheduleInfoResponseTest` updated for the new keys)


ddeskov-limechain:
The `Run Examples` failure is unrelated to this change: it dies in `Setup Android SDK` (`Failed to find package 'tools'`) before any example runs. Google stopped serving the deprecated `tools` package that the action requests by default (android-actions/setup-android#537). Tracked in #2952 with a one-line workflow fix. `TCK Tests` passes here.

mustafauzunn (approved):

